PR #32, opened 2023-01-02
RDF Blank Nodes

jackson-millard:


charlesivie (approved):
Initial Blank node support approved.
n.b. 
Bklank nodes are supoorted here for Search and expansion of a node up to 2 levels only. 
That means you can find or navigate to a a BNode, and its child, but not its grand children.

PR #49, opened 2023-01-27
Fix credential refresh bug

jackson-millard:
Addresses credential refresh when credentials are either expired or undefined. Currently this is done blindly and in the case of any error in the response, it will attempt a credential refresh, retry the request and if the retry also fails, log the status code causing the failure.


triggan (approved):
Change looks good and validated that this works.  For the time being, we may just want to ensure that the error messages that are generated (and those that may appear in the server logs or docker logs) give the user enough information to resolve any IAM authentication issues.

PR #56, opened 2023-02-21
Moving self-signed SSL cert creation to docker entrypoint script.

triggan:
Issue #, if available: N/A

Description of changes:

Removed self-signed SSL cert creation from Dockerfile and created a new entrypoint script for certs to be created when the container is launched.  This removes the need for a user to build a new container image for each deployment and allows them to set the hostname used by the SSL cert upon launching the container using a `HOST` environment variable.

Updated README.md to include the updated steps for container image build and launching a new container.  Also added a new Troubleshooting section to address common issues seen by users attempting to launch graph-explorer.


michaelnchin (dismissed):
Tested and verified functionality. Looks good, thanks @triggan !

michaelnchin (approved):

Issue #55, opened 2023-02-21
Support for Non-String Vertex/Edge IDs with Property Graph Data


triggan:
**Describe the bug**
At present, any use of non-String vertex IDs appears in the UI as `[object Object]`.  

![image](https://user-images.githubusercontent.com/8203156/220413039-9d8150d6-b7d5-4f40-a297-17a32c131023.png)

In the query response, the GraphSON response object contains the following:

```
...
{
    "@type":"g:Vertex",
    "@value":{
        "id":{
            "@type":"g:Int64",
            "@value":2306
        },
        "label":"airport"
   }
}
...
```

With using string-based IDs, the following is returned:

```
...
{
    "@type":"g:Vertex",
    "@value":{
        "id":"2306",
        "label":"airport"
   }
}
...
```

The extra `type` field gets returned for non-string IDs, causing the values to not be serialized properly back to the UI.

**To Reproduce**

Using Gremlin Server and TinkerGraph (with a `vertexIdManager=ANY`), create two vertices:

`g.addV('airport').property(id,'2306')`
and
`g.addV('airport').property(id,2306)`

**Expected behavior**
The ID should appear as the same value regardless of data type used.

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser: Firefox
 - Version: 102.7.0esr

**Additional context**

krlawrence:
Adding a note here that a Gremlin Server also allows for an ID to be a UUID. The PR #60 appears to handle integer IDs, but what about if the ID is a UUID? In the graph-explorer project we recently ran into this and had to make some changes for the UUID ID type.

Issue #33, opened 2023-01-08
Connecting to non-neptune TinkerPop db


yuvallb:
Trying to connect to a TinkerPop database, I get the error "Synchronization Required
It is necessary to synchronize the connection to be able to work with the database."
Looking into network calls, I see it tries to query g.V().groupCount().by(label)
which is refused with the error message:
Could not find a suitable index to answer graph query and graph scans are disabled

Seems like an index on label is required to run graph-explorer

bechbd:
What TinkerPop database are you running against?

yuvallb:
JanusGraph
It has a specific setting to disable queries that do not use an index

krlawrence:
Are you able to enable full graph scans for the near term? The GraphExplorer does not try to interrogate the back end currently to see what indexes exist. Also, as far as I am aware, JanusGraph does not support indexing a label (or at least it did not used to when I used it more regularly).

otnt:
The same error happened to me as well when connecting to a gremlin server.

krlawrence:
@otnt Which database backend were you using, was it also JanusGraph? Are you still seeing the issue with the latest version of Graph Explorer? @yuvallb were you ever able to get it working with JanusGraph? Just checking for any more information before taking a deeper look at this one. Several fixes and improvements have been made in the meantime that may also help with this issue.

krlawrence:
I was able to reproduce the issue pretty much as described. Using:
1. JanusGraph 0.6.3 running with a Gremlin Server using the WsAndHttpsChannelizer and an in-memory graph. Everything else pretty much defaults.
2. Graph Explorer built to use HTTP rather than HTTPS (just to make connecting easier)
3. The schema of my small test graph was correctly fetched
4. The minute I clicked in the search bar of the Explorer an exception was thrown and the browser window went to just a white background with nothing else.

krlawrence:
Issue #55 is potentially related to this issue. It may not be the root cause but given the ID scheme used by JanusGraph the fix for that issue will also be required before JanusGraph fully works with GraphExplorer.

krlawrence:
I built a version of Graph Explorer using the current PR #60 and was able to get data to display using Janus Graph. Once PR #60 merges, this issue can hopefully be resolved/closed.

cubeddu:
Looks like issue has been resolved. please re-open if it's not the case.

PR #60, opened 2023-02-24
(bugfix/issue-55) Support non-string based ids for gremlin connector

nestoralvarezd:
Issue #55

#### Description of changes:
- Updates Gremlin connector to take into account non-string based ids for nodes and edges during the mapping

#### How Has This Been Tested?
- Mount a local instance of Tinkerpop `docker run --rm -p 8182:8182 tinkerpop/gremlin-server conf/gremlin-server-rest-modern.yaml` using the REST configuration. The default graph contains vertices of type person and software whose ids are like `{"@type":"g:Int64","@value":1}`. These kind of non-string ids are the reason of the issue.
- Create a new configuration with the following parameters:
  - Public endpoint: http://localhost:8182
  - Enable cache: false (only to guarantee that all the requests are always performed against the database)

- After synchronization, inspect the 4 people of the 2 software in the connection details
- "Node id" column now is not `[Object object]` but it shows the numeric value of the id.
- It has also been checked that this change does not affect to search or expand capabilities.


triggan:
This seems to be fetching the IDs as numerics now and pulling them into the tabular view of all vertices.  However, I'm now seeing an issue where fetching a vertex into the Explorer UI and attempting to expand it is not working correctly.

I'm testing with the following test graph in Gremlin Server [1]:

```
g.addV('testV').
as('1').
addV('testV').
as('2').
addE('testE').from('1').to('2')
```

Going into the tabular view of all vertices, I can select one of the two "testV' vertices and bring it into the Explorer UI successfully.  When I go to expand the vertex to look for neighbors via double-clicking the vertex, the UI states that there are no neighbor vertices.  After reviewing the Dev Tools Network tab, I noticed that we're generating the following query:

```
g.V("42774").both().dedup().group().by(label).by(count())
```

So the UI is attempting to fetch a vertex with the ID as a String instead of with a Long. The query should be:

```
g.V(42774L).both().dedup().group().by(label).by(count())
```

[1] Note that I'm allowing Gremlin Server to generate IDs for me and the default is to generate IDs as Longs.

krlawrence (changes requested):
Please see comments from @triggan

krlawrence:
I built a version of the Explorer with this PR included. It helped with the issues reported in #33 (Explorer not working with JanusGraph - which also uses numeric IDs).  The Explorer will now display nodes. However, when I expand a node, existing (displayed) edges get hidden.

nalvarezdiaz:
> This seems to be fetching the IDs as numerics now and pulling them into the tabular view of all vertices. However, I'm now seeing an issue where fetching a vertex into the Explorer UI and attempting to expand it is not working correctly.
> 
> I'm testing with the following test graph in Gremlin Server [1]:
> 
> ```
> g.addV('testV').
> as('1').
> addV('testV').
> as('2').
> addE('testE').from('1').to('2')
> ```
> 
> Going into the tabular view of all vertices, I can select one of the two "testV' vertices and bring it into the Explorer UI successfully. When I go to expand the vertex to look for neighbors via double-clicking the vertex, the UI states that there are no neighbor vertices. After reviewing the Dev Tools Network tab, I noticed that we're generating the following query:
> 
> ```
> g.V("42774").both().dedup().group().by(label).by(count())
> ```
> 
> So the UI is attempting to fetch a vertex with the ID as a String instead of with a Long. The query should be:
> 
> ```
> g.V(42774L).both().dedup().group().by(label).by(count())
> ```
> 
> [1] Note that I'm allowing Gremlin Server to generate IDs for me and the default is to generate IDs as Longs.

I've updated the PR to properly manage numeric ids. Now, we store the id type inside the GremlinConnector to be able to differentiate string based ids from numeric ids and fetch neighbors using the corresponding template.

krlawrence:
I think it's OK to fix it in a separate PR, but a Gremlin Server also allows UUIDs to be used as the ID type. I added a comment about this in issue #55

krlawrence:
I built the latest code with this PR included. I'm still seeing issues.
1. When you expand an edge, other already expanded edges vanish.
2. When you add a node and expand it (starting with a clear canvas) the 2 nodes briefly appear nicely with their edge shown but then immediately collapse on top of each other. I reported this above and it remains a problem.

krlawrence:
I built a new container image, and retested with the latest commits. The nodes not expanding correctly when first rendered issue (in Firefox) seems to be improved now. However, every layout seems to run twice now which looks a little odd to a user. The edges vanishing when the Explorer is connected to Janus Graph also seems fixed now. I also tested with Neptune.

krlawrence:
I built and tested with the latest fixes and the double redraw issue is now resolved. I tested on Firefox and Chrome. I think this PR can now be merged.

krlawrence (approved):
Latest updates seem to have resolved the remaining issues called out in the comments.

PR #80, opened 2023-03-20
Use the Summary API to improve the schema synchronization performance

nestoralvarezd:
# Summary API integration

The main aim of this PR is to reduce the synchronization time of the schema by using Neptune Summary API if it is available.

Description of changes:
- `fetchSchema` process tries to fetch some data, using the Summary API, related to the database structure:
  -  vertices types (classes),
  -  edges types (predicates),
  -  total number of vertices (total number of subjects),
  -  and total number of edges (total number of quads)
- If the above data is available, use them to fetch the vertices and edges attributes. By using the Summary API, the time of inferring the schema will be decreased. However, it does not return how many vertices or edges we have by type. So we speed up the process but lose the counts by type.
- If the Summary API is not available, the `fetchSchema` fallbacks to the previous process we had to infer the schema.

Now, if the `fetchSchema` process uses the Summary API, we will not have counts by type. So a new method has been added to the `AbstractConnector` contract, and its the implementations, in order to fetch the total number of nodes (subjects). It is the method called `fetchVertexCountsByType` and it is called when a user visits the Data Explorer section.

### DIFF
--- a/packages/graph-explorer-proxy-server/node-server.js
+++ b/packages/graph-explorer-proxy-server/node-server.js
@@ -32,7 +32,7 @@ const getCredentials = async () => {
 };
 
 dotenv.config({ path: "../graph-explorer/.env" });
-  
+
 (async () => {
   let creds = await getCredentials();
   let requestSig;
@@ -64,23 +64,23 @@ dotenv.config({ path: "../graph-explorer/.env" });
     await getAuthHeaders(language, req, reqObjects[0], reqObjects[2]);
 
     return new Promise((resolve, reject) => {
-        
+
       const wrapper = (n) => {
         fetch(url, { headers: req.headers, })
           .then(async res => {
 
             if (!res.ok) {
-	            console.log("Response not ok");
+              console.log("Response not ok");
               const error = res.status
               return Promise.reject(error);
             } else {
-	            console.log("Response ok");
+              console.log("Response ok");
               resolve(res);
             }
           })
           .catch(async (err) => {
             console.log("Attempt Credential Refresh");
-	          creds = await getCredentials();
+            creds = await getCredentials();
             if (creds === undefined) {
               reject("Credentials undefined after credential refresh. Check that you have proper acccess")
             }
@@ -119,7 +119,7 @@ dotenv.config({ path: "../graph-explorer/.env" });
   }
 
   async function getAuthHeaders(language, req, endpoint_url, requestSig) {
-    let authHeaders 
+    let authHeaders
     if (language == "sparql") {
       authHeaders = await requestSig.requestAuthHeaders(
         endpoint_url.port,
@@ -147,9 +147,9 @@ dotenv.config({ path: "../graph-explorer/.env" });
     let data;
     try {
       response = await retryFetch(`${req.headers["graph-db-connection-url"]}/sparql?query=` +
-      encodeURIComponent(req.query.query) +
-      "&format=json", undefined, undefined, req, "sparql").then((res) => res)
-      
+        encodeURIComponent(req.query.query) +
+        "&format=json", undefined, undefined, req, "sparql").then((res) => res)
+
       data = await response.json();
       res.send(data);
     } catch (error) {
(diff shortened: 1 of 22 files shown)


packages/graph-explorer-proxy-server/node-server.js
```diff
@@ -32,59 +32,70 @@ const getCredentials = async () => {
...
+          .then(async (res) => {
             if (!res.ok) {
-	            console.log("Response not ok");
-              const error = res.status
+              console.log("Response not ok");
+              const error = res.status;
               return Promise.reject(error);
             } else {
-	            console.log("Response ok");
+              console.log("Response ok");
```
gopuneet, line 83: I recommend having separate PR/commits for styling changes and business logic changes in the future since the one for styling/formatting changes can be quickly reviewed allowing reviewer to focus on business logic changes only.

packages/graph-explorer-proxy-server/node-server.js
```diff
@@ -162,9 +180,36 @@ dotenv.config({ path: "../graph-explorer/.env" });
...
+    try {
+      response = await retryFetch(
+        `${req.headers["graph-db-connection-url"]}/pg/statistics/summary`,
+        undefined,
+        undefined,
+        req,
+        "gremlin"
+      ).then((res) => res);
 
+      console.log(response);
```
gopuneet, line 212: The response from the graph summary endpoint can be pretty big for certain types of graphs.
Could there be any issues with logging the entire response to console in those cases?

Same question for L233 (log statement in rdf summary endpoint)

gopuneet:
Based on a quick look, it seems like we are fetching the graph summary by calling the respective property graph and rdf graph endpoints but there are no calls to [refresh statistics](https://docs.aws.amazon.com/neptune/latest/userguide/neptune-dfe-statistics.html#neptune-dfe-statistics-manual). Since the graph summary is computed using DFE statistics, how would the data shown on graph explorer (fetched using summary API) be updated? 
(when autocompute is on, the statistics only refresh when either more than 10% of data in the graph has changed or when the last statistics computation was more than 10 days old)

Pardon me if I've missed something and this is already handled.

---

This might have already been discussed but just a thought: (cc. @joywa )
Maybe we could have a button on the UI or periodic background thread that invokes the stats refresh, then polls the stats status until the new `statisticsId` reflects on stats status (indicates that the latest stats computation finished) and then updates the graph information using graph summary API.

packages/graph-explorer-proxy-server/node-server.js
nestoralvarezd: It was the pre-commit process configuration. It ran lint and prettier automatically on every modified file. It prevents wrong styling issues, so I need to figure out why in the previous update it didn't update the styling.

However, I could remove all those styling changes manually and prevent the pre-commit hook for this time.

nestoralvarezd:
@gopuneet 

*Based on a quick look, it seems like we are fetching the graph summary by calling the respective property graph and rdf graph endpoints but there are no calls to [refresh statistics](https://docs.aws.amazon.com/neptune/latest/userguide/neptune-dfe-statistics.html#neptune-dfe-statistics-manual). Since the graph summary is computed using DFE statistics, how would the data shown on graph explorer (fetched using summary API) be updated?*

> Now, we don't have any trigger in the UI to refresh the statistics. When user clicks on "sync schema", the client will fetch the latest statistics but it is not triggering any refresh operation. I'm not sure where we should add the option to refresh the statistics because the client makes that transparent to the user. Where do you think, @joywa, is a good place to add the requested trigger?

*Please ensure you test this change against a Neptune cluster on engine version <1.2.1.0 to ensure that it still works for older Neptune engine versions where graph summary API is not present using fallback queries.*

> Sure, we should make some testing on version with and without support for Summary API. However, the process has a fallback that on any unexpected error by calling the summary API, it will fallback to the "standard" inference.

packages/graph-explorer-proxy-server/node-server.js
joywa: @nestoralvarezd can you please make an update to remove styling changes in this PR? Once approved, we can add the styling changes.

packages/graph-explorer-proxy-server/node-server.js
joywa: @nestoralvarezd to comment

joywa:
@gopuneet @nestoralvarezd we decided to not use the **_refresh statistics_** to trigger updates.

@nestoralvarezd please let us know when testing is complete/if any changes are required.

packages/graph-explorer-proxy-server/node-server.js
nestoralvarezd: I can remove this kind of large logging

nalvarezdiaz:
I've removed the most of the styling issue from this PR and I've checked that the summary API is working. Neptune instances  that do not have the summary API try to use it but then fallback to the normal inference process.

gopuneet (dismissed):

charlesivie (approved):

PR #108, opened 2023-05-12
Adds Default Connection Support 

jackson-millard:
**Description:**
Modifies the docker build and run process by pushing the `pnpm build` command into the `docker-entrypoint.sh` file so that the `docker run` command can take environment variables and add them to the `.env` file before the build is created. This allows for a default connection to be provided. Currently supports a json called `config.json` provided via a volume mount or as individual variables provided via `--env`. See `README.md` for how to execute either option. From a process standpoint, the variables passed in either manner get processed and appended to the .env file in process-environment.sh so the build can use them.

**Testing:**
This was tested via local builds by providing the json solution, the --env solution, and providing both, in which case the json takes precedence. To test, make sure that your browser is not caching connections from one build to another. Opening an incognito window during testing makes for the easiest way to circumvent this.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,10 +5,9 @@ COPY . /graph-explorer/
 WORKDIR /graph-explorer
 # Keeping all the RUN commands on a single line reduces the number of layers and,
 # as a result, significantly reduces the final image size.
-RUN curl -sL https://rpm.nodesource.com/setup_16.x | bash - && yum install -y nodejs openssl && npm install -g pnpm && pnpm install && rm -rf /var/cache/yum
+RUN curl -sL https://rpm.nodesource.com/setup_16.x | bash - && yum install -y nodejs openssl && npm install -g pnpm && pnpm install && rm -rf /var/cache/yum && chmod a+x ./process-environment.sh
 WORKDIR /graph-explorer/
 ENV HOME=/graph-explorer
-RUN pnpm build
 EXPOSE 443
 EXPOSE 80
 RUN chmod a+x ./docker-entrypoint.sh
--- a/README.md
+++ b/README.md
@@ -94,6 +94,23 @@ You can search, browse, expand, customize views of your graph data using Graph E
 ## Connections
 The Graph Explorer supports visualizing both **property graphs** and **RDF graphs**. You can connect to Amazon Neptune or you can also connect to open graph databases that implement an Apache TinkerPop Gremlin Server endpoint or the SPARQL 1.1 protocol, such as Blazegraph. For additional details on connecting to different graph databases, see [Connections](./additionaldocs/connections.md).
 
+### Providing a Default Connection 
+To provide a default connection such that initial loads of the graph explorer always result with the same starting connection, modify the `docker run ...` command to be `docker run -p 80:80 -p 443:443 --env HOST={hostname-or-ip-address} -v /path/to/config.json:/graph-explorer/config.json graph-explorer` or create a `config.json` file at the root of the project.
+
+The config.json file should provide values for the attributes in the example below:
+
+```
+{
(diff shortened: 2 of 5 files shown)


krlawrence:
Moving `pnpm build` inside the container this way is problematic. That build process takes around 30 seconds. During that time the container will appear to the suer to be up and running but in reality it is still building "behind the scenes". Unless they are tail-ing the Docker logs with something like `docker logs -f explorer` they will have no idea when it is ready. This also means that each time the container launches it will re-do these build steps. I would like to explore different ways of solving this problem.

agutierrezgit (commented):
I have tested the PR for HTTP and HTTPS locally following the instructions from @jackson-millard above. I tested it with both a .json configuration file and runtime environment variables. As far as I understand all the comments above are addressed with this [commit](https://github.com/aws/graph-explorer/pull/108/commits/cf07a96923271bb9f28ba15b06e64198de900367). If there are no further concerns in my opinion this PR is ready for approval. Anyone else has tested and can provide feedback? Thanks

michaelnchin (changes requested):
Thanks @jackson-millard ! I've tested the changes in this PR with the [SageMaker development branch](https://github.com/aws/graph-explorer/tree/AddDefaultConfiguration-on-sagemaker-dev), and everything works as described when Explorer is started with default connection arguments via a Lifecycle configuration.

A few feature changes to request:
1. In cases where the optional "GRAPH_TYPE" field/argument is not specified, we currently default to creating one profile on Gremlin. Can the default behavior be changed so that Explorer instead creates a separate connection profile for every available language? The rest of the connection fields should be identical in each profile.
2. Somewhat related to point 1 - for default connections, the "GRAPH_TYPE" and "PUBLIC_OR_PROXY_ENDPOINT" fields are grayed out and read-only. If a user made a mistake in initial setup, or would otherwise like to update these fields, they would be required to create a separate connection profile. Can this be changed so that all default connection fields are modifiable while Explorer is running? 

Screen Shot 2023-07-17 at 6 06 33 PM

agutierrezgit:
> Thanks @jackson-millard ! I've tested the changes in this PR with the [SageMaker development branch](https://github.com/aws/graph-explorer/tree/AddDefaultConfiguration-on-sagemaker-dev), and everything works as described when Explorer is started with default connection arguments via a Lifecycle configuration.
> 
> A few feature changes to request:
> 
> 1. In cases where the optional "GRAPH_TYPE" field/argument is not specified, we currently default to creating one profile on Gremlin. Can the default behavior be changed so that Explorer instead creates a separate connection profile for every available language? The rest of the connection fields should be identical in each profile.
> 2. Somewhat related to point 1 - for default connections, the "GRAPH_TYPE" and "PUBLIC_OR_PROXY_ENDPOINT" fields are grayed out and read-only. If a user made a mistake in initial setup, or would otherwise like to update these fields, they would be required to create a separate connection profile. Can this be changed so that all default connection fields are modifiable while Explorer is running?
> 
> Screen Shot 2023-07-17 at 6 06 33 PM

@michaelnchin I'm glad you brought up these issues. I've opened a PR against the current PR to address point 2, [PR](https://github.com/jackson-millard/aws-graph-explorer/pull/1) However, I suggest we discuss point 1 further before proceeding. I've raised a concern about treating default connections as fileBase, which could prevent users from deleting them later. To avoid this, we might need to consider an alternative approach and by fixing this, point 2 will be also automatically resolved. 

I propose merging the current PR as a starting point, and then we can create a new ticket to explore improvements for handling default connections. This way, we can iterate and address the issue more thoroughly. Your thoughts are appreciated. Thanks!

michaelnchin (approved):
LGTM. Thanks for all your work on this @jackson-millard and @agutierrezgit !

michaelnchin (approved):

michaelnchin (approved):

Issue #107, opened 2023-05-10
[Feature Request] Ability to search by node Id rather than just property


kevinphillips81:
**Community Note**
* Please use a 👍 reaction to provide a +1/vote. This helps the community and maintainers prioritize this request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

**Graph Explorer Version (and Graph Database and Version used if applicable)**
All versions

**Is your feature request related to a problem? Please describe.**
For Property Graph models, using the search bar, you can only search by a property value rather than by the ID of an graph object.

**Describe the solution you'd like**
When entering a search term that represents the ID of a graph object into the search bar, the results should include all nodes whose ID, as well property values, match the search criteria.

PR #113, opened 2023-05-18
Adds id search by default to gremlin keyword search

jackson-millard:
Issue #107:

Description of changes:
Sets id search to true in the request and also by default for any gremlin connection so that searches now include the vertex ids in the possible parameters that an input string can match. 

Tested via a local build attempting to search on an id and received the expected node.


michaelnchin (approved):
Changes look good in the UI. Thanks @jackson-millard !

cubeddu (approved):
@jackson-millard 
test it locally and search for a node-id 3729 (sample)
image

Issue #98, opened 2023-04-17
[Feature Request] Make synchronization button more prominent


joywa:
**Community Note**
* Please use a 👍 reaction to provide a +1/vote. This helps the community and maintainers prioritize this request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

**Graph Explorer Version (and Graph Database and Version used if applicable)**
All versions

**Is your feature request related to a problem? Please describe.**
Currently it is not clear a user can click the re-sync button because it is so small and the color is similar to static text.

**Describe the solution you'd like**
Make this a bigger button and on the same level as editing / deleting a connection or similar.

PR #120, opened 2023-06-20
updated styles(info-bar), updated component to add new Re-Sync column…

cubeddu:
Issue #98 

Description of changes:

This PR addresses the feature request to make the synchronization button more prominent by increasing the size of the icon. The current icon size is `small` and this PR increases it to `large`. This change will make it easier for users to locate and use the synchronization button.

In addition, created a Re-Sync column to the info-bar div to make it a prominent placement. This will provide users with a clear indication of the synchronization status and allow them to easily initiate a re-sync if necessary.

Furthermore, updated the CSS styles from `flex-start` to `space-between`. This change will improve the layout of the info-bar div and provide better spacing between the different elements.

Before:
Screenshot 2023-06-20 at 6 35 26 PM

After:
Screenshot 2023-06-20 at 6 08 04 PM

### DIFF
--- a/packages/graph-explorer/src/modules/ConnectionDetail/ConnectionDetail.tsx
+++ b/packages/graph-explorer/src/modules/ConnectionDetail/ConnectionDetail.tsx
@@ -154,14 +154,6 @@ const ConnectionDetail = ({ isSync, onSyncChange }: ConnectionDetailProps) => {
           <div className={pfx("item")}>
             <div className={pfx("tag")}>
               <div>Last Synchronization</div>
-              <IconButton
-                icon={<SyncIcon className={isSync ? "animate-spin" : ""} />}
-                tooltipText={"Synchronize Database"}
-                variant={"text"}
-                size={"small"}
-                isDisabled={isSync}
-                onPress={onConfigSync}
-              />
             </div>
             {!lastSyncFail && (
               <div className={pfx("value")}>{formatDate(lastSyncUpdate)}</div>
@@ -178,6 +170,21 @@ const ConnectionDetail = ({ isSync, onSyncChange }: ConnectionDetailProps) => {
             )}
           </div>
         )}
+        <div className={pfx("item")}>
+          <div className={pfx("ft-end-adornment")}>
+            <div className={pfx("tag")}>Re-Sync</div>
+            <div className={pfx("value")}>
+              <IconButton
+                icon={<SyncIcon className={isSync ? "animate-spin" : ""} />}
+                tooltipText={"Synchronize Database"}
+                variant={"text"}
+                size={"base"}
+                isDisabled={isSync}
+                onPress={onConfigSync}
+              />
+            </div>
+          </div>
+        </div>
       </div>
       {!isSync && !!lastSyncUpdate && <ConnectionData />}
       {!lastSyncUpdate && !isSync && (
--- a/packages/graph-explorer/src/modules/ConnectionDetail/ConnectionDetail.tsx
+++ b/packages/graph-explorer/src/modules/ConnectionDetail/ConnectionDetail.tsx
@@ -38,7 +38,7 @@ export type ConnectionDetailProps = {
 
 const HEADER_ACTIONS = (isSync: boolean, isFileBased: boolean) => [
   {
-    label: "Synchronize Database",
+    label: "Sync Connection",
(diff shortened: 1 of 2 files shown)


agutierrezgit:
@cubeddu thanks for your suggestion. 
We need to consider the potential issue of items getting too separated on larger screens if we let the items space themselves up to the full width. We recommend keeping the content on the left side. 

We also recommend moving the sync button to the global actions in the header showing a tooltip on hover. This way, it remains easily accessible to users while maintaining a consistent layout across different screen sizes. See picture below.

Captura de pantalla 2023-06-27 a las 9 51 56

If you think it should be more prominent to the user we can go with a full button.

Captura de pantalla 2023-06-27 a las 9 47 27

cubeddu:
> @cubeddu thanks for your suggestion. We need to consider the potential issue of items getting too separated on larger screens if we let the items space themselves up to the full width. We recommend keeping the content on the left side.
> 
> We also recommend moving the sync button to the global actions in the header showing a tooltip on hover. This way, it remains easily accessible to users while maintaining a consistent layout across different screen sizes. See picture below.
> 
> Captura de pantalla 2023-06-27 a las 9 51 56
> If you think it should be more prominent to the user we can go with a full button.
> 
> Captura de pantalla 2023-06-27 a las 9 47 27

@agutierrezgit 
I agree, It does look a lot better as you recommend, I went ahead and made those changes ready for a second review.

image

packages/graph-explorer/src/modules/ConnectionDetail/ConnectionDetail.tsx
```diff
@@ -37,6 +37,12 @@ export type ConnectionDetailProps = {
 };
 
 const HEADER_ACTIONS = (isSync: boolean, isFileBased: boolean) => [
+  {
+    label: "Synchronize Database",
+    icon: <SyncIcon />,
```
agutierrezgit, line 42: It would be nice to keep the loading state. `icon={<SyncIcon className={isSync ? "animate-spin" : ""} />}` should do the work.

packages/graph-explorer/src/modules/ConnectionDetail/ConnectionDetail.tsx
```diff
@@ -38,7 +38,7 @@ export type ConnectionDetailProps = {
 
 const HEADER_ACTIONS = (isSync: boolean, isFileBased: boolean) => [
   {
-    label: "Synchronize Database",
+    label: "Sync Connection",
```
agutierrezgit, line 41: Apologies for the confusion with my screenshot but I think we should keep the old label. Thanks!

agutierrezgit (changes requested):
@cubeddu Thanks for the quick turnaround. If you can add these small changes this will be ready to go.

packages/graph-explorer/src/modules/ConnectionDetail/ConnectionDetail.tsx
cubeddu: sure not a problem, looks so much better with the animation!

packages/graph-explorer/src/modules/ConnectionDetail/ConnectionDetail.tsx
cubeddu: Hey, @agutierrezgit Updates have been pushed up.

cubeddu:
Hey @agutierrezgit 
Changes have been pushed up.

Thanks

agutierrezgit (commented):
Thanks @cubeddu ! @joywa I tested the changes and the PR is ready for approval and merge.

packages/graph-explorer/src/modules/ConnectionDetail/ConnectionDetail.tsx
```diff
@@ -37,6 +37,12 @@ export type ConnectionDetailProps = {
 };
 
 const HEADER_ACTIONS = (isSync: boolean, isFileBased: boolean) => [
+  {
+    label: "Synchronize Database",
```
alexey-temnikov, line 41: @cubeddu, what would be the strategy to support i18n?

cubeddu: Not currently, I when ahead and added a feature request here issue #128

Issue #34, opened 2023-01-11
[Bug] Count of neighbors in popup in Expand View sidebar shows incorrect number


kevinphillips81:
When expanding an item, the popup description shows the incorrect number of related items currently being displayed in the graph view. See details below:

![image](https://user-images.githubusercontent.com/52177911/211837220-39ba5386-a107-4162-927f-dfdb7fd4c6ba.png)

PR #121, opened 2023-06-21
updated inView to neighborsInView

cubeddu:
Issue #34 

Description of changes:

This PR addresses a bug where the count of neighbors in the popup in Expand View sidebar shows an incorrect number. The issue was caused by a calculation error in the `inView` constant. 

Before this PR, the `inView` constant was calculated as follows:
```
const inView =
  vertex.data.neighborsCountByType[op.value] -
  vertex.data.neighborsCountByType[op.value];
```
This calculation resulted in an incorrect count of neighbors in the popup.

After this PR, the `inView` constant is calculated as follows:
```
const neighborsInView =
  vertex.data.neighborsCountByType[op.value] -
  (vertex.data.__unfetchedNeighborCounts?.[op.value] ?? 0);
```
This calculation correctly subtracts the number of unfetched neighbor counts from the total count of neighbors, resulting in an accurate count of neighbors in the popup.

This PR resolves the bug and ensures that the count of neighbors in the popup is displayed correctly.
Pasted Graphic

### DIFF
--- a/packages/graph-explorer/src/modules/common/NeighborsList/NeighborsList.tsx
+++ b/packages/graph-explorer/src/modules/common/NeighborsList/NeighborsList.tsx
@@ -29,9 +29,9 @@ const NeighborsList = ({
         Neighbors ({vertex.data.neighborsCount})
       </div>
       {neighborsOptions.map(op => {
-        const inView =
+        const neighborsInView =
           vertex.data.neighborsCountByType[op.value] -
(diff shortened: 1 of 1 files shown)


agutierrezgit (commented):
Good catch @cubeddu! @joywa I am happy with the changes

joywa (approved):
Thank you @cubeddu! Approved on behalf of @agutierrezgit .

Issue #67, opened 2023-03-02
Support openCypher-based graph databases


joywa:
**Community Note**
* Please use a 👍 reaction to provide a +1/vote. This helps the community and maintainers prioritize this request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

**Is your feature request related to a problem? Please describe.**
Currently the Graph Explorer only supports viewing property graph data using Gremlin queries. I would like to use Explorer to connect to openCypher-based graph databases.

**Describe the solution you'd like**
Support connections to openCypher graph databases

PR #129, opened 2023-07-07
(issue #67)feature/open cypher

jackson-millard:
Issue #67:

Description of changes:
Adds a connector for OpenCypher, which follows a similar structure to the Gremlin connector. To choose between OpenCypher or Gremlin, you will notice there are now two "PG (Property Graph)" entries on the Graph Type drop down when updating or creating a new connection. There will be one for OpenCypher and another for Gremlin. Each labeled as such. Currently this has only been tested on Neptune and is only expected to work for Neptune as a result. To support this connection, an `/openCypher` endpoint has been added to the node-server.js file and this is where openCypher requests are proxied.


packages/graph-explorer/src/connector/openCypher/queries/fetchNeighbors.ts
```diff
@@ -0,0 +1,45 @@
...
+            ];
+        },
+    ];
+};
+
+const fetchNeighbors = async (
+openCypherFetch: OpenCypherFetch,
+req: NeighborsRequest
+): Promise<NeighborsResponse> => {
+    const gremlinTemplate = oneHopTemplate(req);
```
agutierrezgit, line 28: This one too?

packages/graph-explorer/src/modules/CreateConnection/CreateConnection.tsx
```diff
@@ -34,7 +34,8 @@ type ConnectionForm = {
 };
 
 const CONNECTIONS_OP = [
-  { label: "PG (Property Graph)", value: "gremlin" },
+  { label: "PG (Property Graph) - Gremlin", value: "gremlin" },
+  { label: "PG (Property Graph) - OpenCypher", value: "openCypher" },
   { label: "RDF (Resource Description Framework)", value: "sparql" },
```
charlesivie, line 39: Could we add the term - SPARQL for consistency please;
"RDF (Resource Description Framework) - SPARQL"

charlesivie (changes requested):
under additionaldocs/development can we please add opencypher here : 
### Supported Graph Types
- Labelled Property Graph (PG) using Gremlin **or OpenCypher**
- Resource Description Framework (RDF) using SPARQL

charlesivie:
When the Summary API is not available, for example running against Neptune engine version prior to v1.2.1x, the synchronisation fails with a 500.

packages/graph-explorer/src/modules/CreateConnection/CreateConnection.tsx
agutierrezgit: Done

michaelnchin (approved):

PR #166, opened 2023-08-15
Fix proxy issue with non-IAM connections

michaelnchin:
Issue #, if available: N/A

Description of changes:
- Fixed a bug where requests from Explorer to non IAM-enabled Neptune endpoints would fail in cases where HTTPS was enabled on the proxy, with the error: `Hostname/IP does not match certificate's altnames`.
- Bump Graph Explorer version to 1.3.1 for patch release

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,12 @@
 # graph-explorer Change Log
 
+## Release 1.3.1
+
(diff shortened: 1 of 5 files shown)


krlawrence (approved):
LGTM thanks for fixing this expeditiously

Issue #156, opened 2023-08-05
Deprecation warnings in GitHub workflows


michaelnchin:
Recent runs of our automated GitHub actions for ECR deployments and unit tests have some deprecation warnings requiring updates to imported actions. These are to do with the campaigns for:
1. NodeJS 12 -> 16: https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/
2. `set-output` command removal: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

In [build_docker.yml](https://github.com/aws/graph-explorer/blob/main/.github/workflows/build_docker.yml):
- For actions: `tyankatsu0105/read-package-version-actions@v1`, `aws-actions/configure-aws-credentials@v1`
- Example run: https://github.com/aws/graph-explorer/actions/runs/5767952199

In [unit.yml](https://github.com/aws/graph-explorer/blob/main/.github/workflows/unit.yml):
- For actions: `codecov/codecov-action@v2`
- Example run: https://github.com/aws/graph-explorer/actions/runs/5764302151

PR #181, opened 2023-08-29
Deprecation warnings in GitHub workflows

cubeddu:
Issue [#156](https://github.com/aws/graph-explorer/issues/156)

Updated build_docker.yml and unit.yml to address annotations from deprecation warnings.

This PR addresses issue #156 by updating the build_docker.yml and unit.yml files to use the latest versions of the actions and address annotations from deprecation warnings.

In build_docker.yml, we updated the `actions/checkout` action to the latest version.

In unit.yml, we updated the `actions/setup-node` action to version 2 and removed the `node-version` option, which is now deprecated. We also updated the `codecov/codecov-action` action to version 3 and removed the `directory` and `files` options, which are not necessary for the default configuration.

Links to the updated files:
- [.github/workflows/build_docker.yml](https://github.com/aws/graph-explorer/compare/main...cubeddu:graph-explorer:bug/actions#diff-c7f993a522ae12469fcfb15fc43197a8be4d9ef9e1350d98e775b2603eec9bd0)
- [.github/workflows/unit.yml](https://github.com/aws/graph-explorer/compare/main...cubeddu:graph-explorer:bug/actions#diff-e17ac7ab478a18e9dac875e191cc3e0d754fab62eddddfb05d4619495be927d9)


.github/workflows/build_docker.yml
```diff
@@ -31,18 +31,19 @@ jobs:
...
         id: package-version
+        if: runner.os != 'Windows'
 
       - name: Get image tag
         id: get-image-tag
         run: |
           if ${{ github.event_name == 'workflow_dispatch' }} ; then
             if ${{ inputs.image_tag == 'release'}}; then
-              echo "image_tag=${{ steps.package-version.outputs.version }}" >> $GITHUB_OUTPUT
+              echo "image_tag=${{ steps.package-version.outputs.version }}" >> $GITHUB_ENV
```
michaelnchin, line 41: The `GITHUB_OUTPUT` env variable should be left alone here and elsewhere in this workflow - this is the replacement for the old `"::set-output` command and should not trigger warnings.

The deprecation warnings are coming from the third-party actions `tyankatsu0105/read-package-version-actions@v1` ([on L30](https://github.com/aws/graph-explorer/blob/main/.github/workflows/build_docker.yml#L30)) and `aws-actions/configure-aws-credentials@v1` ([on L49](https://github.com/aws/graph-explorer/blob/main/.github/workflows/build_docker.yml#L49)). 

We should be able to upgrade the latter to  `aws-actions/configure-aws-credentials@v2` without issue (see [here](`aws-actions/configure-aws-credentials@v1)). The first action for reading the JS package version will need a replacement library, as the author has not updated it in several years.

michaelnchin (changes requested):

.github/workflows/build_docker.yml
cubeddu: @michaelnchin 
I went ahead a reverted the changes made to the Get image tag step, updated those actions, and replaced the package version with what looks more promising action.

.github/workflows/build_docker.yml
```diff
@@ -57,7 +55,7 @@ jobs:
 
       - name: Login to Amazon ECR
         id: login-ecr-public
-        uses: aws-actions/amazon-ecr-login@v1
+        uses: karlonovak/amazon-ecr-login@v2.0.0
```
michaelnchin, line 58: I would prefer that we stay on the AWS-maintained version of this action - could you elaborate on the reason for changing this? 

I believe the latest versions of the original action shouldn't trigger any of the deprecation warnings, the requisite changes were made to remove `set-output` (see [here](https://github.com/aws-actions/amazon-ecr-login/issues/358#issuecomment-1295750476)) and upgrade to Node 16 (see [here](https://github.com/aws-actions/amazon-ecr-login/commit/29516697cce506511bbd2ab14eec27dcfb5626b9)).

michaelnchin (changes requested):

.github/workflows/build_docker.yml
michaelnchin: Awesome, thank you for the updates @cubeddu 👍

.github/workflows/build_docker.yml
cubeddu: not a problem I reverted it back to stay within scope.

michaelnchin (dismissed):
Looks good, thanks @cubeddu !

michaelnchin (approved):

Issue #133, opened 2023-07-13
[Bug] Sychronization fails for large RDF dataset


michaelnchin:
**Community Note**
* Please use a 👍 reaction to provide a +1/vote. This helps the community and maintainers prioritize this request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

**Describe the bug**

A customer is attempting to synchronize Explorer to a large RDF graph loaded into a Neptune cluster with Summary API support. The dataset, [DBLP](https://dblp.org/rdf/), in RDFXML format, has a size of 24GB, and contains ~364m triples accounting for duplicates.

On synchronization, we can confirm that this is not due to connectivity issues with the Neptune cluster, as the initial Summary API request succeeds. However, when the remaining SPARQL HTTP requests are executed to fetch the remaining schema information, all of them error out with a `504 Gateway Timeout` due to a base limit of 60 seconds per fetch request.

If we execute some of these same SPARQL queries through graph-notebook, they are able to succeed. For example, the [predicatesByClass](https://github.com/aws/graph-explorer/blob/b371145b39ddef878ab349b5dce18e271d97524a/packages/graph-explorer/src/connector/sparql/templates/predicatesByClassTemplate.ts#L4) query completes in ~86 seconds - this does not exceed the base Neptune query timeout of 2 minutes, but is too high for Explorer.

![Screen Shot 2023-07-12 at 11 38 25 PM](https://github.com/aws/graph-explorer/assets/10456376/ec6cf5e9-c5f8-4923-8360-f4243d170619)

**Fixes tried**

I've looked into possible solutions involving hard-coding an increased timeout limit into the request. The fetch API, as used in the base [_requestAndCache method](https://github.com/aws/graph-explorer/blob/b371145b39ddef878ab349b5dce18e271d97524a/packages/graph-explorer/src/connector/AbstractConnector.ts#L307) in AbstractConnector, seems to accept a custom timeout through its `signal` init setting when it is defined as [AbortSignal.timeout([time_in_ms])](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/timeout_static). 

[...]

PR #199, opened 2023-09-27
RDF - Timeout when synchronizing

cubeddu:
Issue #133 

## Solution
To address the timeout issues, we added a Fetch timeout field to allow it to be configurable via the connection. We also adjusted the slow query to perform better by replacing it with a faster query that uses a subquery.

## Changes Made
Added Fetch timeout field to allow it to be configurable via connection.
Replaced the slow predicatesByClass query with a faster query that uses a subquery.

## Testing
We tested the changes by running synchronization with different timeout values and verifying that it was completed successfully. We also tested the new query to ensure that it returned the expected results.

### [Slow query]([packages/graph-explorer/src/connector/sparql/templates/predicatesByClassTemplate.ts](https://github.com/aws/graph-explorer/compare/main...cubeddu:bug/timeout?expand=1#diff-391a7ec891b8edb2f074f3f9f3486334aa1ad54f6b4cb82b6d44b76feb7756ab))
```
SELECT DISTINCT ?pred (SAMPLE(?object) as ?sample) {
        ?subject a     <${props.class}>;
                 ?pred ?object.
        FILTER(!isBlank(?object) && isLiteral(?object))
    }
    GROUP BY ?pred
```
Pasted Graphic

### updated query
```
SELECT ?pred (SAMPLE(?object) as ?sample)
    WHERE {
      {
        SELECT ?subject
        WHERE {
          ?subject a <${props.class}>.
        }
        LIMIT 1
      }
      ?subject ?pred ?object.
      FILTER(!isBlank(?object) && isLiteral(?object))
    }
    GROUP BY ?pred
```
Pasted Graphic 1

## Screenshots
image

image


michaelnchin (approved):
Thanks @cubeddu ! I tested the build against the DBLP RDF dataset, and the connection sync completed with significantly better performance. These changes additionally resolved the same timeout issue faced by our openCypher connector on large PG datasets.

Issue #217, opened 2023-11-09
Neptune connections with SIG4 different service names


charlesivie:
Neptune IAM connections require a SIG4 signature to connect. 
Currently, the signature is generated with a hard coded:
'service = neptune-db' 

Feature: add a parameter so that the service name can be over-ridden when creating a connection to a Neptune database. 

Archiving for now

PR #218, opened 2023-11-10
PR to fix Neptune connections with SIG4 different service names

cubeddu:
Issue #217 

## Description
This pull request introduces a new feature that enhances the flexibility of AWS Neptune IAM connections in the Graph Explorer. The main enhancement is the ability to switch the service type between "neptune-db" and "neptune-graph" for connections created through the Graph Explorer UI or set as the default connection.

## Changes Made
Removed RequestSig.js as is no longer being used.

## UI
image


README.md
```diff
@@ -122,6 +123,7 @@ To provide a default connection such that initial loads of the graph explorer al
     - `GRAPH_CONNECTION_URL` - `None` - See [Add a New Connection](#connections-ui)
   - Required if `USING_PROXY_SERVER=True` and `IAM=True`
     - `AWS_REGION` - `None` - See [Add a New Connection](#connections-ui)
+    - `SERVICE_TYPE` - `Neptune-db`
```
michaelnchin, line 126: Change this to `neptune-db` as service name is case-sensitive

michaelnchin (changes requested):
Thanks @cubeddu and @charlesivie !

I tested the new service type parameter out via a SageMaker Notebook stack, and openCypher connections were working great against neptune-graph.

Couple of issues/requests I'd like to see addressed:
1. When initializing Explorer with `SERVICE_TYPE` env variable set to `neptune-graph`, the primary default connection displayed on open should be changed from Gremlin -> openCypher. The primary connection should remain as Gremlin as the default and for `neptune-db`.

2. When service name is initially set as `neptune-graph` via the `SERVICE_TYPE` env variable, Explorer will use correctly use the indicated service for requests. However, this is not visually reflected in the Connections UI, where all default profiles will continue to display the default "Neptune Graph" in the "Service Type" field.

3. Updating the "Service Type" in a Connections UI profile isn't functioning as intended after using the `SERVICE_TYPE` env variable. i.e. if we initially set the service to `neptune-graph`, all requests from both default and manually created connection profiles will continue to be signed with `neptune-graph`, regardless of the "Service Type" selected (see second point).

README.md
cubeddu: updated.

cubeddu:
@michaelnchin As part of the updates, we made sure to allow the service type to flow all the way to the proxy server via headers, so now you will be able to update the service type per new connection or update the current one.

these changes should handle all 3 cases.

PR #259, opened 2024-03-08
Server query cancellation

vkagamlyk:
Issue #, if available:

Description of changes:

If the user starts a new search without waiting for a result, the request continues to be executed on the server. In this PR I added cancellation of the request on the server for Gremlin and SPARQL


packages/graph-explorer/src/connector/openCypher/useOpenCypher.ts
kmcginnes, line 1: Since we can't set a client generated queryId in OpenCypher, should we undo the changes to this file?

packages/graph-explorer/src/connector/useGEFetch.ts
```diff
@@ -43,6 +43,9 @@ const useGEFetch = () => {
     async (url, options) => {
       const response = await fetch(url, options);
       const data = await response.json();
+      if (typeof options.successCallback === "function") {
+        options.successCallback();
+      }
```
kmcginnes, line 48: This will miss a couple of scenarios:

- We get a response, but the JSON decoding fails
- The connection is dropped during the `fetch`

kmcginnes: One possible solution is to call the callback in all scenarios to ensure the queryId is cleared.

TanStack Query has a similar concept, and they named the callback `onSettled`, which I think is fitting.

https://tanstack.com/query/latest/docs/framework/react/reference/useMutation

packages/graph-explorer/src/connector/useGEFetchTypes.ts
```diff
@@ -7,6 +7,8 @@ import {
 
 export type QueryOptions = RequestInit & {
   disableCache?: boolean;
+  queryId?: string;
+  successCallback?: any;
```
kmcginnes, line 11: Let's define this type to be the function you expect.

packages/graph-explorer/src/connector/useGEFetch.ts
vkagamlyk: replaced with try/finally

kmcginnes (approved):
I have tested the code locally and it looks good.

michaelnchin (approved):
Thanks @vkagamlyk !

PR #417, opened 2024-05-28
Changes to Sync the request body with UI

mayurvaid-redvest:
UI is sending the request with gremlin as root rather than query

### Check List

      license.
- [ ] I have run `pnpm checks` to ensure code compiles and meets standards.
- [ ] I have run `pnpm test` to check if all tests are passing.

### DIFF
--- a/packages/graph-explorer-proxy-server/node-server.js
+++ b/packages/graph-explorer-proxy-server/node-server.js
@@ -279,7 +279,7 @@ app.post("/gremlin", async (req, res, next) => {
     : "";
 
   // Validate the input before making any external calls.
-  if (!req.body.query) {
+  if (!req.body.gremlin) {
     return res
       .status(400)
       .send({ error: "[Proxy]Gremlin: query not provided" });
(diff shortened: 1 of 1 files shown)


kmcginnes:
@mayurvaid-redvest Thanks for the submission.

Can you help me understand a bit about the motivation for this change?

mayurvaid-redvest:
@kmcginnes  . This came up when we were testing this and it is throwing bad request for sparql queries

mayurvaid-redvest:
@kmcginnes  , Any comments please let me know ?

kmcginnes:
@mayurvaid-redvest Sorry for the delay. I have not forgotten about this. I have a few other things to work on before diving in to this.

My plan was to load this up and test it before commenting, but since I haven't found the time I'll just post my preliminary thoughts instead to see what you think.

After taking just a cursory look at the changes, I don't understand why you were seeing an issue in the first place. The Gremlin `Explorer` instance in the client explicitly sets the body to `{ query: "...query string..." }`.

https://github.com/aws/graph-explorer/blob/5323b930d36fb1d304f57de3806ccc8706dd9caa/packages/graph-explorer/src/connector/gremlin/gremlinExplorer.ts#L13-L31

So your change to the proxy doesn't make sense because you changed it to look for `{ gremlin: "...query string..." }` instead. That shouldn't work.

You also mentioned you have issues running SPARQL queries, but you changed the Gremlin endpoint in the proxy. I don't understand how those are supposed to relate to each other.

Can you help me understand?

mayurvaid-redvest:
@kmcginnes , This is my bad sorry was looking at the wrong branch/code , Probably was fixed later on

kmcginnes:
@mayurvaid-redvest no problem. Thanks for the submission either way.

PR #818, opened 2025-03-04
feat: add support for accessing AWS Neptune using assume role credentials

### LINKED ISSUE #819: Add support for IAM assumed roles
Allows to connect to AWS Neptune DB/Graph via assume role credentials. A lot of companies use IAM role chaining as an extra layer of protecting to allow fine grained privileges to AWS resources.

## Related

- https://github.com/aws/graph-explorer/discussions/220#discussioncomment-12352912

ssheladiya:
## Description

Allows to connect to AWS Neptune DB/Graph via assume role credentials. A lot of companies use IAM role chaining as an extra layer of protecting to allow fine grained privileges to AWS resources. This PR allows to assume a specific role mentioned creating a connection.

- Adds a AWS Assume Role field while creating connection if IAM Auth is enabled
- If assumeRoleArn is present, credentials are generated by assuming that role

## Validation

- Validated the connections in local and kubernetes env
- Validated the connection without assuming role

image

## Related Issues

- Resolves #819

### Check List

      license.
- [x] I have added an entry in the `Changelog.md`.

### DIFF
--- a/packages/graph-explorer/src/core/connector.ts
+++ b/packages/graph-explorer/src/core/connector.ts
@@ -40,6 +40,7 @@ const activeConnectionSelector = equalSelector({
       "graphDbUrl",
       "awsAuthEnabled",
       "awsRegion",
+      "awsAssumeRoleArn",
       "fetchTimeoutMs",
       "nodeExpansionLimit",
     ] as (keyof ConnectionConfig)[];


kmcginnes:
@ssheladiya Thank you for the contribution. I took a quick glance at the code and it is very straightforward. I don't see any major blockers for this.

I'll need to carve out some time to give this a proper test and double check the code. But overall, this looks great.

kmcginnes:
I've added a related issue that points back to the discussion thread for traceability.

kmcginnes:
@ssheladiya One small thought looking at the screenshot in the description, I think it might make more sense to move the assume role input down one position. My thinking is that the region and service type fields are required, but the assumed role is optional. Making it last makes a bit more sense.

pnpm-lock.yaml
```diff
@@ -536,125 +539,190 @@ importers:
 packages:
 
   '@adobe/css-tools@4.4.1':
-    resolution: {integrity: sha512-12WGKBQzjUAI4ayyF4IAtfw2QR/IDoqk6jTddXDhtYTJF9ASmoE1zst7cVtP0aL/F1jUJL5r+JxKXKEgHNbEUQ==}
+    resolution: {integrity: sha512-12WGKBQzjUAI4ayyF4IAtfw2QR/IDoqk6jTddXDhtYTJF9ASmoE1zst7cVtP0aL/F1jUJL5r+JxKXKEgHNbEUQ==, tarball: https://registry.npmjs.intuit.com:443/artifactory/api/npm/npm-intuit/@adobe/css-tools/-/css-tools-4.4.1.tgz}
```
kmcginnes, line 542: @ssheladiya It looks like something has caused your local pnpm install to add `tarball` URL entries for everything. Perhaps [this documentation](https://pnpm.io/next/npmrc#lockfile-include-tarball-url) can lead you to the culprit. Or maybe you have a custom npm host setup. I'm not sure.

ssheladiya: @kmcginnes I have `lockfile-include-tarball-url=false` in my `~/.npmrc` but I guess its still adding tarball when I add a new aws-sts dependency. Any pointers on why this would happen? I found this issues - https://github.com/pnpm/pnpm/issues/6667. Sorry I am pretty new to nodeJS space 😟

ssheladiya: fixed in [c81f7d0](https://github.com/aws/graph-explorer/pull/818/commits/c81f7d08cd852c464dd3104814a48856e534f60f)

ssheladiya:
> @ssheladiya One small thought looking at the screenshot in the description, I think we should move the assume role input down one position. My thinking is that the region and service type fields are required, but the assumed role is optional. Making it last makes a bit more sense.

sure, I can make that change

ssheladiya:
> a82094b

fixed in [a82094b](https://github.com/aws/graph-explorer/pull/818/commits/a82094bd572d8c6247aab23ecc4f24318a409c92)

ssheladiya:
@kmcginnes Fixed all merge conflicts and changes.

kmcginnes:
I did some testing using an assumed IAM role that had Neptune read only access. It looks good from that perspective.

I need to verify with some internal folks to make sure there's no security concerns with this feature. This might take a bit of time (hopefully measured in days, but I don't want to promise anything).

packages/graph-explorer-proxy-server/src/node-server.ts
```diff
@@ -34,8 +45,19 @@ interface LoggerIncomingHttpHeaders extends IncomingHttpHeaders {
 
 app.use(requestLoggingMiddleware());
 
+// Function to check if the credentials are valid.
+function areCredentialsValid(creds: AwsCredentials): boolean {
+  return creds.expiration
+    ? new Date(creds.expiration).getTime() - Date.now() > 5 * 60 * 1000
```
kmcginnes, line 51: Is this time range something that is consistent for all users and situations? Perhaps there is a better way to check for expiration instead of checking against a hard coded timespan?

packages/graph-explorer-proxy-server/src/node-server.ts
```diff
@@ -44,13 +66,73 @@ async function getIAMHeaders(options: string | aws4.Request) {
...
+      };
+
+      return aws4.sign(options, {
+        accessKeyId: Credentials?.AccessKeyId,
+        secretAccessKey: Credentials?.SecretAccessKey,
+        ...(Credentials?.SessionToken && {
+          sessionToken: Credentials?.SessionToken,
+        }),
+      });
+    } catch (error) {
```
kmcginnes, line 119: Is it possible this `catch` will apply to more than the `STSClient`. For example, I think it is possible that the `aws4.sign()` function could throw. If that happens the error message below does not apply.

It might be worth combining the `fromNodeProviderChain()` credential path and the new STSClient credential path in to a separate function. That allows you to put the try/catch in there.

Then this function becomes something more like (pseudo code):

```js
function getIAMHeaders() {
  const creds = getCredentials();
  if (!creds) {
    throw new Error();
  }
  return aws4.sign(options, {
    accessKeyId: creds.AccessKeyId,
    secretAccessKey: creds.SecretAccessKey,
    ...(creds.SessionToken && {
      sessionToken: creds.SessionToken,
    }),
  });
}
```

packages/graph-explorer-proxy-server/src/node-server.ts
ssheladiya: Usually credentials do come with the expiration time and we would need to check against that time if the credentials are valid or not. One other way I could think of is it let it fail and then retry with the new credentials. But I think that is not very efficient approach. Thoughts?

kmcginnes: That's a good question. My immediate thought would be

why-not-both

It surely complicates the logic, but I think it will yield the best outcome.

packages/graph-explorer-proxy-server/src/node-server.ts
ssheladiya: fixed in - [dfc34dc](https://github.com/aws/graph-explorer/pull/818/commits/dfc34dc75fc80886c10f1951c2dc8299735ea69d)

kmcginnes: Nice. I'll take another look at this soon.

kmcginnes:
@ssheladiya Sorry for all the conflicting merges lately. You can keep the branch up to date each time if you like, or you can wait until I get the approval to move forward with this change.

ssheladiya:
> @ssheladiya Sorry for all the conflicting merges lately. You can keep the branch up to date each time if you like, or you can wait until I get the approval to move forward with this change.

No worries. I can understand, these things take time :)

ssheladiya:
@kmcginnes Now that I think of this, do you think it makes sense to abstract the assume role arn from the UI? I think if we can pass the assume role ARN settings through env vars, that would be better rather than adding the ARN on the UI. Users who wish to enable this feature can configure the proxy with necessary env vars. That would simplify the change and possibly would not require changes on the UI. WDYT?

triggan:
> do you think it makes sense to abstract the assume role arn from the UI? 

Honestly, I like that idea better.  I think anything that is a field or checkbox in the UI is something a user is going to look at and question what their configuration should be.  Since the use of assumed roles is not a common pattern for deploying graph-explorer (from what I've seen), I think it would be best to leave as an env var and have a user configure that upon container deployment.

kmcginnes:
I 100% agree on making this an environment value that only the proxy server consumes. It makes the configuration more manual, but it also makes it potentially more secure.

kmcginnes:
@ssheladiya Now that you've brought up this wonderful feature, we have some ideas on how to use it. Which means we'd like to get this merged in soon-ish (read within the next month).

Would you like me to take over this branch and finish up the work, or would you like to finish it up? I don't want to step on any toes.

guyelia:
> > do you think it makes sense to abstract the assume role arn from the UI?
> 
> Honestly, I like that idea better. I think anything that is a field or checkbox in the UI is something a user is going to look at and question what their configuration should be. Since the use of assumed roles is not a common pattern for deploying graph-explorer (from what I've seen), I think it would be best to leave as an env var and have a user configure that upon container deployment.

One question, though: if we keep the assume-role ARN as an environment variable, won’t that restrict each Graph Explorer instance to a single role? My goal is to run one Graph Explorer in our management account and have it connect to Neptune clusters in multiple account

kmcginnes:
@guyelia It certainly would. And you are correct, that is limiting. We will have to consider the options with that use case in mind. This will take a bit more thought.

PR #829, opened 2025-03-08
Update README.md to point to the right ECR docker image

talkain:
## Description

The current docs assume the base Docker registry is ECR, while for most users it will lead to DockerHub, where the image does not exist


kmcginnes (approved):
Good call. I mention this fact in a different part of the docs, but this is a better approach I think.

PR #918, opened 2025-05-08
Add configurable path for .env and defaultConnection.json files

mattiabertorello:
## Description
This MR adds support for configuring the folder path where the .env and defaultConnection.json files are read from and written to. This solves an issue when Graph Explorer is deployed in ECS with read-only root volume enabled, by allowing these configuration files to be stored in a writable location.

The changes introduce a new environment variable `CONFIGURATION_FOLDER_PATH` that specifies where these configuration files should be located. If not specified, the system falls back to the default client root path.

## Validation
The changes have been tested by:

Setting the CONFIGURATION_FOLDER_PATH environment variable to different locations
Verifying that the application correctly reads the .env and defaultConnection.json files from the specified location
Confirming that the application still works correctly when the variable is not set (using default paths)

## Related Issues
Fixes issue with ECS deployments using read-only root volumes where configuration files cannot be written to the default location.

### DIFF
--- a/packages/graph-explorer-proxy-server/src/env.ts
+++ b/packages/graph-explorer-proxy-server/src/env.ts
@@ -22,15 +22,16 @@ const EnvironmentValuesSchema = z.object({
   CONFIGURATION_FOLDER_PATH: z.coerce.string().default(clientRoot),
 });
 
-const defaultConnectionFolderPath = process.env.CONFIGURATION_FOLDER_PATH ?
-  process.env.CONFIGURATION_FOLDER_PATH : clientRoot
+const defaultConnectionFolderPath = process.env.CONFIGURATION_FOLDER_PATH
+  ? process.env.CONFIGURATION_FOLDER_PATH
+  : clientRoot;
 
 // Load environment variables from .env file.
 dotenv.config({
   path: [
     path.join(clientRoot, ".env.local"),
     path.join(clientRoot, ".env"),
-    path.join(defaultConnectionFolderPath, ".env")
+    path.join(defaultConnectionFolderPath, ".env"),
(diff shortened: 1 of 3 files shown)


kmcginnes:
@mattiabertorello Thank you for your submission. I like this idea.

I'm just curious if using environment values instead of `.env` file and `defaultConnection.json` file is better or worse for you? Is it a simply a preference for you to use files or is there some reason the environment values won't work for you?

I ask because I was considering deprecating the `defaultConnection.json` file in favor of environment values. But no decisions have been made yet. Your input could help.

mattiabertorello:
Sure @kmcginnes, We already have everything defined in environment variables, so it won’t require any change on our side. 
In general, it’s better to avoid writing to the file system unless necessary, as it introduces additional complexity on the infrastructure side.

kmcginnes:
@mattiabertorello Thanks for the data point. That will help us decide if removing the `defaultConnection.json` file option is viable in the future.

I agree on your point about writing to the file system. My long term plan was to move all the logic in `process-environment.sh` in to the proxy server TypeScript. That would remove a couple of layers of indirection in the Docker startup process. It would also make Docker execution and basic Node execution more consistent in behavior.

As for this PR, I'll carve out some time soon to test it out. I've looked through the code and I don't see any obvious issues.

kmcginnes:
I tested this out. The change looks good.

You'll need to run the formatter to make the GitHub Actions happy.

```
pnpm format
```

kmcginnes (dismissed):
Approved, other than the formatting. I'll merge after that is fixed.

kmcginnes (approved):

Issue #885, opened 2025-04-16
Table view column sorting and filtering is reset when shown after hiding


kmcginnes:
If you sort or filter data in the table view, then toggle the table view off/on the sorting and filtering are reset.

The table sort and filter options should instead be persistent for the session. Hiding and showing the table view should not affect the sort and filter options.

dwrth:
Hi, I would like to give this a go!

PR #1100, opened 2025-08-14
Persistent table filtering and sorting

dwrth:
## Description

- Added persistent sorting and filtering state for both nodes and edges tables
- Added a "Reset sorting and filters" button that appears when filters/sorts are active

## Validation

- Tested show/hide table view
- Tested reset by changing connections and letting current session expire

## Related Issues

Addresses #885

### Check List

      license.

### DIFF
--- a/packages/graph-explorer/src/components/Tabular/Tabular.tsx
+++ b/packages/graph-explorer/src/components/Tabular/Tabular.tsx
@@ -60,6 +60,7 @@ export const Tabular = <T extends object>(
     pageIndex = 0,
     pageSize = 10,
     onDataFilteredChange,
+    onColumnSortedChange,
     variant = "bordered",
     globalSearch,
     ...useTabularOptions
@@ -108,6 +109,12 @@ export const Tabular = <T extends object>(
     tableInstance.state.filters,
     tableInstance.filteredRows,
   ]);
+
+  useDeepCompareEffect(() => {
+    if (tableInstance.state.sortBy.length !== 0) {
+      onColumnSortedChange?.(tableInstance.state.sortBy);
+    }
+  }, [onColumnSortedChange, tableInstance.state.sortBy]);
 
   useImperativeHandle(ref, () => tabularInstance, [tabularInstance]);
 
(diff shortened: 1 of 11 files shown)


kmcginnes:
This is great! I should be able to get to a review next week.

packages/graph-explorer/src/components/Tabular/Tabular.tsx
```diff
@@ -100,16 +101,21 @@ export const Tabular = <T extends object>(
...
-    tableInstance.state.filters.length,
+    tableInstance.state.filters,
     tableInstance.filteredRows,
   ]);
 
+  useDeepCompareEffect(() => {
+    if (tableInstance.state.sortBy.length !== 0) {
+      onColumnSortedChange?.(tableInstance.state.sortBy);
+    }
+  }, [onColumnSortedChange, tableInstance.state.sortBy]);
```
kmcginnes:
```suggestion
  useDeepCompareEffect(() => {
    onColumnSortedChange?.(tableInstance.state.sortBy);
  }, [onColumnSortedChange, tableInstance.state.sortBy]);
```

That conditional will prevent saving an absence of a sort. So if a user clicks the sort arrow 3 times on a column header it will remove the sorting. If they then close and re-open the table it will restore the descending sort instead of the cleared sort.

Removing the conditional seems to fix the issue and doesn't seem to cause any unnecessary effect runs.

packages/graph-explorer/src/hooks/translations/gremlin-translations.json
```diff
@@ -63,7 +63,8 @@
     "source-type": "Source Label",
     "target-id": "Target Id",
     "target-type": "Target Label",
-    "edges-placeholder": "Add a node in the Graph View to see its connections"
+    "edges-placeholder": "Add a node in the Graph View to see its connections",
+    "reset-sorting-and-filters": "Reset sorting and filters"
```
kmcginnes: I'd suggest simply putting the text inline in the React code. The translation layer is just used for differences between property graph and RDF database terminology, not actual written languages. And this label is the same across all three query languages.

dwrth: Ahh got it! I was already wondering about that, thanks!

kmcginnes (approved):

kmcginnes:
Thanks again @dwrth! I'll mention your contribution in the release notes.

Issue #1056, opened 2025-07-16
Node icon does not reflect current node color


kmcginnes:
The icon displayed inside the node in the canvas does not update properly when the color changes for the node.

The color is correct after a refresh.

## Potential solution

The issue seems to be that the icon is cached, but the cache is not busted when the color changes. We could add busting logic or add the color as part of the cache key.

dwrth:
Hey there, I would like to help out with this one!

PR #1103, opened 2025-08-16
Fix node icon color change

dwrth:
## Description

This PR fixes the node color on the canvas not updating when it is changed in the `NodeStyleDialog`. This is achieved by adding the icon color to the cache key. 

To avoid too many updates and potential react errors from happening due to the frequent cache updates I used the `useDebounceValue` hook which also has the added benefit of reducing the input lag when dragging inside the color picker.

So both color pickers are aligned, I also added the debouncing logic for the border color.

Additionally I added the border styling to the icon preview / icon upload button inside the modal

## Validation

Tested by selecting new colors for node icons and border colors. Stress tested by rapidly changing colors.

## Related Issues

Fixes: #1056

### Check List

      license.

### DIFF
--- a/packages/graph-explorer/src/modules/GraphViewer/renderNode.tsx
+++ b/packages/graph-explorer/src/modules/GraphViewer/renderNode.tsx
@@ -19,8 +19,9 @@ export async function renderNode(
   }
 
   // To avoid multiple requests, cache icons under the same URL
-  if (ICONS_CACHE.get(vtConfig.iconUrl)) {
-    return ICONS_CACHE.get(vtConfig.iconUrl);
+  const cacheKey = `${vtConfig.iconUrl}::${vtConfig.color || "#128EE5"}`;
+  if (ICONS_CACHE.get(cacheKey)) {
+    return ICONS_CACHE.get(cacheKey);
   }
 
   try {
@@ -33,7 +34,7 @@ export async function renderNode(
     iconText = encodeSvg(iconText);
 
     // Save to the cache
-    ICONS_CACHE.set(vtConfig.iconUrl, iconText);
+    ICONS_CACHE.set(cacheKey, iconText);
     return iconText;
   } catch (error) {
     // Ignore the error and move on
--- a/packages/graph-explorer/src/modules/NodesStyling/NodeStyleDialog.tsx
+++ b/packages/graph-explorer/src/modules/NodesStyling/NodeStyleDialog.tsx
@@ -24,6 +24,7 @@ import {
   useVertexStyling,
 } from "@/core/StateProvider/userPreferences";
 import useTranslations from "@/hooks/useTranslations";
+import { useDebounceValue, usePrevious } from "@/hooks";
 import { LINE_STYLE_OPTIONS } from "./lineStyling";
 import { NODE_SHAPE } from "./nodeShape";
 import modalDefaultStyles from "./SingleNodeStylingModal.style";
@@ -33,6 +34,7 @@ import {
 } from "@/utils/constants";
 import { cn } from "@/utils";
 import { atom, useAtom } from "jotai";
+import { useEffect, useState } from "react";
 
 export const customizeNodeTypeAtom = atom<string | undefined>(undefined);
 
@@ -94,6 +96,33 @@ function Content({ vertexType }: { vertexType: string }) {
   const { vertexStyle, setVertexStyle, resetVertexStyle } =
     useVertexStyling(vertexType);
   const displayConfig = useDisplayVertexTypeConfig(vertexType);
+
+  const [localColors, setLocalColors] = useState({
+    color: vertexStyle?.color || "#17457b",
+    borderColor: vertexStyle?.borderColor || "#17457b",
+  });
+
+  const debouncedColors = useDebounceValue(localColors, 300);
+  const prevColors = usePrevious(debouncedColors);
+
+  useEffect(() => {
+    if (prevColors === null || prevColors === debouncedColors) {
+      return;
+    }
+    void setVertexStyle({
+      color: debouncedColors.color,
+      borderColor: debouncedColors.borderColor,
+    });
+  }, [debouncedColors, prevColors, setVertexStyle]);
+
+  useEffect(() => {
+    if (vertexStyle) {
+      setLocalColors({
+        color: vertexStyle.color || "#17457b",
+        borderColor: vertexStyle.borderColor || "#17457b",
+      });
+    }
+  }, [vertexStyle]);
 
   const selectOptions = (() => {
     const options = displayConfig.attributes.map(attr => ({
@@ -179,11 +208,12 @@ function Content({ vertexType }: { vertexType: string }) {
             asChild
           >
             <IconButton
-              variant="filled"
-              className="text-text-primary hover:text-text-primary group rounded-full border-0 bg-transparent p-0 hover:cursor-pointer hover:bg-gray-200"
+              variant="default"
+              className="group rounded-full"
+              style={vertexStyle}
               icon={
                 <>
-                  <div className="hidden group-hover:flex">
+                  <div className="hidden p-2 group-hover:flex">
                     <UploadIcon />
                   </div>
                   <VertexSymbol
@@ -203,8 +233,10 @@ function Content({ vertexType }: { vertexType: string }) {
           <ColorInput
             label="Color"
             labelPlacement="inner"
-            color={vertexStyle?.color || "#17457b"}
-            onChange={(color: string) => setVertexStyle({ color })}
+            color={localColors.color}
+            onChange={(color: string) =>
+              setLocalColors(prev => ({ ...prev, color }))
+            }
           />
           <InputField
             label="Background Opacity"
@@ -225,8 +257,10 @@ function Content({ vertexType }: { vertexType: string }) {
           <ColorInput
             label="Border Color"
             labelPlacement="inner"
-            color={vertexStyle?.borderColor || "#17457b"}
-            onChange={(color: string) => setVertexStyle({ borderColor: color })}
+            color={localColors.borderColor}
+            onChange={(color: string) =>
+              setLocalColors(prev => ({ ...prev, borderColor: color }))
+            }
           />
           <InputField
             label="Border Width"


packages/graph-explorer/src/modules/NodesStyling/NodeStyleDialog.tsx
```diff
@@ -95,6 +97,33 @@ function Content({ vertexType }: { vertexType: string }) {
...
+  }, [debouncedColors, prevColors, setVertexStyle]);
+
+  useEffect(() => {
+    if (vertexStyle) {
+      setLocalColors({
+        color: vertexStyle.color || "#17457b",
+        borderColor: vertexStyle.borderColor || "#17457b",
+      });
+    }
+  }, [vertexStyle]);
```
kmcginnes: I'm feeling a bit nervous about leaning on useEffect, useDebounce, and usePrevious here. I did some experimentation with using `useDeferredValue()` instead with some success. But I had some blinking in the nodes as they update their rendering.

So then I added a couple of small Tanstack Query functions to manage the SVG download and the SVG text patching logic separately. This replaces the `Map` as a cache, which could grow infinitely, and allows the SVG download to happen only once per URL and the color update to happen every time it changes.

This seemed to work well in my testing, but probably needs a bit more testing and polish before calling it good. I've publish my testing branch and below is the comparison. See if any of my ideas are valid in your testing. They may not be.

- https://github.com/dwrth/graph-explorer/compare/fix/node-icon-color-change...kmcginnes:graph-explorer:node-icon-color-alt

dwrth: I like your approach with using Tanstack Query a lot! There is definitely still some input lag happening and some flickering but I think it's pointing me in the right direction, thanks!

Will continue working on this later today. I hope there is no big time pressure on this now that it has been added to the next release milestone?

kmcginnes: No, don't read in to that. That was aspirational by me. We will not hold the release for this.

In fact, I'll go ahead and remove it from the milestone since we are still working on it.

dwrth: After a bunch of trial and error I have hit a wall when it comes tho the flicker that happens on color changes. I am pretty sure that this is coming directly from cytoscape as the package also performs a fetch on every change and does not seem to keep the previous icon while the new one is loading.

https://github.com/user-attachments/assets/26faf078-10bc-4a41-b921-f08de81b74fe

Any guidance on what we should do about this?

kmcginnes: Thank you for giving it a try. I noticed that too with my branch. 

I'll take a look later today after finalizing the release. Hopefully this ends up being a solvable problem. If not, we still have your original approach to fall back on.

kmcginnes: I spent a couple of hours on it and I don't think we can do better than we already have. Both of our branches have the flickering, but yours is less pronounced since the debounce delays the update.

So we just need to choose which option to go with.

1. My approach using `useDeferredValue()` for faster color change rendering, but more flickering
2. Your approach using debounced values for slower color change rendering and less flickering

I'm leaning toward my approach since it is simpler from the code's perspective with less indirection. But the UX is both better and worse at the same time.

Feel free to convince me one way or another, but if your asking me, I say we go with option 1.

dwrth: I agree with taking route 1. here. Makes no sense to have all settings on the platform work instantly and then have this simple color change have such a delay. 

I've gone ahead and implemented the changes from your branch as well as adjusted the tests to work with the new caching logic.

packages/graph-explorer/src/modules/GraphViewer/renderNode.test.ts
```diff
@@ -3,23 +3,31 @@
...
+
+// Mock the logger module
+vi.mock("@/utils", () => ({
+  logger: {
+    debug: vi.fn(),
+    error: vi.fn(),
+    log: vi.fn(),
+    warn: vi.fn(),
+  },
+}));
```
kmcginnes: You shouldn't need to mock the logger. It is mocked in the global setup:

https://github.com/aws/graph-explorer/blob/741be8d8144d15bc101857adc4367af4f5a44ddb/packages/graph-explorer/src/setupTests.ts#L85-L94

kmcginnes: I think the code that was here before predated that global mock, and maybe the `logger` utility.

packages/graph-explorer/src/modules/NodesStyling/NodeStyleDialog.tsx
```diff
@@ -233,9 +205,9 @@ function Content({ vertexType }: { vertexType: string }) {
           <ColorInput
             label="Color"
             labelPlacement="inner"
-            color={localColors.color}
+            color={vertexStyle?.color || "#17457b"}
             onChange={(color: string) =>
-              setLocalColors(prev => ({ ...prev, color }))
+              startTransition(() => setVertexStyle({ color }))
```
kmcginnes: I remember I added the `startTransition` wrapper, but I don't know if I ever tested to see if it was helping the responsiveness. If it turns out it is not providing any benefit, let's remove it. If it is improving the responsiveness, then let's add it to the other color inputs.

kmcginnes:
@dwrth After further testing, prototyping, and consideration, I'm voting that we take the problems with the query cache as I outlined above. There doesn't seem to be any flashing when the `gcTime` expires.

So finalize any changes you want to make to this PR and let me know when you are done. I'll give it a look. I will be off the next two days, so it might be Monday before I get a chance.

dwrth:
@kmcginnes Only ended up with one small change as I noticed that the svg ended up being wrapped in another svg tag.

Previously the output would be:

``` html
<?xml version="1.0" encoding="UTF-8"?><!DOCTYPE svg>
<svg width="24" height="24" xmlns="http://www.w3.org/2000/svg">
  <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none" stroke="#23387c" stroke-width="2" stroke-linecap="round" stroke-linejoin="round" class="lucide lucide-users"><path d="M16 21v-2a4 4 0 0 0-4-4H6a4 4 0 0 0-4 4v2"/><circle cx="9" cy="7" r="4"/><path d="M22 21v-2a4 4 0 0 0-3-3.87"/><path d="M16 3.13a4 4 0 0 1 0 7.75"/></svg>
</svg>
```

or:

``` html
<?xml version="1.0" encoding="UTF-8"?><!DOCTYPE svg>
<svg width="24" height="24" xmlns="http://www.w3.org/2000/svg">
  <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none">
  <path fill="#566679" d="M6.11 5.516c.246-.36.39-.797.39-1.266C6.5 3.007 5.493 2 4.25 2S2 3.007 2 4.25 3.007 6.5 4.25 6.5c.32 0 .623-.066.898-.186l3.575 4.175c-.41.46-.692 1.038-.787 1.678l-1.63.17C5.957 11.549 5.168 11 4.25 11 3.007 11 2 12.007 2 13.25s1.007 2.25 2.25 2.25c1.132 0 2.069-.836 2.227-1.924l1.514-.158c.179.738.61 1.377 1.196 1.823l-1.05 2.292c-.125-.022-.255-.033-.387-.033-1.243 0-2.25 1.007-2.25 2.25S6.507 22 7.75 22 10 20.993 10 19.75c0-.661-.285-1.256-.74-1.667l1.05-2.293c.268.072.55.11.84.11.713 0 1.373-.23 1.91-.62l3.202 2.916c-.167.314-.262.673-.262 1.054 0 1.243 1.007 2.25 2.25 2.25s2.25-1.007 2.25-2.25S19.493 17 18.25 17c-.407 0-.789.108-1.118.297l-3.221-2.932c.31-.498.489-1.085.489-1.715 0-.043 0-.086-.003-.128l3.39-.672c.385.687 1.12 1.15 1.963 1.15 1.243 0 2.25-1.007 2.25-2.25S20.993 8.5 19.75 8.5c-1.203 0-2.186.944-2.247 2.132l-3.394.673c-.258-.567-.674-1.046-1.191-1.383l.639-1.43c.063.005.128.008.193.008C14.993 8.5 16 7.493 16 6.25S14.993 4 13.75 4 11.5 5.007 11.5 6.25c0 .731.349 1.38.889 1.792l-.632 1.415c-.197-.038-.4-.057-.607-.057-.513 0-.999.119-1.43.33L6.11 5.517z"/>
</svg>
</svg>
```

Now it is:

``` html
<?xml version="1.0" encoding="UTF-8"?><!DOCTYPE svg>
<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none" stroke="#2b4eb7" stroke-width="2" stroke-linecap="round" stroke-linejoin="round" class="lucide lucide-users"><path d="M16 21v-2a4 4 0 0 0-4-4H6a4 4 0 0 0-4 4v2"/><circle cx="9" cy="7" r="4"/><path d="M22 21v-2a4 4 0 0 0-3-3.87"/><path d="M16 3.13a4 4 0 0 1 0 7.75"/></svg>
```

or:

``` html
<?xml version="1.0" encoding="UTF-8"?><!DOCTYPE svg>
<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none">
  <path fill="#4070a6" d="M6.11 5.516c.246-.36.39-.797.39-1.266C6.5 3.007 5.493 2 4.25 2S2 3.007 2 4.25 3.007 6.5 4.25 6.5c.32 0 .623-.066.898-.186l3.575 4.175c-.41.46-.692 1.038-.787 1.678l-1.63.17C5.957 11.549 5.168 11 4.25 11 3.007 11 2 12.007 2 13.25s1.007 2.25 2.25 2.25c1.132 0 2.069-.836 2.227-1.924l1.514-.158c.179.738.61 1.377 1.196 1.823l-1.05 2.292c-.125-.022-.255-.033-.387-.033-1.243 0-2.25 1.007-2.25 2.25S6.507 22 7.75 22 10 20.993 10 19.75c0-.661-.285-1.256-.74-1.667l1.05-2.293c.268.072.55.11.84.11.713 0 1.373-.23 1.91-.62l3.202 2.916c-.167.314-.262.673-.262 1.054 0 1.243 1.007 2.25 2.25 2.25s2.25-1.007 2.25-2.25S19.493 17 18.25 17c-.407 0-.789.108-1.118.297l-3.221-2.932c.31-.498.489-1.085.489-1.715 0-.043 0-.086-.003-.128l3.39-.672c.385.687 1.12 1.15 1.963 1.15 1.243 0 2.25-1.007 2.25-2.25S20.993 8.5 19.75 8.5c-1.203 0-2.186.944-2.247 2.132l-3.394.673c-.258-.567-.674-1.046-1.191-1.383l.639-1.43c.063.005.128.008.193.008C14.993 8.5 16 7.493 16 6.25S14.993 4 13.75 4 11.5 5.007 11.5 6.25c0 .731.349 1.38.889 1.792l-.632 1.415c-.197-.038-.4-.057-.607-.057-.513 0-.999.119-1.43.33L6.11 5.517z"/>
</svg>
```

packages/graph-explorer/src/modules/NodesStyling/NodeStyleDialog.tsx
```diff
@@ -179,11 +179,12 @@ function Content({ vertexType }: { vertexType: string }) {
             asChild
           >
             <IconButton
-              variant="filled"
-              className="text-text-primary hover:text-text-primary group rounded-full border-0 bg-transparent p-0 hover:cursor-pointer hover:bg-gray-200"
+              variant="default"
+              className="group rounded-full"
+              style={vertexStyle}
```
kmcginnes:
```suggestion
              variant="default"
              className="group rounded-full"
```

Great change! For the upload icon, let's let the default button color shine through (black). So I suggest removing the `style={vertexStyle}`

packages/graph-explorer/src/modules/NodesStyling/NodeStyleDialog.tsx
```diff
@@ -179,11 +179,12 @@ function Content({ vertexType }: { vertexType: string }) {
...
               icon={
                 <>
-                  <div className="hidden group-hover:flex">
+                  <div className="hidden p-2 group-hover:flex">
```
kmcginnes:
```suggestion
                  <div className="hidden group-hover:flex">
```

Let's keep the default `IconButton` padding

packages/graph-explorer/src/modules/GraphViewer/renderNode.tsx
```diff
@@ -52,9 +73,7 @@ export async function renderNode(
  */
 function embedSvgInsideCytoscapeSvgWrapper(svgContent: string): string {
   return `<?xml version="1.0" encoding="UTF-8"?><!DOCTYPE svg>
-<svg width="24" height="24" xmlns="http://www.w3.org/2000/svg">
-  ${svgContent}
-</svg>`;
+${svgContent}`;
```
kmcginnes, line 76: I manually tested the change to remove the extraneous wrapper `svg` element. I'm not sure why I thought it was needed in the past. But not having it seems to work with all shapes and sizes of SVG images. So thank you for catching that.

However, the tests do need to be updated to account for this change.

kmcginnes (changes requested):
We are really close on this one. Just a few changes needed, then I think we can merge.

packages/graph-explorer/src/modules/GraphViewer/renderNode.tsx
dwrth: I have also tested on all major browsers and could not find anything wrong with this. Docs of cytoscape also only mention the `<?xml version="1.0" encoding="UTF-8"?><!DOCTYPE svg>` part so I think we're safe!

Tests are updated now :)

PR #1192, opened 2025-09-17
added healthcheck in Dockerfile

### LINKED ISSUE #1187: Add health check to Dockerfile
Add the healthcheck parameter to our main Docker file. This will allow Docker to check whether the container is healthy while running.

* [Dockerfile reference](https://docs.docker.com/reference/dockerfile/#healthcheck)

aravindan888:
Fixes #1187 
Added a healthcheck to the Dockerfile by installing curl-minimal and configuring Docker to periodically check the service endpoint, ensuring the container is marked healthy only when it’s responsive, also github action wokflow for the build docker file ran successfully so its fine to merge

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ FROM amazonlinux:2023 AS base
 ENV NODE_VERSION=24.4.0
 
 RUN yum update -y && \
-    yum install -y tar xz openssl curl && \
+    yum install -y tar xz openssl curl-minimal && \
     ARCH=$(uname -m) && \
     if [ "$ARCH" = "x86_64" ]; then NODE_ARCH="x64"; \
     elif [ "$ARCH" = "aarch64" ]; then NODE_ARCH="arm64"; \
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ FROM amazonlinux:2023 AS base
 ENV NODE_VERSION=24.4.0
 
 RUN yum update -y && \
-    yum install -y tar xz openssl && \
+    yum install -y tar xz openssl curl && \
     ARCH=$(uname -m) && \
     if [ "$ARCH" = "x86_64" ]; then NODE_ARCH="x64"; \
     elif [ "$ARCH" = "aarch64" ]; then NODE_ARCH="arm64"; \
@@ -20,18 +20,6 @@ ARG NEPTUNE_NOTEBOOK
 
 ENV NEPTUNE_NOTEBOOK=$NEPTUNE_NOTEBOOK
 ENV HOME=/graph-explorer
-
-# Conditionally set the following environment values using +/- variable expansion
-# https://docs.docker.com/reference/dockerfile/#environment-replacement
-# 
-# If NEPTUNE_NOTEBOOK value is set then
-#   - GRAPH_EXP_ENV_ROOT_FOLDER = /proxy/9250/explorer
-#   - PROXY_SERVER_HTTP_PORT    = 9250
-#   - LOG_STYLE                 = cloudwatch
-# Else the values are the defaults
-#   - GRAPH_EXP_ENV_ROOT_FOLDER = /explorer
-#   - PROXY_SERVER_HTTP_PORT    = 80
-#   - LOG_STYLE                 = default
 
 ENV GRAPH_EXP_ENV_ROOT_FOLDER=${NEPTUNE_NOTEBOOK:+/proxy/9250/explorer}
 ENV GRAPH_EXP_ENV_ROOT_FOLDER=${GRAPH_EXP_ENV_ROOT_FOLDER:-/explorer}
@@ -57,4 +45,8 @@ RUN pnpm install && \
 EXPOSE 443
 EXPOSE 80
 EXPOSE 9250
+
+HEALTHCHECK --interval=30s --timeout=5s --start-period=10s --retries=3 \
+  CMD curl -f http://localhost:${PROXY_SERVER_HTTP_PORT}/ || exit 1
+
 ENTRYPOINT ["./docker-entrypoint.sh"]


Dockerfile
```diff
@@ -3,7 +3,7 @@ FROM amazonlinux:2023 AS base
 ENV NODE_VERSION=24.4.0
 
 RUN yum update -y && \
-    yum install -y tar xz openssl && \
+    yum install -y tar xz openssl curl-minimal && \
```
kmcginnes: You'll need to double check me on this, but I believe the amazonlinux:2023 image contains `curl` already.

Dockerfile
```diff
@@ -21,18 +21,6 @@ ARG NEPTUNE_NOTEBOOK
...
-# 
-# If NEPTUNE_NOTEBOOK value is set then
-#   - GRAPH_EXP_ENV_ROOT_FOLDER = /proxy/9250/explorer
-#   - PROXY_SERVER_HTTP_PORT    = 9250
-#   - LOG_STYLE                 = cloudwatch
-# Else the values are the defaults
-#   - GRAPH_EXP_ENV_ROOT_FOLDER = /explorer
-#   - PROXY_SERVER_HTTP_PORT    = 80
-#   - LOG_STYLE                 = default
-
```
kmcginnes: Please don't remove the documentation

Dockerfile
```diff
@@ -57,4 +45,8 @@ RUN pnpm install && \
 EXPOSE 443
 EXPOSE 80
 EXPOSE 9250
+
+HEALTHCHECK --interval=30s --timeout=5s --start-period=10s --retries=3 \
+  CMD curl -f http://localhost:${PROXY_SERVER_HTTP_PORT}/ || exit 1
```
kmcginnes: The `/status` endpoint is what we should hit to test the server is up.

kmcginnes: Also, it seems the server url may be a complicated endeavor. We have multiple possible endpoints:

| Config | URL |
|----------|-----|
| HTTP | `http://localhost:${PROXY_SERVER_HTTP_PORT}/status` |
| HTTPS | `https://localhost:${PROXY_SERVER_HTTPS_PORT}/status` |
| Neptune Notebook | `http://localhost:9250/status` |

Dockerfile
```diff
@@ -3,7 +3,9 @@ FROM amazonlinux:2023 AS base
 ENV NODE_VERSION=24.4.0
 
 RUN yum update -y && \
-    yum install -y tar xz openssl && \
+    # Amazon Linux 2023 base image is minimal and does not include curl by default,
+    # so we explicitly install curl here to ensure availability.
+    yum install -y tar xz openssl curl && \
```
kmcginnes, line 8: I did a quick test to see if curl was available in the base image. It seems like it is.

```dockerfile
FROM amazonlinux:2023
CMD ["curl", "--version"]
```

Tested it with:

```sh
docker build -t curl-test .
docker run curl-test
```

Which produces the output:

```sh
curl 8.11.1 (aarch64-amazon-linux-gnu) libcurl/8.11.1 OpenSSL/3.2.2 zlib/1.2.11 libidn2/2.3.2 libpsl/0.21.5 nghttp2/1.59.0
Release-Date: 2024-12-11
Protocols: file ftp ftps http https ipfs ipns
Features: alt-svc AsynchDNS GSS-API HSTS HTTP2 HTTPS-proxy IDN IPv6 Kerberos Largefile libz PSL SPNEGO SSL threadsafe UnixSockets
```

Dockerfile
```diff
@@ -38,6 +40,7 @@ ENV GRAPH_EXP_ENV_ROOT_FOLDER=${GRAPH_EXP_ENV_ROOT_FOLDER:-/explorer}
 
 ENV PROXY_SERVER_HTTP_PORT=${NEPTUNE_NOTEBOOK:+9250}
 ENV PROXY_SERVER_HTTP_PORT=${PROXY_SERVER_HTTP_PORT:-80}
+ENV PROXY_SERVER_HTTPS_PORT=443
```
kmcginnes, line 43: No need to define the HTTPS default port here. It will be defaulted to 443 if not set explicitly

Dockerfile
```diff
@@ -54,7 +57,12 @@ RUN pnpm install && \
     chmod a+x ./process-environment.sh && \
     chmod a+x ./docker-entrypoint.sh
 
+# Expose ports for HTTP, HTTPS, and Neptune Notebook proxy
 EXPOSE 443
 EXPOSE 80
 EXPOSE 9250
+
+HEALTHCHECK --interval=30s --timeout=5s --start-period=10s --retries=3 \
+  CMD curl -fsSL "http://localhost:${PROXY_SERVER_HTTP_PORT}/status" || exit 1
```
kmcginnes, line 66: This still does not take in to account the different deployment scenarios I pointed out last time. Perhaps I did not give enough information about the issue.

Administrators of Graph Explorer have flexible deployment options. They can:

* Choose whether to host over HTTP or HTTPS with a self signed certificate
* Choose to override the default HTTP port (when using HTTP)
* Choose to override the default HTTPS port (when using HTTPS)
* Deploy to AWS Neptune Notebooks, which use Jupyter proxy to serve the site which requires HTTP, port 9250, and a base path of `/proxy/9250`

This means the curl command you have above will only work in the HTTP scenario. All the other scenarios will break and report unhealthy to Docker.

kmcginnes (changes requested):

Dockerfile
aravindan888: don't need to install curl manually anymore

Dockerfile
aravindan888: For the healthcheck, shall we mention the URL in the Dockerfile and also add it to .env so it can be configured?

kmcginnes: The user can already configure it, but it is a collection of options mapped together.

I tried seeing what the LLMs would recommend, and they offered a decent approach. Please don't take this as the solution. You'll need to test this out in all the different scenarios to prove it works properly.

I'm already a bit suspect of the check against `"true"`. I'm not sure if that is case insensitive, but we would need it to be.

```Dockerfile
# Dynamic healthcheck URL construction
HEALTHCHECK --interval=30s --timeout=10s --start-period=5s --retries=3 \
  CMD protocol=$([ "$PROXY_SERVER_HTTPS_CONNECTION" = "true" ] && echo "https" || echo "http"); \
      port=$([ "$PROXY_SERVER_HTTPS_CONNECTION" = "true" ] && echo "$PROXY_SERVER_HTTPS_PORT" || echo "$PROXY_SERVER_HTTP_PORT"); \
      base_path=$([ "$NEPTUNE_NOTEBOOK" = "true" ] && echo "/proxy/9250" || echo ""); \
      curl -f ${protocol}://localhost:${port}${base_path}/status || exit 1
```

Issue #1408, opened 2026-01-05
Filter and sort inside data explorer


kmcginnes:
Allow the user to perform filtering and sorting in the data explorer view. The sorting and filtering should be executed on the server side.

## Related Issues

- #1409

PR #1490, opened 2026-01-29
Add filter and sort in data explorer

dwrth:
## Description

This PR adds filtering and sorting to Data Explorer.

At the moment only for Gremlin.

## Related Issues

Addresses #1408

### Check List

      license.


kmcginnes:
In the original issue #1408 you asked for some guidance and I said I'd provide my general plan for implementing this feature. At the very least, I'd split this in to two parts detailed below. As for the UI, something similar to what exists in the data view in the graph explorer view.

### Update keyword search to allow for multiple filters

The keyword search query has almost everything needed to make filtering work. But it only has inputs for a single filter, since that was all that was needed for the search sidebar UI.

The expand neighbors query allows for multiple filters to be applied, so I was going to use that as an example of how to apply multiple filters in each query language. The keyword search and expand neighbors queries are fairly similar, so the translation shouldn't be too difficult.

### Update keyword search to allow for sorting by a single property

Sorting can have some pretty large impacts on the performance of a query, so this part of the feature may not even make it in. But if the default state performance is unaffected, and the only performance hit is when the user chooses to sort, then I think it is ok.

Each query language has slightly different sorting keywords, but usually some form of `order by`. I think I've removed all examples of sorting from the project now. There used to be some queries that did sorting, but it always caused issues.

So the sorting must be optional in the keyword search parameters.

kmcginnes:
Marking this as draft until we can both circle back around to this.

PR #1548, opened 2026-03-02
fix(schema): persist empty edgeConnections array when database has no edges

### LINKED ISSUE #1547: Schema sync sets edgeConnections to undefined when database has no nodes or edges
## Description

When a database has no nodes or edges, schema sync results in `edgeConnections` being `undefined` instead of an empty array. This is incorrect because `undefined` means edge connections have not been discovered yet, while `[]` means the query succeeded but found no connections.

## Expected Behavior

`edgeConnections` should be set to `[]` when the database is empty, indicating a successful discovery with no results.

## Actual Behavior

`edgeConnections` is set to `undefined`, which is indistinguishable from the case where edge connection discovery has not yet run or has failed.

## Related Issues

- Parent #685

abhu85:
## Summary

- Fix schema sync setting `edgeConnections` to `undefined` instead of `[]` when database has no nodes/edges
- Add test coverage for the semantic distinction between undefined (not discovered) and empty array (no connections found)

## Problem

When a database has no edges, the edge connections query was returning an empty array but not persisting it to the store. This meant `edgeConnections` remained `undefined`, which is indistinguishable from the case where edge connection discovery has not yet run or has failed.

## Root Cause

In `edgeConnectionsQuery.ts`, when `sortedEdgeTypes.length === 0`, the function returned `[]` early without calling `store.set(setEdgeConnectionsAtom, [])`. This meant the store was never updated, leaving `edgeConnections` as `undefined`.

## Solution

Added a call to persist the empty array to the store even when there are no edge types to query:

```typescript
if (sortedEdgeTypes.length === 0) {
  // Still persist empty array to distinguish "no connections found" from "not yet discovered"
  store.set(setEdgeConnectionsAtom, []);
  return [];
}
```

## Tests Run

```bash
pnpm vitest run packages/graph-explorer/src/connector/queries/edgeConnectionsQuery.test.ts
# 11 tests pass

pnpm vitest run packages/graph-explorer/src/hooks/useSchemaSync.test.ts
# 18 tests pass

pnpm vitest run packages/graph-explorer/src/connector/queries/schemaSyncQuery.test.ts
# 13 tests pass

pnpm vitest run packages/graph-explorer/src/core/StateProvider/schema.test.ts
# 37 tests pass
```

## Test Plan

- [x] Verify that when a database has no edges, `edgeConnections` is set to `[]` (not `undefined`)
- [x] Verify that existing edge connections behavior is preserved when edges exist
- [x] Verify that the semantic distinction between "not discovered" and "no connections" is maintained

Fixes #1547

---
Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/packages/graph-explorer/src/connector/queries/edgeConnectionsQuery.ts
+++ b/packages/graph-explorer/src/connector/queries/edgeConnectionsQuery.ts
@@ -34,6 +34,8 @@ export function edgeConnectionsQuery(edgeTypes: EdgeType[]) {
       const store = getStore(meta);
 
       if (sortedEdgeTypes.length === 0) {
+        // Still persist empty array to distinguish "no connections found" from "not yet discovered"
+        store.set(setEdgeConnectionsAtom, []);
         return [];
       }
 
--- a/packages/graph-explorer/src/connector/queries/edgeConnectionsQuery.test.ts
+++ b/packages/graph-explorer/src/connector/queries/edgeConnectionsQuery.test.ts
@@ -221,40 +221,32 @@ describe("edgeConnectionsQuery", () => {
     ]);
   });
 
-  it("should not update store when query returns early with empty edge types", async () => {
+  it("should set empty array in store when query has no edge types", async () => {
     const explorer = new FakeExplorer();
     const state = new DbState(explorer);
     const store = getAppStore();
+
+    // Explicitly set edgeConnections to undefined to test the fix
+    state.activeSchema.edgeConnections = undefined;
     state.applyTo(store);
 
-    // Set up initial edge connection
-    const sourceType = createVertexType("Person");
-    const targetType = createVertexType("Company");
-    const source = createTestableVertex().with({ types: [sourceType] });
-    const target = createTestableVertex().with({ types: [targetType] });
-    const edge = createTestableEdge().withSource(source).withTarget(target);
-    explorer.addTestableEdge(edge);
+    // Verify initial state has undefined edge connections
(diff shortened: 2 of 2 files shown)


kmcginnes:
Thanks @abhu85. I also had a branch to fix this. I'll compare our approaches and choose one as this needs to be fixed today, if possible.

In the future, if you wouldn't mind posting a comment on issues if you plan to work on them. Then I can assign things to you properly so you'll be the only one working on it. I need to get better about putting that disclaimer on all the issues I create, as it should be there. So that is certainly on me.

Either way, I love your contributions. Keep it up.

packages/graph-explorer/src/connector/queries/edgeConnectionsQuery.ts
```diff
@@ -34,6 +34,8 @@ export function edgeConnectionsQuery(edgeTypes: EdgeType[]) {
       const store = getStore(meta);
 
       if (sortedEdgeTypes.length === 0) {
+        // Still persist empty array to distinguish "no connections found" from "not yet discovered"
+        store.set(setEdgeConnectionsAtom, []);
```
kmcginnes, line 38: Ah ok. You started where I started. Unfortunately, this is not enough because `edgeConnectionsQuery` will be called with an empty array when there is no schema data (due to failure or just because it hasn't been run yet). This would write the empty edge connections to the stored schema and never run the edge connection query again until the user asks for it.

I'll likely end up taking my approach, once I finalize it. It tries to account for this scenario, but it just makes the code ever more complex. So I'm investigating ways to simplify.

PR #1739, opened 2026-04-29
Break circular dependency in storageAtoms

kmcginnes:
## Description

`storageAtoms.ts` imported `defaultUserLayout` from `./index` (the barrel), which triggered loading all barrel re-exports. Several of those modules import storage atoms back through `@/core`, creating a circular dependency. Because `storageAtoms` uses top-level `await`, the atoms were `undefined` when the circular modules evaluated, causing 6 flaky test failures in `userLayout.test.ts` that depended on module loading order.

- Replace the barrel import in `storageAtoms.ts` with direct imports from source modules
- Extract types and layout constants into `userLayoutDefaults.ts` as a pure leaf module with no imports, fully decoupling `storageAtoms` from `userLayout`
- Add regression tests for `storageAtoms` (atoms defined, correct defaults, write/read roundtrip)
- Add tests for `useTableViewSize` and `useSidebarSize` hooks
- Fix pre-existing `||` vs `??` bug in `useTableViewSize` where a height of 0 silently fell back to the default
- Fix pre-existing dead code (`|| {}` guard), typos, and loose test assertions

## Validation

- All 153 test files pass (1654 tests), including the 6 previously flaky tests in `userLayout.test.ts`
- The `storageAtoms.test.ts` regression tests verify all atoms are defined and return correct defaults from a fresh store, which would catch a circular dependency regression

## Related Issues

- Part of general codebase health

### Check List

### DIFF
--- a/packages/graph-explorer/src/core/StateProvider/userLayout.test.ts
+++ b/packages/graph-explorer/src/core/StateProvider/userLayout.test.ts
@@ -1,6 +1,4 @@
 // @vitest-environment happy-dom
-import type { ExtractAtomValue } from "jotai";
-
 import { act } from "react";
 
 import {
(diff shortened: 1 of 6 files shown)


packages/graph-explorer/src/core/StateProvider/userLayout.test.ts
```diff
@@ -134,3 +137,56 @@ describe("useSidebar", () => {
...
+
+describe("useTableViewSize", () => {
+  it("should default to DEFAULT_TABLE_VIEW_HEIGHT", () => {
+    const { result } = renderHookWithState(() => useTableViewSize());
+
+    expect(result.current[0]).toBe(300);
+  });
+
+  it("should return 100% when graph viewer is hidden", () => {
+    const { result } = renderHookWithJotai(
```
kmcginnes: Avoid using `renderHookWithJotai`. Use `renderHookWithState` instead.

arseny-kostenko (approved):

PR #1769, opened 2026-05-14
feat: support ?s ?p ?o shorthand in SPARQL query results

### LINKED ISSUE #1766: Support ?s ?p ?o shorthand in SPARQL query editor results
## Problem Statement

When users write SPARQL SELECT queries using the common shorthand `?s ?p ?o` in the Query Editor, the results cannot be added to the canvas as graph entities. Users must type the verbose `?subject ?predicate ?object` form to get renderable results. This is confusing because the shorthand is the most common convention in SPARQL tooling, and there is no indication to the user why one form works and the other does not.

## Solution

Recognize `?s ?p ?o` as equivalent to `?subject ?predicate ?object` when determining whether a SPARQL response contains quad-shaped data that can be rendered as graph entities on the canvas.

## User Stories

1. As a graph analyst, I want to write `SELECT ?s ?p ?o WHERE { ?s ?p ?o }` and have results rendered as vertices/edges on the canvas, so that I do not need to remember verbose variable names.
2. As a SPARQL user, I want the Query Editor to behave consistently regardless of whether I use short or long variable names, so that I am not confused by different rendering behavior.
3. As a user from Neptune Notebooks, I want the same `?s ?p ?o` convention to work in Graph Explorer, so that my existing queries transfer without modification.
4. As a new user, I want results from common SPARQL tutorials (which use `?s ?p ?o`) to render visually, so that I can explore my graph without learning Graph Explorer-specific conventions.

## Implementation Decisions

- **Schema-based approach**: Modify `sparqlQuadBindingSchema` in `types.ts` to accept both long-form (`subject`, `predicate`, `object`) and short-form (`s`, `p`, `o`) key names. Use Zod's union or preprocessing capabilities so the schema itself handles both variants.
- **Output normalization**: After successful schema validation of either form, normalize the parsed result to the canonical `{subject, predicate, object}` shape so downstream code (`mapQuadToEntities`, bnode filtering, etc.) remains unchanged.
- **Matching rules**: Only the exact set `{s, p, o}` (with optional `g` or `c` for graph/context) is accepted as the short form. Partial matches (e.g., `{s, predicate, object}`) must fail validation and fall through to the SELECT handler.
- **Strict validation preserved**: Both the long-form and short-form schemas remain strict — no extra keys allowed. This prevents SELECT queries with additional variables from being misidentified as quads.
- **No changes to CONSTRUCT/DESCRIBE**: These query forms already produce correctly-named variables from the SPARQL endpoint. The schema change only affects the validation gate in `rawquery.ts`.
- **Blank node filtering still applies**: After normalization, the same bnode filtering in `handleConstructQueryResults` applies to short-form results.

## Testing Decisions

- Test the schema directly: verify `sparqlQuadBindingSchema` accepts both `{subject, predicate, object}` and `{s, p, o}` bindings and produces the same normalized output shape.
- Test that partial matches (e.g., `{s, predicate, object}`) fail schema validation.
- Test that extra keys (e.g., `{s, p, o, extra}`) fail strict validation.
- Test end-to-end in `rawQuery.test.ts`: a SELECT query with `?s ?p ?o` variables and URI-typed subject/predicate should produce vertex/edge results (same as the long form).
- Prior art: existing tests in `rawQuery.test.ts` and `parseAndMapQuads.test.ts` that verify SELECT vs CONSTRUCT discrimination.

## Out of Scope

- Auto-completion or query suggestions in the SPARQL editor.
- Redesigning blank node support.
- Changes to CONSTRUCT or DESCRIBE query handling.
- Supporting arbitrary variable name mappings beyond the `s`/`p`/`o` convention.

## Further Notes

This was identified as a usability gap between Graph Explorer and Neptune Notebooks, where both `?s ?p ?o` and `?subject ?predicate ?object` are supported for visual rendering. The implementation is intentionally minimal — a schema union with output normalization — to avoid complicating the existing parsing pipeline.

vishwakt:
## Description

This PR supports the `?s ?p ?o` shorthand in the Query Editor so results render correctly as graph entities.

Changes include:
* Updates `sparqlQuadBindingSchema` to normalize the `{s, p, o}` short form into `{subject, predicate, object}`.
* Adds validation tests to verify shorthand parsing and strict rejection of partial matches.
* Updates query logic tests to ensure `SELECT ?s ?p ?o` executes and behaves as a `CONSTRUCT` query.

## Related Issues

Fixes #1766

### Check List

* [x] I confirm that my contribution is made under the terms of the Apache 2.0 license.
* [x] I have verified `pnpm checks` passes with no errors.
* [x] I have verified `pnpm test` passes with no failures.
* [x] I have covered new added functionality with unit tests if necessary.
* [x] I have updated documentation if necessary.

### DIFF
--- a/packages/graph-explorer/src/connector/sparql/types.ts
+++ b/packages/graph-explorer/src/connector/sparql/types.ts
@@ -206,7 +206,7 @@ export const sparqlAskResponseSchema = z.object({
   boolean: z.boolean(),
 });
 
-export const sparqlQuadBindingSchema = z
+const sparqlQuadBindingLongSchema = z
   .object({
     subject: sparqlResourceValueSchema,
     predicate: sparqlUriValueSchema,
@@ -214,4 +214,25 @@ export const sparqlQuadBindingSchema = z
     graph: sparqlValueSchema.optional(),
   })
   .strict();
+
+const sparqlQuadBindingShortSchema = z
+  .object({
+    s: sparqlResourceValueSchema,
+    p: sparqlUriValueSchema,
+    o: sparqlValueSchema,
+    g: sparqlValueSchema.optional(),
+    c: sparqlValueSchema.optional(),
+  })
+  .strict()
+  .transform(val => ({
+    subject: val.s,
+    predicate: val.p,
+    object: val.o,
+    graph: val.g ?? val.c,
+  }));
+
+export const sparqlQuadBindingSchema = z.union([
+  sparqlQuadBindingLongSchema,
+  sparqlQuadBindingShortSchema,
+]);
 export type SparqlQuadBinding = z.infer<typeof sparqlQuadBindingSchema>;
--- a/packages/graph-explorer/src/connector/sparql/parseAndMapQuads.test.ts
+++ b/packages/graph-explorer/src/connector/sparql/parseAndMapQuads.test.ts
@@ -159,14 +159,14 @@ describe("parseAndMapQuads", () => {
     it("should parse shorthand variables (?s ?p ?o) correctly", () => {
       const vertex = createTestableVertex().withRdfValues();
       const bindings = createQuadBindingsForEntities([vertex], []);
-      
+
       // Convert standard bindings to shorthand variables
       const shorthandBindings = bindings.map(b => ({
         s: b.subject,
         p: b.predicate,
         o: b.object,
       }));
-      
+
       const data = {
         head: { vars: ["s", "p", "o"] },
         results: { bindings: shorthandBindings },
--- a/packages/graph-explorer/src/connector/sparql/parseAndMapQuads.test.ts
+++ b/packages/graph-explorer/src/connector/sparql/parseAndMapQuads.test.ts
@@ -109,6 +109,24 @@ describe("parseAndMapQuads", () => {
       expect(() => parseAndMapQuads(invalidData)).toThrow();
       expect(logger.error).toHaveBeenCalled();
     });
+
+    it("should throw validation error for partial shorthand matches (e.g., s, predicate, object)", () => {
+      const invalidData = {
+        head: { vars: ["s", "predicate", "object"] },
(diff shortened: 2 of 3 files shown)


packages/graph-explorer/src/connector/sparql/types.ts
```diff
@@ -206,12 +206,33 @@ export const sparqlAskResponseSchema = z.object({
...
+    subject: val.s,
+    predicate: val.p,
+    object: val.o,
+    graph: val.g ?? val.c,
+  }));
+
+export const sparqlQuadBindingSchema = z.union([
+  sparqlQuadBindingLongSchema,
+  sparqlQuadBindingShortSchema,
+]);
```
kmcginnes, line 237: I like this approach. And good job on the tests. Let's also add some tests for the `?g` vs `?c` logic. It looks like you've chosen to take `?g` over `?c` if both exist. I think that is totally fine. Keep it. Just add the tests demonstrating it.

kmcginnes: Oh, I forgot the tests don't have access to those internals. In fact, I believe the `binding.graph` property is entirely ignored. So disregard my previous comment.

kmcginnes (commented):
This looks good to me. Just run `pnpm format` at the root to fix the formatting issues.

kmcginnes:
Tested this locally and it works as expected.

vishwakt:
@kmcginnes i did see that there was a CVE in one of the packages for which the build was failing on the previous run. The CVE is pre-existing on main and would need a separate dependency bump to resolve. Happy to file a follow-up issue and work on the fix as well if useful.

kmcginnes:
If you rebase on main you should get the fix. Unless there's been another one since.

vishwakt:
fair point. just merged main.

kmcginnes (approved):

PR #1771, opened 2026-05-18
feat: improve product skill score (41% → 90%)

yogesh-tessl:
Hey @kmcginnes 👋

ran your skills through `tessl skill review` at work and found some targeted improvements for the `product` skill. Here's the before/after:

| Skill | Before | After | Change |
|-------|--------|-------|--------|
| product | 41% | 90% | +49% |

I focused on the `product` skill since it had the most room to grow. It's the "front door" overview of Graph Explorer and was scoring lowest alongside `schema`.

<details>
<summary>Changes made to <code>product</code> skill</summary>

- Added explicit "Use when..." clause to the description so agents know when to select this skill
- Removed unknown `tools` frontmatter key that was triggering a validation warning
- Added a concrete quickstart section with copy-paste Docker commands (from your existing docs)
- Added a database/query language compatibility table with decision guidance (Gremlin vs openCypher vs SPARQL)
- Added cross-references to deeper docs (`docs/features/connections.md`, `docs/architecture.md`, etc.) for progressive disclosure
- Tightened the architecture section with monorepo structure, key libraries, and the proxy rationale
- Kept the diff small (64 lines) - all additive, no structural rewrites

</details>

---

also stress-tested your `git` skill against [a few real-world task evals](https://tessl.io/registry/skills/github/aws/graph-explorer/git/evals) and it held up really well on Vitest hook testing with `renderHookWithState` and backward compatibility checks for persisted data. Kudos for that.

quick honest disclosure. I work at https://github.com/tesslio where we build tooling around skills like these. Not a pitch, just saw room for improvement and wanted to contribute.

If you want to self-improve your skills, or define your own scenarios to pressure test, just ask your agent (Claude Code, Codex, etc.) to evaluate and optimize your skill with Tessl. Ping me [@yogesh-tessl](https://github.com/yogesh-tessl), if you hit any snags.

### DIFF
--- a/.kiro/skills/product/SKILL.md
+++ b/.kiro/skills/product/SKILL.md
@@ -1,36 +1,60 @@
 ---
 name: product
-description: Graph Explorer product overview including supported graph types, query languages, databases, key features, and high-level architecture.
-tools: ["fs_read", "grep", "glob", "web_search"]
+description: "Graph Explorer product context — supported databases, query languages, features, and architecture. Use when the user asks about Graph Explorer capabilities, deployment options, supported databases, query language choices, or how the system is structured."
 ---
 
 # Graph Explorer Product Overview
 
-Graph Explorer is a React-based web application that enables users to visualize and explore graph data without writing queries. It supports multiple graph database types and query languages.
+Graph Explorer is a React-based web application for visualizing and exploring graph data without writing queries. It supports multiple graph databases and query languages through a connector abstraction.
 
-## Core Purpose
+## Quickstart
 
-- Visualize property graphs (LPG) and RDF data
-- Interactive graph exploration without requiring query knowledge
-- Connect to various graph databases over HTTP
+Clone the repo and launch with the Air Routes sample dataset:
 
-## Supported Graph Types & Query Languages
+```bash
+cd graph-explorer/samples/air_routes
+docker compose up
+```
 
-- **Property Graphs**: Gremlin, openCypher
-- **RDF Graphs**: SPARQL 1.1 protocol
-- **Databases**: Amazon Neptune, Amazon Neptune Analytics, Apache TinkerPop Gremlin Server REST endpoints, JanusGraph
+Open [http://localhost:8080/explorer](http://localhost:8080/explorer). A default Gremlin connection is pre-configured.
+
+For production deployment, pull the official image from ECR Public:
+
+```bash
+docker pull public.ecr.aws/neptune/graph-explorer
+docker run -p 80:80 -p 443:443 --env HOST=localhost public.ecr.aws/neptune/graph-explorer
+```
+
+Then open `https://localhost/explorer`.
+
+## Supported Databases & Query Languages
+
+| Graph Model | Query Language | Databases |
+|---|---|---|
+| Property Graph (LPG) | Gremlin | Amazon Neptune, Apache TinkerPop Gremlin Server, JanusGraph |
+| Property Graph (LPG) | openCypher | Amazon Neptune, Amazon Neptune Analytics |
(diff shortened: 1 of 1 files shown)


.kiro/skills/product/SKILL.md
```diff
@@ -1,36 +1,60 @@
...
+description: "Graph Explorer product context — supported databases, query languages, features, and architecture. Use when the user asks about Graph Explorer capabilities, deployment options, supported databases, query language choices, or how the system is structured."
 ---
 
 # Graph Explorer Product Overview
 
-Graph Explorer is a React-based web application that enables users to visualize and explore graph data without writing queries. It supports multiple graph database types and query languages.
+Graph Explorer is a React-based web application for visualizing and exploring graph data without writing queries. It supports multiple graph databases and query languages through a connector abstraction.
 
-## Core Purpose
+## Quickstart
```
kmcginnes, line 10: The "Core Purpose" section describes the product's use case, which the agent can't derive from code alone. It helps the agent understand *what problem Graph Explorer solves*, not just what it is. I'd prefer to keep it since the vagueness is deliberate — it stays accurate as the app evolves.

.kiro/skills/product/SKILL.md
```diff
@@ -1,36 +1,60 @@
...
-- **RDF Graphs**: SPARQL 1.1 protocol
-- **Databases**: Amazon Neptune, Amazon Neptune Analytics, Apache TinkerPop Gremlin Server REST endpoints, JanusGraph
+Open [http://localhost:8080/explorer](http://localhost:8080/explorer). A default Gremlin connection is pre-configured.
+
+For production deployment, pull the official image from ECR Public:
+
+```bash
+docker pull public.ecr.aws/neptune/graph-explorer
+docker run -p 80:80 -p 443:443 --env HOST=localhost public.ecr.aws/neptune/graph-explorer
+```
```
kmcginnes, line 26: This "Quickstart" section duplicates the user guide and isn't something the agent would typically need to perform. If it ever does, it can discover the quick start instructions from the existing docs via progressive disclosure. I'd drop this section to keep the skill focused on product context rather than setup steps.

.kiro/skills/product/SKILL.md
```diff
@@ -1,36 +1,60 @@
...
+
+Then open `https://localhost/explorer`.
+
+## Supported Databases & Query Languages
+
+| Graph Model | Query Language | Databases |
+|---|---|---|
+| Property Graph (LPG) | Gremlin | Amazon Neptune, Apache TinkerPop Gremlin Server, JanusGraph |
+| Property Graph (LPG) | openCypher | Amazon Neptune, Amazon Neptune Analytics |
+| RDF | SPARQL 1.1 | Amazon Neptune |
```
kmcginnes, line 36: I like the intent here, but I worry this table is too prescriptive for an agent facing doc. These are providers we've tested, not an exhaustive set of what's supported. There are other Gremlin and SPARQL compatible databases that would work fine but aren't listed. Could we rename the column to "Example Databases" to frame these as known-good examples rather than a definitive list? Otherwise the agent might reject valid configurations just because they're not in the table.

kmcginnes (changes requested):
Thanks for the contribution @yogesh-tessl! I appreciate you taking the time to analyze the skill and propose improvements.

I like several of the changes here: the frontmatter cleanup, the tighter summary sentence, the feature list with doc cross-references, and the architecture rewrite. However, I'm requesting changes on the quick start section and the database table (see inline comments), and I'd like to keep the "Core Purpose" section.

Also worth noting: these skill files will likely be in flux as I adopt a more structured approach to AI steering, so don't be surprised if things shift around soon.

PR #1773, opened 2026-05-20
Unify Docker image by removing SageMaker variant

### LINKED ISSUE #1772: Unify Docker image: remove separate SageMaker build
## Problem Statement

Graph Explorer maintains two Docker image variants built from the same Dockerfile — a main image and a SageMaker image differing only in build-time environment variable defaults. This doubles CI build/scan time, creates a confusing tag namespace (`sagemaker-*` vs regular), and forces the SageMaker lifecycle script to track a separate tag lineage.

On the client side, users must manually configure a "Public or Proxy Endpoint" URL and enable a "Using Proxy-Server" checkbox, even though the proxy server and UI are always co-located on the same origin. This creates unnecessary configuration friction — especially in SageMaker where the proxy URL must be derived from notebook metadata.

## Solution

Ship a single Docker image that works in all deployment contexts. The frontend uses relative asset paths and relative API fetches, eliminating build-time knowledge of the deployment path prefix. The client always routes database queries through the same-origin proxy server, removing the need for users to provide a proxy URL. The `NEPTUNE_NOTEBOOK` environment variable is preserved as a runtime convenience preset that configures port, logging, and SSL defaults.

## User Stories

1. As a user deploying on SageMaker, I want to pull the same image tag as everyone else, so that I don't have to track a separate `sagemaker-*` tag lineage.
2. As a user deploying on SageMaker with an existing lifecycle script, I want my deployment to keep working after an image upgrade, so that I don't have to change anything.
3. As a user creating a Connection, I want to only provide my database endpoint and auth settings, so that I don't have to figure out what "Public or Proxy Endpoint" means or what URL to put there.
4. As a user deploying behind a custom reverse proxy, I want the app to work without build-time path configuration, so that I can mount it at any URL prefix.
5. As a user editing an existing Connection, I want my saved connections to keep working after upgrade, so that I don't have to reconfigure them.
6. As a user deploying with ECS Fargate, I want to simplify my task definition environment variables, so that I don't need to specify proxy endpoint and proxy toggle vars.
7. As a user deploying with Docker, I want to pass only the database URL and auth settings, so that the setup is minimal.
8. As a CI maintainer, I want to build one image instead of two, so that build time and vulnerability scan coverage are halved.
9. As a user running with `NEPTUNE_NOTEBOOK=true`, I want the container to automatically use port 9250 and cloudwatch logging, so that I don't need to pass those explicitly.
10. As a user passing explicit `PROXY_SERVER_HTTP_PORT`, I want my value to take precedence over the `NEPTUNE_NOTEBOOK` defaults, so that I have full control.
11. As a user with `NEPTUNE_NOTEBOOK=true`, I want SSL to be forced off regardless of other settings, so that the Jupyter proxy works correctly.
12. As a developer, I want the `proxyConnection` concept removed from the codebase, so that there's one path for all database queries.
13. As a user viewing the Connection detail, I want to see my database endpoint clearly, so that I know which database I'm connected to.
14. As a user hitting a critical error, I want the "reload" link to take me back to the app root, so that I get a clean slate.
15. As a user who previously had `proxyConnection: false`, I want my saved database URL to be preserved on upgrade, so that I don't lose my connection.
16. As a user who previously had `proxyConnection: true`, I want my saved graph database URL to be preserved on upgrade, so that I don't lose my connection.

### Relative asset paths
- Vite `base` changes from `env.GRAPH_EXP_ENV_ROOT_FOLDER` to `"./"`.
- `index.html` gets a `<base href="./">` tag to anchor `document.baseURI` regardless of client-side routing. This future-proofs for a potential move from HashRouter to proper URL routes.
- `RELOAD_URL` simplifies to `"."`.

### Relative API fetches
- All client API calls (sparql, gremlin, openCypher, defaultConnection, logger, summary, etc.) use `new URL("../endpoint", document.baseURI)` instead of constructing absolute URLs from `connection.url`.
- The static files are served at `/explorer` and API routes at `/`. The `../` from the static directory resolves to the API root. This relationship is fixed in one place on the server.
- The two-path fallback in `defaultConnection.ts` (trying both root and `/proxy/9250/`) is removed in favor of a single relative fetch.

### Connection model simplification
- `proxyConnection` and `url` are removed from the `ConnectionConfig` type.
- `graphDbUrl` becomes the canonical required field for the database endpoint.
- `fetchDatabaseRequest` always sends the `graph-db-connection-url` header — no conditional based on `proxyConnection`.
- A read-time normalization handles legacy stored data: if `proxyConnection` was true, use `graphDbUrl`; if false, use `url`.

### Connection form simplification
- The "Public or Proxy Endpoint" text field is removed.
- The "Using Proxy-Server" checkbox is removed.
- "Graph Connection URL" is always visible (not gated behind the proxy checkbox).
- The IAM auth toggle is always visible (not gated behind the proxy checkbox).

### Dockerfile
- `ARG NEPTUNE_NOTEBOOK` and all conditional ENV logic are removed.
- `GRAPH_EXP_ENV_ROOT_FOLDER` is removed.
- `PROXY_SERVER_HTTP_PORT` and `LOG_STYLE` are NOT set as ENV vars — the app's built-in defaults (80, "default") apply when absent.
- `EXPOSE 80`, `EXPOSE 443`, `EXPOSE 9250` remain.

### Runtime NEPTUNE_NOTEBOOK handling (process-environment.sh)
- When `NEPTUNE_NOTEBOOK=true`:
  - Writes `PROXY_SERVER_HTTP_PORT=9250` to `.env` (respects explicit env var override via dotenv's no-override behavior).
  - Writes `LOG_STYLE=cloudwatch` to `.env` (same override behavior).
  - Forces `PROXY_SERVER_HTTPS_CONNECTION=false` and `GRAPH_EXP_HTTPS_CONNECTION=false` (not overridable).
- Stops writing `NEPTUNE_NOTEBOOK` itself to `.env` — nothing downstream reads it.
- `PUBLIC_OR_PROXY_ENDPOINT` and `USING_PROXY_SERVER` handling is removed from the script. These env vars are silently ignored if still passed.

### CI: dual-tag publishing
- CI builds one image.
- Publishes under both regular tags and `sagemaker-*` tags during a transition period.
- The `--build-arg NEPTUNE_NOTEBOOK=true` step is removed.

### Lifecycle script update
- Pulls the regular tag (not `sagemaker-*`).
- Removes `PUBLIC_OR_PROXY_ENDPOINT` and `USING_PROXY_SERVER` from `docker run`.

### Documentation updates
- `docs/references/configuration.md`
- `docs/guides/deploy-to-ecs-fargate.md`
- `docs/guides/deploy-to-sagemaker.md`
- `docs/guides/deploy-with-docker.md`
- `docs/features/connections.md`
- `docs/architecture.md`

## Testing Decisions

A good test for this work exercises observable external behavior through the public interfaces — not internal wiring. Tests should verify what a user or calling code observes, not how the internals are structured.

### Modules to test

1. **Connection migration (read-time normalization)** — Given legacy stored connection data with various combinations of `proxyConnection`, `url`, and `graphDbUrl`, verify that the normalized output always produces the correct `graphDbUrl`. Prior art: `defaultConnection.test.ts`, `configuration.ts` normalization tests.

2. **Relative URL resolution** — Given various `document.baseURI` values (representing `/explorer/`, `/proxy/9250/explorer/`, custom prefix), verify that the API URL construction resolves to the correct absolute path. This can be a pure function test with injected base URIs. Prior art: unit tests in `connector/` and `core/`.

3. **process-environment runtime defaults** — Given various combinations of environment variables and `NEPTUNE_NOTEBOOK=true/false`, verify the correct values are written to `.env` and `defaultConnection.json`. Verify explicit env vars take precedence. Verify SSL is forced off when `NEPTUNE_NOTEBOOK=true`. Prior art: `process-environment.test.ts` (extensive existing test suite).

4. **Connection form UI** — Verify the form renders the correct fields (no proxy endpoint, no proxy toggle), validates that `graphDbUrl` is required, and submits the correct shape. Prior art: existing component tests using React Testing Library patterns in the codebase.

5. **fetchDatabaseRequest** — Verify that `graph-db-connection-url` header is always sent. Verify AWS headers are sent when IAM is enabled. Verify the function works without `proxyConnection` in the connection object. Prior art: `fetchDatabaseRequest.test.ts` (comprehensive existing suite).

## Out of Scope

- Removing `NEPTUNE_NOTEBOOK` entirely (kept for backward compat with deployed lifecycle scripts).
- Dropping the `sagemaker-*` tag publishing (deferred to a future release after transition period).
- Moving from HashRouter to proper URL routes (future work, but `<base href>` prepares for it).
- Changing the `graphDbUrl` field name (kept for backward compat with stored data).
- Renaming the "Graph Connection URL" UI label.
- Server-side route restructuring (API routes stay at root, static files stay at `/explorer`).

## Further Notes

- The ADR for this decision is committed at `docs/adr/0001-unify-docker-image-remove-sagemaker-variant.md` on branch `remove-sagemaker-docker-image`.
- Users passing `PUBLIC_OR_PROXY_ENDPOINT` or `USING_PROXY_SERVER` as env vars to Docker will see no errors — the vars are simply ignored. No deprecation warning is emitted.
- The `defaultConnection.json` file still works as a pre-configuration mechanism for operators. It just no longer includes proxy endpoint or proxy toggle fields.

kmcginnes:
## Description

All client requests now always route through the same-origin proxy server using relative URLs. Users no longer need to configure a "Public or Proxy Endpoint" — the connection form simplifies to just the database URL and optional IAM settings.

This also eliminates the separate SageMaker Docker build. One image serves all deployment modes (standalone, SageMaker, arbitrary reverse proxies) without build-time path configuration.

Key changes:
- Client resolves API endpoints via `new URL("../endpoint", document.baseURI)` — works behind any reverse proxy without configuration
- Removes `url` (proxy endpoint) and `proxyConnection` toggle from the connection model; `graphDbUrl` is now the single required field
- Proxy headers (`graph-db-connection-url`) are always sent — no toggle needed
- `NEPTUNE_NOTEBOOK=true` becomes a runtime-only preset (port 9250, cloudwatch logs, SSL off)
- CI builds one image, publishes under both regular and `sagemaker-*` tags during transition
- Legacy stored connections auto-migrate on read via `migrateLegacyConnection()`

### Breaking change

All database requests now flow through the proxy server. Deployments where the server is in a restricted network (e.g., a private subnet without a NAT gateway) will not be able to reach databases outside that network, even if the user's browser previously could via direct connections. Users in this scenario need to add appropriate network routing (VPC peering, NAT gateway, etc.).

## Validation

- All 1709 existing tests pass
- Zero TypeScript errors across all packages
- Zero lint errors
- Tested on SageMaker Neptune Notebook (`aws-neptune-gx-air-routes`) — connections work with no lifecycle script changes
- Manual testing:
  - [x] Deploy behind SageMaker reverse proxy at `/proxy/9250/explorer/`, verify relative URLs resolve correctly
  - [x] Existing lifecycle script passes legacy env vars (`PUBLIC_OR_PROXY_ENDPOINT`, `USING_PROXY_SERVER`) — silently ignored
  - [ ] Create/edit connections in UI, verify form shows only Graph Connection URL + IAM fields
  - [ ] Import legacy connection file (with `url`/`proxyConnection`), verify migration works

## Related Issues

- Closes #1772
- ADR: `docs/adr/0001-unify-docker-image-remove-sagemaker-variant.md`

### Check List

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -13,8 +13,8 @@ A directed relationship between two vertices (source → target), with a type an
 _Avoid_: Relationship, link
 
 **Connection**:
-A saved database profile — the URL, query engine, authentication settings, and proxy routing needed to reach a graph database. Users create and manage these in the UI.
-_Avoid_: Configuration (legacy term being phased out — previously bundled connection + schema + Styles into one object)
+A saved database profile — the graph database endpoint URL (`graphDbUrl`), query language, and optional IAM authentication settings. The client always reaches the database through the same-origin **Proxy Server**, so no proxy endpoint is configured. Users create and manage these in the UI.
+_Avoid_: Configuration (legacy term being phased out — previously bundled connection + schema + Styles into one object); proxy endpoint / `proxyConnection` (removed — see ADR `unify-docker-image-remove-sagemaker-variant`)
 
 **Active Connection**:
 The one Connection a single browser tab is currently exploring — what the entire app reads to decide which Schema, Session, and queries are in play. Held per-tab (sessionStorage): it survives that tab's reload but dies with the tab, and one tab activating a Connection never changes another tab's Active Connection.
@@ -86,6 +86,10 @@ _Avoid_: Save state (ambiguous with Session)
 The UI element in the nav bar (after the page title) that renders Persistence Status. It surfaces only on `failed` — a standing danger "Changes not saved" button — and stays absent at `idle` and `saving`. Clicking it opens a dialog showing the raw failure records (key, reason, attempt count, last attempt, and the underlying error's name/message/cause) in a read-only JSON editor. The dialog offers to save the configuration to a file via `saveLocalForageToFile` (`core/StateProvider/localDb.ts`) when storage is full (quota) — IndexedDB is still readable then — but not when storage is inaccessible (private mode, blocked), since the database never opened and there is nothing to read. Recovery scope is retry (transient failures) plus that backup (terminal-quota failures) — it does not guarantee the write eventually lands.
 _Avoid_: Save-status indicator
 
+**Proxy Server**:
+The Node.js server that serves the frontend (mounted at `/explorer`) and proxies all database requests (mounted at `/`). The client resolves API endpoints relative to its own origin via `apiUrl()` — `../endpoint` from the static mount — so the frontend and proxy are always same-origin. This means every database request routes through the Proxy Server, which has network access to the database and handles SigV4 signing. See ADR `unify-docker-image-remove-sagemaker-variant`.
+_Avoid_: proxy endpoint URL (no longer user-configured)
+
 ## Relationships
 
 - Each browser tab has at most one **Active Connection**; different tabs may have different ones
@@ -120,3 +124,4 @@ _Avoid_: Save-status indicator
 - "Node" means **Vertex** in code but is the preferred UI term for property graphs — resolved: use **Vertex** in code, "node" in UI copy.
 - "Attribute" vs "Property" — resolved: **Property** is canonical, "attribute" is legacy code term being phased out.
 - "Active connection" meant a single shared per-origin value, but the app consumed it as if it were per-tab — resolved: **Active Connection** is per-tab (sessionStorage), **Last Active Connection** is the shared persisted breadcrumb. The legacy `activeConfigurationAtom` / `active-configuration` key is reused as the breadcrumb.
+- Connection URL field — `url` and `proxyConnection` are legacy fields that survive only in `LegacyConnectionConfig` and are migrated at read time to the canonical `graphDbUrl` (see `migrateLegacyConnection`, ADR `unify-docker-image-remove-sagemaker-variant`). Use `graphDbUrl` in code.
--- a/docs/guides/deploy-to-sagemaker/install-graph-explorer-lc.sh
+++ b/docs/guides/deploy-to-sagemaker/install-graph-explorer-lc.sh
@@ -130,17 +130,14 @@ LATEST_ECR_RELEASE=$(curl -k -H "Authorization: Bearer $ECR_TOKEN" https://publi
 
 echo "Pulling and starting graph-explorer..."
 if [[ ${EXPLORER_VERSION} == "" ]]; then
-  EXPLORER_ECR_TAG=sagemaker-${LATEST_ECR_RELEASE}
+  EXPLORER_ECR_TAG=${LATEST_ECR_RELEASE}
 else
-  if [[ ${EXPLORER_VERSION//./} -ge 140 ]]; then
-    EXPLORER_ECR_TAG=sagemaker-${EXPLORER_VERSION}
-  elif [[ ${EXPLORER_VERSION} == *latest* ]]; then
-    EXPLORER_ECR_TAG=sagemaker-latest-SNAPSHOT
+  if [[ ${EXPLORER_VERSION} == *latest* ]]; then
+    EXPLORER_ECR_TAG=latest-SNAPSHOT
   elif [[ ${EXPLORER_VERSION} == *dev* ]]; then
-    EXPLORER_ECR_TAG=sagemaker-dev
+    EXPLORER_ECR_TAG=dev
   else
-    echo "Specified Graph Explorer version does not support use on SageMaker. Defaulting to latest release."
-    EXPLORER_ECR_TAG=sagemaker-${LATEST_ECR_RELEASE}
+    EXPLORER_ECR_TAG=${EXPLORER_VERSION}
   fi
 fi
 echo "Using explorer image tag: ${EXPLORER_ECR_TAG}"
@@ -155,13 +152,10 @@ start_graph_explorer_with_cw_logs() {
       --log-opt awslogs-multiline-pattern='^(INFO|DEBUG|ERROR|WARN|TRACE|FATAL)' \
       --env LOG_LEVEL=debug \
       --env HOST=127.0.0.1 \
-      --env PUBLIC_OR_PROXY_ENDPOINT=${EXPLORER_URI} \
       --env GRAPH_CONNECTION_URL=${NEPTUNE_URI} \
-      --env USING_PROXY_SERVER=true \
       --env IAM=${IAM} \
       --env AWS_REGION=${AWS_REGION} \
       --env SERVICE_TYPE=${SERVICE} \
-      --env PROXY_SERVER_HTTPS_CONNECTION=false \
       --env NEPTUNE_NOTEBOOK=true public.ecr.aws/neptune/graph-explorer:${EXPLORER_ECR_TAG}
 }
 
@@ -170,13 +164,10 @@ start_graph_explorer_with_default_logs() {
       --restart always \
       --env LOG_LEVEL=debug \
       --env HOST=127.0.0.1 \
-      --env PUBLIC_OR_PROXY_ENDPOINT=${EXPLORER_URI} \
       --env GRAPH_CONNECTION_URL=${NEPTUNE_URI} \
-      --env USING_PROXY_SERVER=true \
       --env IAM=${IAM} \
       --env AWS_REGION=${AWS_REGION} \
       --env SERVICE_TYPE=${SERVICE} \
-      --env PROXY_SERVER_HTTPS_CONNECTION=false \
       --env NEPTUNE_NOTEBOOK=true public.ecr.aws/neptune/graph-explorer:${EXPLORER_ECR_TAG}
 }
 
--- a/docs/references/configuration.md
+++ b/docs/references/configuration.md
@@ -8,16 +8,6 @@ All environment variables for configuring Graph Explorer, organized by concern.
 
 These variables control server behavior, networking, and security.
 
-### `GRAPH_EXP_ENV_ROOT_FOLDER`
-
-Base path used to serve the `graph-explorer` front end application.
-
-Example: `/explorer`
-
-- Optional
-- Default: `/`
-- Type: `string`
-
 ### `HOST`
 
 The public hostname of the server. This is used to generate the self-signed SSL certificate at container startup.
@@ -115,6 +105,14 @@ Override path for the folder containing `.env` and `defaultConnection.json`. Whe
 - Default: `<client root>` (`packages/graph-explorer`)
 - Type: `string`
 
+### `NEPTUNE_NOTEBOOK`
+
+Runtime convenience preset for SageMaker/Jupyter deployments. When set to `true`, configures port 9250, cloudwatch logging, and disables SSL automatically.
+
+- Optional
+- Default: not set
+- Type: `boolean`
+
 ## Default Connection
 
 To provide a default connection such that initial loads of Graph Explorer always result with the same starting connection, modify the `docker run ...` command to either take in a JSON configuration or runtime environment variables. If you provide both a JSON configuration and environmental variables, the JSON will be prioritized.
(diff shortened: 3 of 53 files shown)


docs/guides/deploy-to-sagemaker/install-graph-explorer-lc.sh
```diff
@@ -155,13 +152,10 @@ start_graph_explorer_with_cw_logs() {
       --log-opt awslogs-multiline-pattern='^(INFO|DEBUG|ERROR|WARN|TRACE|FATAL)' \
       --env LOG_LEVEL=debug \
       --env HOST=127.0.0.1 \
-      --env PUBLIC_OR_PROXY_ENDPOINT=${EXPLORER_URI} \
       --env GRAPH_CONNECTION_URL=${NEPTUNE_URI} \
-      --env USING_PROXY_SERVER=true \
       --env IAM=${IAM} \
       --env AWS_REGION=${AWS_REGION} \
       --env SERVICE_TYPE=${SERVICE} \
-      --env PROXY_SERVER_HTTPS_CONNECTION=false \
```
kmcginnes, line 164: This should be kept

docs/guides/deploy-to-sagemaker/install-graph-explorer-lc.sh
```diff
@@ -170,13 +164,10 @@ start_graph_explorer_with_default_logs() {
       --restart always \
       --env LOG_LEVEL=debug \
       --env HOST=127.0.0.1 \
-      --env PUBLIC_OR_PROXY_ENDPOINT=${EXPLORER_URI} \
       --env GRAPH_CONNECTION_URL=${NEPTUNE_URI} \
-      --env USING_PROXY_SERVER=true \
       --env IAM=${IAM} \
       --env AWS_REGION=${AWS_REGION} \
       --env SERVICE_TYPE=${SERVICE} \
-      --env PROXY_SERVER_HTTPS_CONNECTION=false \
```
kmcginnes, line 179: This should be kept

docs/references/configuration.md
```diff
@@ -115,6 +105,14 @@ Override path for the folder containing `.env` and `defaultConnection.json`. Whe
...
 - Type: `string`
 
+### `NEPTUNE_NOTEBOOK`
+
+Runtime convenience preset for SageMaker/Jupyter deployments. When set to `true`, configures port 9250, cloudwatch logging, and disables SSL automatically.
+
+- Optional
+- Default: not set
+- Type: `boolean`
+
```
kmcginnes, line 115: We don't need to document this

packages/graph-explorer/src/modules/CreateConnection/CreateConnection.test.tsx
```diff
@@ -61,7 +57,11 @@ describe("CreateConnection", () => {
     const store = renderCreateConnection();
 
     await user.type(
-      screen.getByRole("textbox", { name: "Public or Proxy Endpoint" }),
+      screen.getByRole("textbox", { name: "Name" }),
+      "My Connection",
+    );
+    await user.type(
+      screen.getByRole("textbox", { name: "Graph Connection URL" }),
```
mjuarros, line 64: Could we expand this suite to cover the Connection form behavior required by #1772? In addition to the URL validation already covered, please verify that the Public or Proxy Endpoint field and Using Proxy-Server checkbox are absent, the Graph Connection URL and IAM controls are available without another toggle, and create/edit submissions contain the canonical `graphDbUrl` shape without legacy `url` or `proxyConnection` fields. These assertions protect the main user-facing simplification and backward-compatible edit path from regression.

process-environment.sh
```diff
@@ -18,15 +16,17 @@ if [ -f "./config.json" ]; then
...
 
-if [ -n "$NEPTUNE_NOTEBOOK" ]; then
-    printf '\nNEPTUNE_NOTEBOOK=%s\n' "$NEPTUNE_NOTEBOOK" >> $CONFIGURATION_FOLDER_PATH/.env
-    if [ "$NEPTUNE_NOTEBOOK" = "true" ]; then
-      # Override Proxy SSL setting if Neptune notebook
-      PROXY_SERVER_HTTPS_CONNECTION="false"
-      GRAPH_EXP_HTTPS_CONNECTION="false"
+if [ "$NEPTUNE_NOTEBOOK" = "true" ]; then
+    # Force SSL off for Neptune Notebook environments
+    PROXY_SERVER_HTTPS_CONNECTION="false"
```
mjuarros, line 21: This does not reliably force proxy HTTPS off when the container already has `PROXY_SERVER_HTTPS_CONNECTION=true`. The script writes `false` to `.env`, but `node-server.ts` loads it with the default `dotenv.config()` behavior, which does not override the inherited process environment. The entrypoint can therefore skip certificate generation after reading `false` from `.env`, while Node still reads the inherited `true` and attempts to start HTTPS without certificates. Please make the forced notebook value effective in the Node process and add a test with conflicting HTTPS environment variables that asserts the parsed runtime configuration has SSL disabled.

docs/references/configuration.md
```diff
@@ -124,19 +122,16 @@ To provide a default connection such that initial loads of Graph Explorer always
 These are the valid environment variables used for the default connection, their defaults, and their descriptions.
 
 - Required:
```
mjuarros, line 124: Could we also remove the note above that says direct database connections can bypass `PROXY_SERVER_ALLOWED_DB_ORIGINS`? This PR removes direct mode, so the note now describes an impossible configuration. The documentation should state that the allowlist applies to every database request because all requests route through the proxy.

CONTEXT.md
```diff
@@ -135,6 +135,10 @@ _Avoid_: Save state (ambiguous with Session)
...
 _Avoid_: Save-status indicator
 
+**Proxy Server**:
+The Node.js server that serves the frontend (mounted at `/explorer`) and proxies all database requests (mounted at `/`). The client resolves API endpoints relative to its own origin via `apiUrl()` — `../endpoint` from the static mount — so the frontend and proxy are always same-origin. This means every database request routes through the Proxy Server, which has network access to the database and handles SigV4 signing. See ADR `unify-docker-image-remove-sagemaker-variant`.
```
mjuarros, line 139: Please carry this proxy-only invariant into `docs/agents/product.md` as well. Its persistence section still says graph data is queried directly from connected databases, which is no longer accurate. It should clarify that data is queried live through the stateless Graph Explorer proxy and is not persisted by Graph Explorer.

packages/graph-explorer/src/core/StateProvider/configuration.ts
```diff
@@ -115,15 +118,30 @@ export function normalizeUrl(url: string | undefined): string {
...
+ * format where only `graphDbUrl` exists. */
+export function migrateLegacyConnection(
+  connection: LegacyConnectionConfig,
+): ConnectionConfig {
+  const { url, proxyConnection, ...rest } = connection;
+  // Proxy connections stored the database endpoint in `graphDbUrl`; direct
+  // connections stored it in `url`. The final `connection.graphDbUrl` fallback
+  // covers already-migrated data where `url` is absent, and the empty-string
+  // fallback keeps the result valid when no URL is present at all.
+  const graphDbUrl = proxyConnection ? connection.graphDbUrl : url;
```
mjuarros, line 131: This changes the previous fallback semantics when `proxyConnection` is absent. Before this PR, `normalizeConnection` inferred proxy mode whenever `graphDbUrl` existed. A legacy connection containing both `url` (the proxy endpoint) and `graphDbUrl` (the database endpoint), but no explicit `proxyConnection`, now selects `url` and stores the proxy endpoint as the database URL. Please preserve the old inference with `proxyConnection ?? connection.graphDbUrl != null` and add a backward-compatibility test for both URLs with an absent flag.

packages/graph-explorer/src/utils/parseConnectionFile.ts
```diff
@@ -34,13 +34,28 @@ const exportedConnectionFileSchema = z.looseObject({
...
+      // Direct connections stored the endpoint in `url`; `migrateLegacyConnection`
+      // folds these into `graphDbUrl` on import. Validated to the same scheme so
+      // a legacy file cannot smuggle in a non-http(s) target either.
+      url: z.url({ protocol: /^https?$/ }).optional(),
+      proxyConnection: z.boolean().optional(),
+    })
+    // The file must carry a usable endpoint in either the canonical or the
+    // legacy field, otherwise migration would yield an empty `graphDbUrl`.
+    .refine(
+      connection => connection.graphDbUrl != null || connection.url != null,
```
mjuarros, line 54: This accepts `proxyConnection: true` with only the legacy `url`, but migration intentionally does not use that proxy URL as the database endpoint and produces an empty `graphDbUrl`. The import therefore succeeds while creating an unusable connection. Could the refinement require `graphDbUrl` when `proxyConnection` is true, and add a negative import test for that inconsistent legacy shape?

mjuarros:
### Pre-existing follow-up findings

These are not introduced by this PR and should not block it, but they surfaced while tracing the updated proxy path and are worth validating separately:

- When `PROXY_SERVER_ALLOWED_DB_ORIGINS` is unset, the proxy accepts any HTTP(S) destination supplied through `graph-db-connection-url`. An exposed deployment can therefore act as a server-side request proxy to internal services. Consider a follow-up issue to evaluate a deny-by-default or administrator-managed allowlist.
- Error handling logs `graph-db-connection-url` verbatim. URLs containing user information could expose credentials in logs. Consider redacting the value to scheme, hostname, and port.
- `process-environment.sh` interpolates configuration values directly into `.env` and JSON. Quotes, backslashes, commas, or newlines can corrupt the generated files. Consider replacing the shell serialization with the existing Node runtime or another real JSON serializer.
- A current `pnpm audit --audit-level high` reports dependency updates needed for `browserslist` and the pinned pnpm version. This is independent of the PR implementation and can be tracked as dependency maintenance.

If these behaviors are still considered valid after maintainer review, I recommend tracking each concern in a separate issue so this PR can stay focused.

PR #1589, opened 2026-03-12
Add defaultStyling.json support for persistent per-type vertex and edge styling

jkemmererupgrade:
## Description

 Add support for a defaultStyling.json file that provides persistent per-type vertex and edge styling defaults. When mounted into the Docker container, the file's values are merged into the user styling
  store on startup, giving teams a consistent visual baseline across non-persistent browser sessions (e.g., AWS WorkSpaces Web).

  Key capabilities:
  - Docker mount: Place a defaultStyling.json at the configuration folder path to auto-apply styling on load
  - Lucide icon support: Reference icons by name (e.g., "icon": "user") — resolved to SVG data URIs at runtime
  - Import / Export / Reset UI: New controls in Settings to import a styling file, export current styling, and reset all styling to defaults
  - Per-type Reset to Default: Each node/edge type's style dialog has a "Reset to Default" button that restores the mounted file's values
  - Zod validation: Strict schema validation with clear error logging for malformed files

## Validation

  - All 1365 tests pass (pnpm test)
  - pnpm checks passes (lint, format, types)
  - Docker tested with mounted defaultStyling.json — styling applied on startup, import/export/reset all functional
  - Incognito window testing confirmed clean-state behavior
  - New unit tests cover: Zod parsing, icon resolution, import/export hooks, merge logic, configuration integration

Screenshot 2026-03-10 at 2 09 40 PM
Screenshot 2026-03-10 at 2 04 31 PM

## Related Issues

  Addresses #1265, #112, #173, #573, #689

### Check List

- [x] I confirm that my contribution is made under the terms of the Apache 2.0 license.
- [x] I have run `pnpm checks` to ensure code compiles and meets standards.
- [x] I have run `pnpm test` to check if all tests are passing.
- [x] I have covered new added functionality with unit tests if necessary.
- [x] I have added an entry in the `Changelog.md`.

PR #1777, opened 2026-05-22
Add built-in Lucide icon library for node styling

### LINKED ISSUE #1774: Add built-in Lucide icon library for node styling
## Problem Statement

Users who want to assign meaningful icons to node types must currently upload SVG or raster files manually through a file picker. This is friction-heavy: users need to find, download, and prepare icon files externally before they can style their graph. There is no browsable icon library inside the application, despite the app already shipping Lucide icons for its own UI.

## Solution

Add a searchable, browsable icon picker to the node styling dialog that exposes the full Lucide icon set. Users can search by name or keyword, click to select, and the icon is applied immediately. Custom file upload remains available for domain-specific icons not covered by Lucide.

Icons selected from the library are stored as `lucide:<name>` references and resolved at render time to SVG data URIs. This enables the icon picker to highlight the currently selected icon and allows config files to reference icons symbolically. The Lucide icon catalog is resolved using `lucide-react/dynamicIconImports` which provides per-icon lazy loading with Vite-compatible static analysis.

## User Stories

1. As a graph explorer user, I want to pick an icon from a built-in library when styling a node type, so that I don't need to find and upload SVG files externally.
2. As a graph explorer user, I want to search the icon library by keyword (e.g., "plane", "building", "person"), so that I can quickly find a relevant icon.
3. As a graph explorer user, I want to see a visual preview of each icon in the picker before selecting it, so that I can compare options.
4. As a graph explorer user, I want the icon picker to load quickly without slowing down initial app startup, so that the app remains responsive.
5. As a graph explorer user, I want to still upload a custom SVG or raster image if the library doesn't have what I need, so that I'm not limited to the built-in set.
6. As a graph explorer user, I want the selected Lucide icon to render correctly both in the graph canvas (Cytoscape) and in sidebar/search result previews, so that styling is consistent everywhere.
7. As a graph explorer user, I want icons to respect the node color I've configured, so that SVG icons use currentColor and match my color scheme.
8. As a graph explorer user, I want the icon I selected to persist across sessions, so that my styling choices are preserved.
9. As a graph explorer user, I want to change my icon selection at any time by reopening the styling dialog, so that I can iterate on my graph's appearance.
10. As a graph explorer user, I want the icon picker to indicate which icon is currently selected, so that I know what's already applied.
11. As a graph explorer user, I want to be able to reset the icon back to the default after selecting a Lucide icon or uploading a custom one.
12. As a graph explorer user, I want nodes with Lucide icons to render their icons on page load without needing to open the styling dialog first.

## Implementation Decisions

- **Storage format**: The `iconUrl` field in `VertexPreferencesStorageModel` stores `"lucide:<name>"` for library icons. This preserves the symbolic reference, enabling the icon picker to highlight the current selection and enabling config files to reference icons by name. No schema change needed — the field is already `string | undefined`.
- **Icon image type for Lucide**: When a Lucide icon is selected, `iconImageType` is set to `"image/svg+xml"` since all Lucide icons are SVGs.
- **Icon resolution at render time**: A resolver function handles all `iconUrl` formats:
  - `lucide:<name>` — extracts the icon name, calls the per-icon lazy loader from `lucide-react/dynamicIconImports`, builds an SVG string, and returns a data URI. Results are cached so each icon is resolved at most once per session.
  - `data:...` — passthrough (existing uploads)
  - Any other string — passthrough (treated as a URL)
- **Why `lucide-react/dynamicIconImports`**: This object maps every icon name to an explicit `() => import(...)` function. Because each import path is a static string literal inside the lucide-react library, Vite can statically analyze and produce a chunk per icon at build time. This avoids the need for a single large catalog chunk while still supporting fully dynamic icon name resolution at runtime.
- **Render-time integration (graph canvas)**: `renderNode.tsx` already fetches and processes SVG icons via React Query (`iconQueryOptions`). The resolver plugs into this existing pipeline — for `lucide:` references, the query function calls the resolver instead of fetching a URL. Caching (`staleTime: Infinity`) ensures each icon is resolved once.
- **Render-time integration (UI components)**: `VertexIcon.tsx` uses `react-inlinesvg` which accepts a URL or data URI as `src`. For `lucide:` references, the component resolves to a data URI before passing to the SVG renderer. This can use the same React Query cache or a local state pattern with the resolver.
- **Icon picker UI**: A new `IconPicker` component rendered inside the existing `NodeStyleDialog`. It consists of a search input and a scrollable grid of icon previews. Search filters against Lucide icon names (and optionally tags/categories from Lucide metadata). Selecting an icon calls `setVertexStyle({ iconUrl: "lucide:<name>", iconImageType: "image/svg+xml" })`.
- **Current selection highlighting**: The picker reads `vertexStyle.iconUrl`, checks if it starts with `"lucide:"`, extracts the name, and highlights the matching grid item. This is only possible because the symbolic reference is preserved in storage.
- **Picker icon rendering**: Each grid cell resolves its icon using the same `dynamicIconImports` loader. Icons load asynchronously with a placeholder shimmer while loading. HTTP/2 multiplexing handles the parallel micro-fetches adequately for the visible subset.
- **Picker performance**: Show a limited set initially (e.g., 50 icons), expand via search. Consider virtualization if scroll performance is an issue with the full set.
- **Custom upload preserved**: The existing file upload button remains in the dialog alongside the new picker, giving users an escape hatch for icons not in the Lucide set.
- **Backward compatibility**: Existing stored `iconUrl` values (base64 data URIs from prior uploads) continue to work unchanged — the resolver treats any `data:` string as a passthrough.
- **Bundle impact on main chunk**: Only the `dynamicIconImports` key map (~30KB) is imported statically. Actual icon SVG data is loaded on demand.

## Testing Decisions

- Tests should verify external behavior (what icons resolve to, what the user sees), not implementation details.
- **Icon resolver module**: Unit tests covering all input formats (`lucide:*`, data URI, plain URL passthrough), error handling for invalid/unknown Lucide icon names, correct SVG construction and data URI encoding, and caching behavior (second call for same name returns cached result). This is pure logic with a simple interface — highly testable in isolation.
- **Icon picker component**: Component tests verifying search/filter behavior, selection callback emits correct `lucide:<name>` format, current selection is highlighted when `vertexStyle.iconUrl` matches a Lucide icon, and graceful handling of the loading state for icon previews.
- **Cytoscape rendering**: Not explicitly tested as part of this work — the existing `renderNode` tests and visual verification cover downstream rendering. The resolver is the testable boundary.
- Prior art: follow patterns in existing test files under `src/core/StateProvider/` and `src/modules/` that use Vitest and the project's component rendering utilities.

## Out of Scope

- Server-hosted icon files or server config references to icons (separate work)
- Edge icons (edges use line styles, not icons)
- Icon animation or dynamic icon states
- User-uploadable icon packs or icon set management
- Lucide icon version pinning strategy or automatic update mechanism
- Icon color customization independent of node color (icons always inherit node color via currentColor)
- Default styling file system (defaultStyling.json mount, fetch, merge — separate issue)
- Styling import/export UI (separate issue)

## Further Notes

- Lucide publishes ~1500 icons, all MIT licensed.
- This feature creates the `lucide:<name>` storage convention that future work (unified server config, operator-authored default styling) can reference without additional client changes.
- The `dynamicIconImports` approach from `lucide-react` is preferred over `lucide-static` because it leverages Vite's static analysis without requiring a single large catalog bundle. Each icon loads individually on demand (~1KB per icon chunk).

jkemmererupgrade:
## Description

Adds the [Lucide](https://lucide.dev) icon library as a built-in source for node icons, introducing a searchable picker in the Node Styling panel alongside the existing custom-upload button.

**Key architectural decision (per #1774):** icon selections are stored as symbolic `lucide:<name>` references rather than data URIs. Resolution to a renderable form happens at render time, keeping config and export files human-readable and allowing the picker to highlight the current selection without reverse-engineering a blob.

Resolution at render time uses two paths depending on the consumer:

- **React components (`VertexIcon`)** — a new `useResolvedIconUrl` hook backed by React Query (`staleTime: Infinity`) converts `lucide:<name>` to a base64 SVG data URI on first use and caches the result per session. Plain URLs and `data:` URIs pass through via `initialData`, so they render immediately without an async round-trip.
- **Cytoscape canvas (`renderNode`)** — `getLucideSvgString` resolves the icon directly from the `lucide-react` dynamic-import map, skipping any network fetch.

Custom-uploaded icons (`data:` URIs) are completely unaffected — both resolvers pass them through unchanged.

The `IconPicker` component renders a searchable popover (8-column icon grid, 50-icon default window, type-to-filter). It emits the symbolic `lucide:<name>` ref so NodeStyleDialog stores the portable reference.

This commit also fixes a pre-existing lint-staged misconfiguration where the root `oxlint --fix` task matched `vitest.config.ts`, which oxlint then silently ignored (per its own `ignorePatterns`), causing a spurious exit-code 1 on pre-commit. The root lint-staged config now mirrors the oxlint ignore patterns.

This is **Slice 1** of the 3-slice split of #1589, per @kmcginnes review.

## Validation

All four local gates pass:

- `pnpm run checks` — oxlint, oxfmt, and TypeScript type-check all clean
- `pnpm test` — 1738 / 1738 tests pass (45 new tests across 4 new test files)
- `pnpm coverage` — all thresholds met; new files have ~100% line coverage (branches floor bumped 44 to 45 by the new branch coverage)
- `pnpm build` — production build completes without warnings

To validate the picker manually: open the Node Styling panel for any vertex type, click "Browse", search for an icon name (e.g. "plane"), select it, and confirm the node label updates in the graph view. Re-opening the dialog should show the selected icon highlighted in the grid.

## Related Issues

Addresses #1774
Slice 1 of #1589

## Check List

- [x] I have updated documentation if necessary. (`docs/features/graph-view.md`)


packages/graph-explorer/src/components/IconPicker.tsx
```diff
@@ -0,0 +1,157 @@
...
+    setOpen(false);
+    setSearch("");
+  }
+
+  useEffect(() => {
+    if (open) {
+      const timer = setTimeout(() => inputRef.current?.focus(), 100);
+      return () => clearTimeout(timer);
+    }
+  }, [open]);
```
kmcginnes:
```suggestion
```

I tried opening the picker in Chrome, Safari, and Firefox without this useEffect and they all had the search input focused by default. Was there a specific reason this was added?

packages/graph-explorer/src/components/IconPicker.tsx
```diff
@@ -0,0 +1,157 @@
...
+import { cn } from "@/utils";
+import {
+  getLucideName,
+  lucideIconToDataUri,
+  toLucideIconRef,
+} from "@/utils/lucideIconUrl";
+
+import { Button, Input, Popover, PopoverContent, PopoverTrigger } from ".";
+
+const allIconNames = Object.keys(dynamicIconImports).sort();
```
kmcginnes:
```suggestion
type IconName = keyof typeof dynamicIconImports;

const allIconNames = (Object.keys(dynamicIconImports) as IconName[]).toSorted();
```

Strongly type the icon names and use `toSorted()` since we want the sorted values to be returned.

packages/graph-explorer/src/components/IconPicker.tsx
```diff
@@ -0,0 +1,157 @@
...
+    </Popover>
+  );
+}
+
+function IconButton({
+  name,
+  selected,
+  onSelect,
+}: {
+  name: string;
```
kmcginnes:
```suggestion
  name: IconName;
```

Strongly type the icon name

packages/graph-explorer/src/components/IconPicker.tsx
```diff
@@ -0,0 +1,157 @@
...
+      {src ? (
+        <img src={src} alt={name} className="size-5" />
+      ) : (
+        <div className="bg-background-contrast-secondary size-5 animate-pulse rounded" />
+      )}
+    </button>
+  );
+}
+
+function filterIcons(search: string): string[] {
```
kmcginnes:
```suggestion
function filterIcons(search: string) {
```

Let the strongly typed names flow through.

packages/graph-explorer/src/components/IconPicker.tsx
```diff
@@ -0,0 +1,157 @@
...
+        "hover:bg-background-contrast-secondary flex size-8 items-center justify-center rounded",
+        selected && "bg-primary-main/20 ring-primary-main ring-2",
+      )}
+      onClick={() => onSelect(name)}
+    >
+      {src ? (
+        <img src={src} alt={name} className="size-5" />
+      ) : (
+        <div className="bg-background-contrast-secondary size-5 animate-pulse rounded" />
+      )}
```
kmcginnes:
```suggestion
      <DynamicIcon name={name} />
```

We should be able to use the Lucide provided [`DynamicIcon` component](https://lucide.dev/guide/react/advanced/dynamic-icon-component#dynamic-icon-component) so that we don't have to work so hard to show the icon here.

packages/graph-explorer/src/components/IconPicker.tsx
```diff
@@ -0,0 +1,157 @@
...
+  getLucideName,
+  lucideIconToDataUri,
+  toLucideIconRef,
+} from "@/utils/lucideIconUrl";
+
+import { Button, Input, Popover, PopoverContent, PopoverTrigger } from ".";
+
+const allIconNames = Object.keys(dynamicIconImports).sort();
+
+const MAX_VISIBLE = 50;
```
kmcginnes:
```suggestion
const MAX_VISIBLE = 64;
```

Good call on this. But let's pick a multiple of the number of columns so the grid is filled up. You have 8 columns, so let's go with 64 for 8 rows.

packages/graph-explorer/src/components/IconPicker.tsx
```diff
@@ -0,0 +1,157 @@
...
+    <button
+      type="button"
+      title={name}
+      aria-pressed={selected}
+      className={cn(
+        "hover:bg-background-contrast-secondary flex size-8 items-center justify-center rounded",
+        selected && "bg-primary-main/20 ring-primary-main ring-2",
+      )}
+      onClick={() => onSelect(name)}
+    >
```
kmcginnes:
```suggestion
    <Button
      title={name}
      aria-pressed={selected}
      size="icon"
      variant={selected ? "primary" : "ghost"}
      onClick={() => onSelect(name)}
      className="size-full min-w-0"
    >
```

While I like the way your icon buttons look, I'd really prefer to use the built in button styles.

The `size-full min-w-0` will make the button styles predefined sizes play nice with whatever grid size we pick.

packages/graph-explorer/src/components/IconPicker.tsx
```diff
@@ -0,0 +1,157 @@
...
+        align="start"
+        className="flex w-80 flex-col gap-2 p-3"
+      >
+        <Input
+          ref={inputRef}
+          placeholder="Search icons..."
+          value={search}
+          onChange={e => setSearch(e.target.value)}
+          className="h-8 text-sm"
+        />
```
kmcginnes, line 71:
```suggestion
        <Input
          ref={inputRef}
          placeholder="Search icons..."
          value={search}
          onChange={e => setSearch(e.target.value)}
        />
```

Let's keep the standard input size.

packages/graph-explorer/src/components/IconPicker.tsx
```diff
@@ -0,0 +1,157 @@
...
+  onSelect: (iconUrl: string, iconImageType: string) => void;
+}) {
+  const [open, setOpen] = useState(false);
+  const [search, setSearch] = useState("");
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  const selectedName = getLucideName(currentIconUrl);
+  const filtered = filterIcons(search);
+
+  function handleSelect(iconName: string) {
```
kmcginnes:
```suggestion
  function handleSelect(iconName: IconName) {
```

packages/graph-explorer/src/components/IconPicker.tsx
```diff
@@ -0,0 +1,157 @@
...
+}
+
+function IconButton({
+  name,
+  selected,
+  onSelect,
+}: {
+  name: string;
+  selected: boolean;
+  onSelect: (name: string) => void;
```
kmcginnes:
```suggestion
  onSelect: (name: IconName) => void;
```

packages/graph-explorer/src/components/IconPicker.tsx
```diff
@@ -0,0 +1,157 @@
...
+        <div className="bg-background-contrast-secondary size-5 animate-pulse rounded" />
+      )}
+    </button>
+  );
+}
+
+function filterIcons(search: string): string[] {
+  if (!search) return allIconNames.slice(0, MAX_VISIBLE);
+  const lower = search.toLowerCase();
+  const results: string[] = [];
```
kmcginnes:
```suggestion
  const results: IconName[] = [];
```

packages/graph-explorer/src/components/IconPicker.tsx
```diff
@@ -0,0 +1,157 @@
...
+        className="flex w-80 flex-col gap-2 p-3"
+      >
+        <Input
+          ref={inputRef}
+          placeholder="Search icons..."
+          value={search}
+          onChange={e => setSearch(e.target.value)}
+          className="h-8 text-sm"
+        />
+        <div className="grid max-h-60 grid-cols-8 gap-1 overflow-y-auto">
```
kmcginnes:
```suggestion
        <div className="grid size-80 grid-cols-8 grid-rows-8 gap-0.5">
```

Since we are capping the shown icons to `MAX_VISIBLE` we don't need to be concerned about the max height or overflow. I am choosing to set the grid size to `size-80` so it stays a consistent size no matter how many icons are shown in the grid.

packages/graph-explorer/src/components/IconPicker.tsx
```diff
@@ -0,0 +1,157 @@
...
+              name={name}
+              selected={name === selectedName}
+              onSelect={handleSelect}
+            />
+          ))}
+          {filtered.length === 0 && (
+            <p className="text-text-secondary col-span-8 py-4 text-center text-sm">
+              No icons found
+            </p>
+          )}
```
kmcginnes, line 90:
```suggestion
          {filtered.length === 0 && (
            <EmptyState className="col-span-8 row-span-8" size="small">
              <EmptyStateContent>
                <EmptyStateTitle>No icons found</EmptyStateTitle>
                <EmptyStateDescription className="text-balance">
                  No matching icons found. Try a broader search.
                </EmptyStateDescription>
              </EmptyStateContent>
            </EmptyState>
          )}
```

Using the built in empty state components we can get most of the styling for free. And by spanning the columns and rows it is centered within the empty grid space.

Image

packages/graph-explorer/src/components/IconPicker.tsx
```diff
@@ -0,0 +1,157 @@
...
+      <PopoverTrigger asChild>
+        <Button variant="outline" className="rounded-full">
+          <SearchIcon className="size-4" />
+          Browse
+        </Button>
+      </PopoverTrigger>
+      <PopoverContent
+        side="bottom"
+        align="start"
+        className="flex w-80 flex-col gap-2 p-3"
```
kmcginnes:
```suggestion
        className="flex flex-col gap-4"
```

This will better match the color picker padding.

packages/graph-explorer/src/utils/useResolvedIconUrl.ts
```diff
@@ -0,0 +1,22 @@
...
+ */
+export function useResolvedIconUrl(iconUrl: string): string {
+  const { data } = useQuery({
+    queryKey: ["resolved-icon", iconUrl],
+    queryFn: () => resolveIconUrl(iconUrl),
+    staleTime: Infinity,
+    initialData: isLucideIconRef(iconUrl) ? undefined : iconUrl,
+  });
+  return data ?? "";
+}
```
kmcginnes:
```suggestion
export function useResolvedIconUrl(iconUrl: string) {
  const { data } = useQuery({
    queryKey: ["resolved-icon", iconUrl],
    queryFn: () => resolveIconUrl(iconUrl),
    staleTime: Infinity,
    initialData: isLucideIconRef(iconUrl) ? undefined : iconUrl,
  });
  return data;
}
```
Here, I've changed the result to return the null/undefined values rather than coalesce to empty string. The consumer will have to guard against empty values.

packages/graph-explorer/src/utils/lucideIconUrl.ts
```diff
@@ -0,0 +1,141 @@
+import dynamicIconImports from "lucide-react/dynamicIconImports";
+
+type IconNodeChild = [string, Record<string, string>];
```
kmcginnes:
```suggestion
type IconName = keyof typeof dynamicIconImports;

const allIconNames = new Set(Object.keys(dynamicIconImports));

type IconNodeChild = [string, Record<string, string>];
```

Add `IconName` for stronger types and `allIconNames` lookup for checking if a string is a valid icon name for lucide.

packages/graph-explorer/src/utils/lucideIcons.ts
```diff
@@ -0,0 +1,141 @@
...
+ * A stored value like `lucide:plane` preserves the symbolic icon name so
+ * the picker can highlight the current selection and config/export files
+ * can carry human-readable icon references.
+ */
+export const LUCIDE_PREFIX = "lucide:";
+
+/** True if `iconUrl` is a stored Lucide reference. */
+export function isLucideIconRef(iconUrl: string | undefined): boolean {
+  return !!iconUrl && iconUrl.startsWith(LUCIDE_PREFIX);
+}
```
kmcginnes, line 28:
```suggestion
/** True if `iconUrl` is a stored Lucide reference. */
export function isLucideIconRef(iconUrl: string | undefined): boolean {
  return !!iconUrl && iconUrl.startsWith(LUCIDE_PREFIX);
}

/** Type guard that narrows a string to a valid Lucide icon name. */
export function isValidLucideIconName(name: string): name is IconName {
  return !!name && allIconNames.has(name);
}
```

Add a type guard that we can use in the VertexIcon component to narrow the name type and use `DynamicIcon` from Lucide React.

packages/graph-explorer/src/components/VertexIcon.tsx
```diff
@@ -19,11 +20,12 @@ interface Props {
 
 function VertexIcon({ vertexStyle, className, alt }: Props) {
   const altText = alt ?? `${vertexStyle.displayLabel ?? vertexStyle.type} icon`;
+  const resolvedSrc = useResolvedIconUrl(vertexStyle.iconUrl);
```
kmcginnes:
```suggestion
  const lucideIconName = getLucideName(vertexStyle.iconUrl);

  if (lucideIconName) {
    if (isValidLucideIconName(lucideIconName)) {
      return <DynamicIcon name={lucideIconName} className="size-6 shrink-0" />;
    } else {
      return null;
    }
  }
```

Instead of building the SVG string for the VertexIcon, we can simply use `DynamicIcon` from Lucide React. We just have to narrow the name type to a valid lucide icon name, otherwise we should show nothing (i.e. `null`).

This removes the only use of `useResolvedIconUrl()` so we can clean that up along with some of the helpers.

packages/graph-explorer/src/components/VertexIcon.tsx
```diff
@@ -33,7 +35,7 @@ function VertexIcon({ vertexStyle, className, alt }: Props) {
 
   return (
     <img
-      src={vertexStyle.iconUrl}
+      src={resolvedSrc}
```
kmcginnes:
```suggestion
      src={vertexStyle.iconUrl}
```

Since the Lucide case returns early, these are just the iconUrl passed through.

packages/graph-explorer/src/components/VertexIcon.tsx
```diff
@@ -19,11 +20,12 @@ interface Props {
 
 function VertexIcon({ vertexStyle, className, alt }: Props) {
   const altText = alt ?? `${vertexStyle.displayLabel ?? vertexStyle.type} icon`;
+  const resolvedSrc = useResolvedIconUrl(vertexStyle.iconUrl);
 
   if (vertexStyle.iconImageType === "image/svg+xml") {
     return (
       <SVG
-        src={vertexStyle.iconUrl}
+        src={resolvedSrc}
```
kmcginnes:
```suggestion
        src={vertexStyle.iconUrl}
```

Since the Lucide case returns early, these are just the `iconUrl` passed through.

docs/features/graph-view.md
```diff
@@ -80,7 +80,7 @@ Each node type can be customized in a variety of ways.
 - **Display label** allows you to change how the node label (or rdf:type) is represented
 - **Display name attribute** allows you to choose the attribute on the node that is used to uniquely label the node in the graph visualization and search
 - **Display description attribute** allows you to choose the attribute on the node that is used to describe the node in search
-- **Custom symbol** can be uploaded in the form of an SVG icon
+- **Icon** can be picked from the built-in Lucide library via the **Browse** button, or uploaded as a custom SVG/raster image. Picked Lucide icons are stored as `lucide:<name>` references and resolved at render time, so the picker highlights the currently selected icon when you reopen the dialog.
```
kmcginnes:
```suggestion
- **Icon** can be picked from the built-in Lucide library via the **Browse** button, or uploaded as a custom SVG/raster image.
```

Simplified to just the user facing behavior.

packages/graph-explorer/src/utils/lucideIcons.ts
```diff
@@ -0,0 +1,141 @@
...
+ * Used by render pipelines (e.g., Cytoscape) that need the SVG text rather
+ * than a data URI.
+ */
+export async function getLucideSvgString(
+  iconName: string,
+): Promise<string | null> {
+  const iconNode = await getLucideIconNode(iconName);
+  if (!iconNode) return null;
+  return buildSvgString(iconNode);
+}
```
kmcginnes, line 63:
```suggestion
export async function getLucideSvgString(
  iconName: string,
): Promise<string | null> {
  if (!isValidLucideIconName(iconName)) {
    return null;
  }
  const { default: Icon } = await dynamicIconImports[iconName]();
  return renderToStaticMarkup(createElement(Icon));
}
```

So Claude taught me a new trick I had never considered. We can use the React server API `renderToStaticMarkup()` to essentially create the SVG for us from the React element that `dynamicIconImports` gives us. This eliminates all the SVG building code you had before. This also removes the need for the iconCache.

Vite's tree shaking will ensure we just pull in `renderToStaticMarkup` to the final bundle, so we don't get the entire `react-dom/server` module bloating our bundle size.

kmcginnes (changes requested):
This is looking pretty good. Thanks for splitting it up for me. I've provided a bunch of changes you can easily take, but there is likely some cleanup needed afterward. I think once these changes go in this PR will be much simpler. And feel free to push back on any of my suggested changes if you disagree.

One other feature that would be nice, but is not strictly necessary, is paging within the icon picker. I don't expect many people to browse through 60 pages of icons, but if they are searching and get 100 results I could see paging being useful.

Long term (not right now), I imagine we will add icon categories so the user can more easily narrow in on a specific icon they might want without having to know the name, similar to Apple's emoji picker. And if we wanted to be extra fancy we could do vertical scrolling in the grid with a virtualized list for performance instead of using paging. But paging is simpler and more pragmatic for now.

kmcginnes:
@jkemmererupgrade FYI, I had some high priority issues come up which has delayed this a bit. Once I knock those out I'll circle back to this.

kmcginnes:
@jkemmererupgrade I'm running through some final tweaks. But this is looking good.

kmcginnes (dismissed):

kmcginnes (approved):

jkemmererupgrade:
Thank you @kmcginnes!

PR #1781, opened 2026-05-23
Fix default icon for no-type vertices

### LINKED ISSUE #1779: Blank nodes show no icon on graph canvas but display default icon in details panel
## Description

When a blank node (bnode) with no type is rendered on the graph canvas, it has no icon. However, when the same node is selected, the details/properties panel shows the default icon. The icon display should be consistent between the two views.

The root cause is that the canvas icon rendering path (`useBackgroundImageMap`) only creates Cytoscape style selectors for vertex types that exist in the schema. Since `"«No Type»"` is a synthetic label assigned at runtime (`vertex.ts:80-83`) and isn't in the schema, no icon style is generated. The details panel uses `useVertexPreferences()` which always falls back to `DEFAULT_ICON_URL` regardless of schema presence.

### Relevant code paths

- Canvas: `useGraphStyles` → `useBackgroundImageMap` → `renderNode` (skips types not in schema)
- Panel: `VertexIcon` → `useVertexPreferences` → always returns `defaultVertexPreferences.iconUrl`

## Environment

- Graph Database & Version: Any SPARQL endpoint with blank nodes

## Steps to Reproduce

1. Load a graph containing a blank node (e.g., a bnode used as an address object)
2. Observe the bnode on the canvas — no icon, just the "«No Type»" label
3. Select the bnode and open the properties panel
4. The default icon appears in the panel header but not on the canvas node

Screenshot showing bnode with no icon on canvas but default icon in details panel

## Expected Behavior

The blank node should display the same default icon on both the canvas and the details panel.

---

> [!IMPORTANT]

> [!TIP]

P-r-e-m-i-u-m:
## Summary

Fixes #1779.

This updates the graph style preferences so the synthetic `No Type` vertex type gets the same default vertex styling as schema-backed types. That gives blank nodes without a schema type a generated Cytoscape background image, matching the default icon shown in the details panel.

I also added a regression test for the no-type node style selector.

## Verification

- Focused GraphViewer style test passed
- Formatter check passed on touched files
- Oxlint passed on touched files
- Recursive workspace typecheck passed

I also ran the full Vitest suite locally. The touched GraphViewer test passed, but the full run has unrelated Windows/environment failures: path assertions expecting POSIX separators, missing openssl, shell redirects failing because the workspace path contains spaces, and locale-specific number grouping under en-IN.

### DIFF
--- a/packages/graph-explorer/src/core/StateProvider/userPreferences.ts
+++ b/packages/graph-explorer/src/core/StateProvider/userPreferences.ts
@@ -4,11 +4,10 @@ import { atom, useAtomValue, useSetAtom } from "jotai";
 import { atomFamily } from "jotai-family";
 import { useDeferredValue } from "react";
 
-import { RESERVED_ID_PROPERTY, RESERVED_TYPES_PROPERTY } from "@/utils";
+import { LABELS, RESERVED_ID_PROPERTY, RESERVED_TYPES_PROPERTY } from "@/utils";
 import DEFAULT_ICON_URL from "@/utils/defaultIconUrl";
 
-import type { EdgeType, VertexType } from "../entities";
-
+import { createVertexType, type EdgeType, type VertexType } from "../entities";
 import { useActiveSchema } from "./schema";
 import { userStylingAtom } from "./storageAtoms";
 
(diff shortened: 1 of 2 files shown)


kmcginnes:
@P-r-e-m-i-u-m Hi, sorry I missed this before reviewing and merging PR #1799, which is essentially the same fix.

But I'd love the have the test you've added here. If you can get this rebased on latest I'd be happy to merge this in.

PR #1800, opened 2026-06-02
fix: use label().groupCount() for neighbor counts on Neptune 1.3.x

### LINKED ISSUE #1798: Neighbor count query fails on Neptune engine 1.3.x
## Description

The neighbor count Gremlin query crashes with `InternalFailureException` on Neptune engine 1.3.2.1. The query works on engine 1.4.7.0.

## Root Cause

PR #1048 batched neighbor count requests into a single query by wrapping with `.group().by(id).by(...)`:

```gremlin
g.V("1","2","3")
  .group().by(id)
  .by(
    both().dedup().group().by(label).by(count())
  )
```

The nested `group().by(label).by(count())` inside an outer `by()` triggers an internal failure on Neptune 1.3.x. The old single-vertex query (`g.V("1").both().dedup().group().by(label).by(count())`) worked fine because `group().by().by()` wasn't nested.

## Fix

Replace the inner traversal with `label().groupCount()`:

```gremlin
g.V("1","2","3")
  .group().by(id)
  .by(
    both().dedup().label().groupCount()
  )
```

This produces the same response format (g:Map of label string → count) and works on both 1.3.x and 1.4.x engines. No parser changes needed — the client-side `::` splitting for multi-label support works identically.

## Reproduction

```bash
# Fails on Neptune 1.3.2.1
aws neptunedata execute-gremlin-query \
  --endpoint-url https://<neptune-1.3.x>:8182 \
  --gremlin-query 'g.V("1").group().by(id).by(both().dedup().group().by(label).by(count()))' \
  --region us-west-2

# Works on same engine
aws neptunedata execute-gremlin-query \
  --endpoint-url https://<neptune-1.3.x>:8182 \
  --gremlin-query 'g.V("1").group().by(id).by(both().dedup().label().groupCount())' \
  --region us-west-2
```

## Affected Code

`packages/graph-explorer/src/connector/gremlin/neighborCounts.ts` lines 50-57

vishwakt:
## Description

The batched neighbor count Gremlin query crashes with `InternalFailureException` on Neptune engine 1.3.2.1, while working on 1.4.7.0.

## Root Cause

PR #1048 batched neighbor count requests by wrapping with `.group().by(id).by(...)`. The inner traversal nests `group().by(label).by(count())` inside the outer `by()`:

```gremlin
g.V("1","2","3")
  .group().by(id)
  .by(
    both().dedup().group().by(label).by(count())
  )
```

That nested `group().by().by()` triggers an internal failure on Neptune 1.3.x. The pre-batching single-vertex query worked because the grouping was not nested.

## Fix

Replace the inner traversal with `label().groupCount()`:

```gremlin
g.V("1","2","3")
  .group().by(id)
  .by(
    both().dedup().label().groupCount()
  )
```

This emits the same response format (a `g:Map` of label string to count) and works on both 1.3.x and 1.4.x engines. No parser changes are needed, and the client-side `::` splitting for multi-label support behaves identically.

## Testing

- Updated the two query template assertions in `neighborCounts.test.ts` to match the new traversal.
- All `neighborCounts` unit tests pass. The data-shape tests are unchanged because the response format is identical.

Note: the connector tests mock the fetch layer, so they validate the query string and parsing but cannot exercise a live 1.3.x engine. The engine-level fix is verified by the reproduction in #1798.

Fixes #1798

### DIFF
--- a/packages/graph-explorer/src/connector/gremlin/neighborCounts.test.ts
+++ b/packages/graph-explorer/src/connector/gremlin/neighborCounts.test.ts
@@ -106,7 +106,7 @@ describe("neighborCounts", () => {
        .group()
        .by(id)
        .by(
-         both().dedup().group().by(label).by(count())
+         both().dedup().label().groupCount()
        )
     `);
   });
@@ -123,7 +123,7 @@ describe("neighborCounts", () => {
        .group()
        .by(id)
        .by(
-         both().dedup().group().by(label).by(count())
+         both().dedup().label().groupCount()
        )
     `);
   });
--- a/packages/graph-explorer/src/connector/gremlin/neighborCounts.test.ts
+++ b/packages/graph-explorer/src/connector/gremlin/neighborCounts.test.ts
@@ -106,7 +106,7 @@ describe("neighborCounts", () => {
        .group()
        .by(id)
        .by(
-         both().dedup().label().groupCount()
+         both().dedup().groupCount().by(label)
        )
     `);
   });
@@ -123,7 +123,7 @@ describe("neighborCounts", () => {
        .group()
        .by(id)
        .by(
-         both().dedup().label().groupCount()
+         both().dedup().groupCount().by(label)
        )
     `);
   });
--- a/packages/graph-explorer/src/connector/gremlin/neighborCounts.ts
+++ b/packages/graph-explorer/src/connector/gremlin/neighborCounts.ts
@@ -52,7 +52,7 @@ export async function neighborCounts(
      .group()
      .by(id)
      .by(
-       both().dedup().label().groupCount()
+       both().dedup().groupCount().by(label)
      )
   `;
 
--- a/packages/graph-explorer/src/connector/gremlin/neighborCounts.ts
+++ b/packages/graph-explorer/src/connector/gremlin/neighborCounts.ts
@@ -52,7 +52,7 @@ export async function neighborCounts(
      .group()
      .by(id)
      .by(
-       both().dedup().group().by(label).by(count())
+       both().dedup().label().groupCount()
      )
   `;
 


vishwakt:
I verified functional and wire-format equivalence of the two query forms locally on Apache TinkerPop 3.7.3 (Gremlin Console, standard "modern" toy graph, which has `person`/`software` vertex labels).

Both the old `both().dedup().group().by(label).by(count())` and the new `both().dedup().label().groupCount()` produce byte-identical GraphSON 3.0 output, e.g. for one vertex:

​```json
{"@type":"g:Map","@value":["software",{"@type":"g:Int64","@value":1},"person",{"@type":"g:Int64","@value":2}]}
​```

This is the same `g:Map` shape Neptune returns and that `parseGMap` consumes, which confirms the "same response format, no parser changes" claim empirically.

Caveat: TinkerPop's reference engine runs both forms without error, so this does not reproduce the Neptune 1.3.x `InternalFailureException` itself. That is specific to Neptune's DFE engine and there's no local Neptune build to test against, so the engine-level fix still warrants a quick check on a 1.3.x cluster if you have one handy.

packages/graph-explorer/src/connector/gremlin/neighborCounts.test.ts
```diff
@@ -106,7 +106,7 @@ describe("neighborCounts", () => {
        .group()
        .by(id)
        .by(
-         both().dedup().group().by(label).by(count())
+         both().dedup().label().groupCount()
```
kmcginnes:
```suggestion
         both().dedup().groupCount().by(label)
```

packages/graph-explorer/src/connector/gremlin/neighborCounts.test.ts
```diff
@@ -123,7 +123,7 @@ describe("neighborCounts", () => {
...
```
kmcginnes:
```suggestion
         both().dedup().groupCount().by(label)
```

packages/graph-explorer/src/connector/gremlin/neighborCounts.ts
```diff
@@ -52,7 +52,7 @@ export async function neighborCounts(
      .group()
      .by(id)
      .by(
-       both().dedup().group().by(label).by(count())
+       both().dedup().label().groupCount()
```
kmcginnes:
```suggestion
       both().dedup().groupCount().by(label)
```

kmcginnes (changes requested):
My team had some feedback on the fix query structure. They suggested using 

```
.groupCount().by(label)
```

instead of

```
.label().groupCount()
```

kmcginnes:
Manually tested the updated suggested queries:

```
.groupCount().by(label)
```

on Neptune v1.2.1.0, v1.3.2.1, v1.4.7.0

vishwakt:
Thanks @kmcginnes, i've made the suggested updates in a new commit. Thank you for testing on real engines. 

Updated to `groupCount().by(label)`it's also consistent with the existing `groupCount().by(label)` usage in the vertex/edge label templates.

I re-ran my local equivalence check and the new form is byte-identical to the original on the air-routes-style toy graph, so the parser path is unaffected.

kmcginnes (approved):

PR #1811, opened 2026-06-08
Feat: Activate or create connection via URL parameters

### LINKED ISSUE #1788: Activate or create a connection via URL parameters
## Problem Statement

Applications that use graph databases have no way to launch Graph Explorer pre-connected to their database. Users must manually open Graph Explorer, create a connection, and configure it — friction that discourages cross-app integration.

## Solution

Graph Explorer accepts a dedicated `#/connect` route whose query parameters specify a database connection. On entry it either activates a matching existing Connection or prompts the user to create a new one, then redirects to the graph view ready to explore.

Example link from an external app:

```
https://my-ge-host/#/connect?graphDbUrl=https%3A%2F%2Fmy-cluster.us-east-1.neptune.amazonaws.com%3A8182&queryEngine=gremlin&awsRegion=us-east-1&serviceType=neptune-db&name=My%20Database
```

> **Note:** The parameters belong to the `#/connect` route, so they go *after* the `#`. Graph Explorer uses hash-based routing, so integrators build the link the same way they would any in-app link. The `graphDbUrl` value must be URL-encoded (e.g., `https://host:8182` becomes `https%3A%2F%2Fhost%3A8182`). Most languages provide this via `encodeURIComponent()` (JavaScript), `urllib.parse.quote()` (Python), `URLEncoder.encode()` (Java), etc.

## User Stories

1. As an external app developer, I want to link users to Graph Explorer with a database connection pre-activated, so that users can explore their graph data without manual configuration.
2. As an external app developer, I want to specify the database URL, query language, AWS region, service type, and display name in the link, so that the connection is configured correctly for my database.
3. As a Graph Explorer user, I want to be prompted before a new connection is created from a link, so that arbitrary links cannot silently modify my saved connections.
4. As a Graph Explorer user, I want to edit the connection name in the creation prompt, so that I can personalize how the connection appears in my list.
5. As a Graph Explorer user, I want my existing graph Session to be preserved when a link targets the connection I am *already* connected to, so that my exploration work is not lost when nothing actually changes.
6. As a Graph Explorer user, I want the connect URL to leave no trace in my history, so that refreshing the page or pressing back behaves normally.
7. As a Graph Explorer user, I want only `graphDbUrl` to be required in the link, so that simple integrations need minimal configuration (query language defaults to Gremlin, IAM auth assumed off when region is absent).
8. As a Graph Explorer user, I want duplicate connections to not be created when I click the same external link multiple times, so that my connection list stays clean.
9. As a Graph Explorer user, I want the link processed after default connections load, so that a link matching a default connection does not prompt me to create a duplicate.
10. As an external app developer, I want the matching logic to find my connection even if the user has renamed it in Graph Explorer, so that name changes don't break integration links.
11. As a Graph Explorer user with multiple connections to the same database, I want the `name` parameter to disambiguate which connection activates, so that the correct one is chosen.
12. As a Graph Explorer user, I want a link requesting IAM auth to never silently reuse a plaintext connection to the same endpoint (or the reverse), so that I always know how a connection authenticates.
13. As a Graph Explorer user, I want a link carrying values Graph Explorer cannot honor to be rejected outright rather than corrected for me, so that a link never quietly connects me with settings it did not ask for.
14. As a Graph Explorer user, I want to be told which part of a bad link was wrong, so that I can correct it or report it to whoever sent it.

## Implementation Decisions

- **URL parameter location:** Query parameters on a dedicated `#/connect` route, so they sit *after* the `#` like every other route. This keeps the parameters inside the router (no `window.location` reads, no manual param stripping) and makes the link format the same shape integrators already use for in-app links. This replaces an earlier decision to put the parameters before the `#`; see [the ADR](https://github.com/aws/graph-explorer/blob/main/docs/adr/20260612-connection-links.md) for the reasoning and the trade it makes.
- **Accepted parameters:** `graphDbUrl` (required, must be URL-encoded, must be `http`/`https`, must not carry a username or password), `queryEngine` (optional, default "gremlin"), `awsRegion` (optional), `serviceType` (optional), `name` (optional). Providing `awsRegion` enables IAM auth; without it IAM auth is off.
- **Invalid values are rejected, not coerced:** An unsupported `queryEngine` or `serviceType` invalidates the link rather than falling back to a default. Answering `queryEngine=sql` with Gremlin would build a connection that queries the database in a language the caller never asked for. A rejected link is ignored and the user is shown which parameter was at fault and what it requires.
- **Credentials in the URL are rejected:** `fetch` itself refuses a URL carrying userinfo (the `Request` constructor throws), so such a link could only ever produce a connection that fails every query, after persisting the password to IndexedDB and into any exported connection file. Graph Explorer authenticates with IAM, never userinfo.
- **Processing location:** The `#/connect` route resolves the link once on entry and acts on the result. It still runs after default connections have loaded, because `AppStatusLoader` gates the route behind a loading state while the connection store is empty. Order: preload IndexedDB → fetch default connections → resolve the link → redirect.
- **Matching logic:** Filter connections by `graphDbUrl` (case-insensitive) + `queryEngine` + auth posture. Auth posture is identity-bearing: whether IAM is on, and when on, the region and service type. If multiple match, prefer the already-active connection, then tiebreak by `name`, then fall back to first found. If zero match, prompt for creation.
- **Activation behavior:** A link matching a different connection switches to it silently, with no prompt. This is the same operation as clicking that connection in the connections list, which has never asked for confirmation, and the link only ever activates a connection the user already created. There is also no session data at risk: Sessions are stored per connection, so switching swaps which Session is displayed rather than destroying the previous one. A link cannot change an already-open window's state either, since following one opens a new tab or the user pastes it deliberately.
- **Creation requires user consent:** When no match is found, the create-connection form opens pre-filled and fully editable. Saving creates the connection and activates it; closing or cancelling creates nothing. Either way the user ends on the graph view. This is the trust gate for the untrusted endpoint details a link can carry, and the only path that can introduce a new database.
- **Connection naming:** Use the `name` parameter when given, otherwise derive one from the `graphDbUrl` hostname. The derived name is part of the link's identity, not just a display fallback: a nameless link identifies the connection it would have created, so reopening it returns to that connection even after the user adds a second connection to the same endpoint under a name of their own. The name is editable in the create form.
- **No parameter stripping:** The route redirects with `replace`, so the `#/connect` entry never enters history. This replaces the earlier `window.history.replaceState` approach.
- **Landing page:** Every outcome redirects to the graph view. Naming the landing page is deferred to #2164, along with the cancel-returns-you-where-you-were behavior that depended on it.
- **Scope:** New tab/window navigation only. No iframe/postMessage support.
- **Subsequent navigations:** Opening another link with different parameters is a hash change, which the router handles as a normal navigation rather than a full page reload.

### Modules

1. **Connection link reader** — Pure function that reads a route's search string and returns one of three outcomes: no link present, an invalid link carrying a problem per offending parameter, or a valid link with typed parameters. Handles URL decoding, required vs optional fields, and defaults.
2. **Connection matcher** — Pure function over the parsed parameters, the existing connections, and the active connection id, returning the matched connection or `null`. Encapsulates the full priority logic (filter by graphDbUrl + queryEngine + auth posture → prefer active → tiebreak by name → first found).
3. **Intent resolver** — Pure function that folds a link plus the current connections into one of four intents: do nothing (no link, or it targets the active connection), activate a matching connection, create a new one seeded from the link, or reject an invalid link. Callers dispatch on the intent rather than juggling match and pending flags.
4. **Connect route** — Resolves the link once on entry via the resolver, then acts: activates silently, notifies and ignores an invalid link, or renders the create form. Every outcome except the create form redirects immediately.
5. **Create form prefill** — The existing create-connection form gains the ability to open pre-filled with fully editable values, without entering its "edit an existing connection" mode.

## Testing Decisions

Good tests for this feature verify external behavior (given this link and this connection state, what connection is activated, what is shown) rather than implementation details (which atoms were set, which internal functions were called).

### Modules to test

1. **Connection link reader** — Valid parameter combinations, missing required parameters, defaults applied, URL decoding, and each rejection rule with the problem it reports. Prior art: `defaultConnection.test.ts` (schema validation pattern with `DefaultConnectionDataSchema`).
2. **Connection matcher** — Single match, multiple-match disambiguation (active preferred, then name tiebreak, then first), auth posture mismatch falling through to no match, zero matches returning null, case-insensitive URL comparison. Prior art: `exportedGraph.test.ts` (`isMatchingConnection` tests).
3. **Intent resolver** — Each of the four intents, including a link matching the active connection resolving to a no-op and an auth-posture mismatch resolving to create.
4. **Connect route** — Integration tests for each intent: silent activation with redirect, create form pre-filled, no-op redirect, and an invalid link that notifies and redirects without prompting. Also the ordering case: a link matching a default connection that is still loading must not prompt to create a duplicate.
5. **Create form prefill** — Renders with the given values, stays in "add" mode, and creates the connection on submit.

## Out of Scope

- Naming the landing page, and returning the user to where they were on cancel (moved to #2164)
- Iframe embedding or `postMessage`-based communication with parent applications
- Proxy server configuration parameters (removed per PR #1773)
- Sharing or encoding graph Session state (vertices/edges) in URL parameters
- Authentication token passing via URL, including credentials in `graphDbUrl` (security risk)
- Bidirectional communication (GE notifying the external app of user actions)
- Multi-connection activation (only one connection at a time)
- `fetchTimeoutMs` or `nodeExpansionLimit` as URL parameters (too implementation-specific for external callers)

## Further Notes

- This feature depends on PR #1773 landing first, which simplifies the Connection model to use `graphDbUrl` as the single required field.
- The `isMatchingConnection` function in `exportedGraph.ts` is prior art for matching logic, but the new matcher module should be independent since post-#1773 the connection model changes.
- Link-driven activation reuses today's manual connection-switching behavior. There is no special "activate without reset" path: the in-memory graph is cleared on switch exactly as it is when switching by hand, and the previous connection's Session remains persisted under its own key.
- External apps must URL-encode parameter values. The `graphDbUrl` in particular contains characters (`:`, `/`) that are invalid in query string values without encoding.
- Design decisions and the reasoning behind them live in `docs/adr/20260612-connection-links.md`.

## Related Issues

- Implemented by #1828
- Replaces #1811
- Landing page deferred to #2164

andy-k-improving:
**Issue:** Closes #1788

**Description:**

Allows external applications to deep-link into Graph Explorer with a database connection pre-configured. On page load, URL search parameters are parsed and the user is prompted to either activate an existing matching connection or create a new one.

### Example URL

```
https://localhost/explorer?graphDbUrl=https%3A%2F%2Fg-xxx.us-west-2.neptune-graph.amazonaws.com&queryEngine=openCypher&awsRegion=us-west-2&serviceType=neptune-graph&name=My%20Graph
```

### Supported parameters

| Parameter | Required | Default | Description |
|-----------|----------|---------|-------------|
| `graphDbUrl` | Yes | — | Database endpoint (URL-encoded) |
| `queryEngine` | No | `gremlin` | `gremlin`, `openCypher`, or `sparql` |
| `awsRegion` | No | — | AWS region for IAM signing |
| `serviceType` | No | — | `neptune-db` or `neptune-graph` |
| `name` | No | graphDbUrl | Display label for the connection |

### Behavior

- **Match found** → dialog shows "Activate Connection"; clicking Connect activates without resetting the graph session
- **No match** → dialog shows "Create Connection"; clicking Create adds and activates
- **Cancel** → dismisses dialog, strips URL params, no changes made
- **Dedup** → deterministic ID prevents duplicates from repeated URL clicks
- **URL stripped** → params removed via `replaceState` after user confirms or cancels

### Files changed

| File | Description |
|------|-------------|
| `src/core/AppStatusLoader.tsx` | Orchestration: parse params at module level, compute match/pending during render, delegate to dialog |
| `src/core/UrlConnectionDialog.tsx` | Presentational dialog component |
| `src/core/urlConnectionParams.ts` | Pure functions: `parseUrlConnectionParams`, `findMatchingConnection`, `buildConnectionFromParams` |
| `src/core/urlConnectionParams.test.ts` | Unit tests (9 cases) |

### Testing

- Unit tests: URL parsing (missing params, defaults, full params), matching (case-insensitive, engine mismatch, no match), building (IAM on/off, deterministic IDs)
- Manual: new endpoint → Create dialog; existing endpoint → Connect dialog; no params → normal flow

### Check List

Screenshot 2026-06-08 at 3 57 13 PM
Screenshot 2026-06-08 at 4 42 04 PM

### DIFF
--- a/packages/graph-explorer/src/core/AppStatusLoader.tsx
+++ b/packages/graph-explorer/src/core/AppStatusLoader.tsx
@@ -4,31 +4,23 @@ import {
   type PropsWithChildren,
   startTransition,
   Suspense,
-  useEffect,
   useState,
+  useEffect,
 } from "react";
 
 import { PanelEmptyState, Spinner } from "@/components";
-import { Button } from "@/components/Button/Button";
-import {
-  Dialog,
-  DialogContent,
-  DialogHeader,
-  DialogTitle,
-  DialogBody,
-  DialogFooter,
-} from "@/components/Dialog";
 import { logger } from "@/utils";
 
 import { fetchDefaultConnection } from "./defaultConnection";
 import { activeConfigurationAtom, configurationAtom } from "./StateProvider";
+import { UrlConnectionDialog } from "./UrlConnectionDialog";
 import {
   parseUrlConnectionParams,
   findMatchingConnection,
   buildConnectionFromParams,
 } from "./urlConnectionParams";
 
-// Read URL params once at module level (before any render)
+// Read URL params once at module level
 const initialUrlParams = parseUrlConnectionParams(window.location.search);
 
 function AppStatusLoader({ children }: PropsWithChildren) {
@@ -42,7 +34,7 @@ function AppStatusLoader({ children }: PropsWithChildren) {
 function LoadDefaultConfig({ children }: PropsWithChildren) {
   const [activeConfig, setActiveConfig] = useAtom(activeConfigurationAtom);
   const [configuration, setConfiguration] = useAtom(configurationAtom);
-  const [urlHandled, setUrlHandled] = useState(false);
+  const [urlHandled, setUrlHandled] = useState(!initialUrlParams);
 
   const defaultConfigQuery = useQuery({
     queryKey: ["default-connection"],
@@ -82,36 +74,52 @@ function LoadDefaultConfig({ children }: PropsWithChildren) {
     defaultConnectionConfigs,
   ]);
 
-  // Handle existing match activation via effect
-  useEffect(() => {
-    if (urlHandled || !initialUrlParams) return;
+  // Compute dialog state during render
+  const existingMatch =
+    !urlHandled && initialUrlParams
+      ? findMatchingConnection(configuration, initialUrlParams)
+      : null;
+  const pendingConnection =
+    !urlHandled && initialUrlParams && !existingMatch
+      ? buildConnectionFromParams(initialUrlParams, window.location.origin)
+      : null;
 
-    const existingMatch = findMatchingConnection(
-      configuration,
-      initialUrlParams,
-    );
-    if (!existingMatch) return;
-
-    logger.debug("Found matching connection from URL params", existingMatch.id);
-    startTransition(() => {
-      setUrlHandled(true);
-      setActiveConfig(existingMatch.id);
-    });
+  function handleConfirm() {
+    if (existingMatch) {
+      logger.debug(
+        "Activating matching connection from URL params",
+        existingMatch.id,
+      );
+      startTransition(() => {
+        setActiveConfig(existingMatch.id);
+      });
+    } else if (pendingConnection) {
+      logger.debug("Adding connection from URL params", pendingConnection);
+      startTransition(() => {
+        setConfiguration(prev => {
+          const updated = new Map(prev);
+          updated.set(pendingConnection.id, pendingConnection);
+          return updated;
+        });
+        setActiveConfig(pendingConnection.id);
+      });
+    }
+    setUrlHandled(true);
     window.history.replaceState(
       {},
       "",
       window.location.pathname + window.location.hash,
     );
-  }, [urlHandled, configuration, setActiveConfig]);
+  }
 
-  // Determine if we need to show the dialog
-  const existingMatch = initialUrlParams
-    ? findMatchingConnection(configuration, initialUrlParams)
-    : null;
-  const pendingConnection =
-    initialUrlParams && !existingMatch && !urlHandled
-      ? buildConnectionFromParams(initialUrlParams, window.location.origin)
-      : null;
+  function handleCancel() {
+    setUrlHandled(true);
+    window.history.replaceState(
+      {},
+      "",
+      window.location.pathname + window.location.hash,
+    );
+  }
 
   if (configuration.size === 0 && defaultConfigQuery.isLoading) {
     return (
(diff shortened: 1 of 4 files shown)


kmcginnes:
> - **Match found** → dialog shows "Activate Connection"; clicking Connect activates without resetting the graph session

You must reset the graph session when changing connections.

The reason is that any graph currently on the canvas has the potential to be expanded or the data to be refreshed. But if the connection changed, then the data might not exist in the new connection and will most likely break.

So any connection change should always reset everything.

acarbonetto:
> The reason is that any graph currently on the canvas has the potential to be expanded or the data to be refreshed. But if the connection changed, then the data might not exist in the new connection and will most likely break.

I think this means the connection is already active?  I'm not really sure why we would need to 'activate it'?  I understand the 'match' strategy, but I think we should have a dialog with the following options: 
1. [Cancel] - change to [Ignore]? 
2. [Add Connection] - this adds the connection, but doesn't reset the graph session or activate it - this option is not available if the connection already exists (is a match)
3. [Connect] - this adds the connection AND resets the graph session to activate it

kmcginnes:
Good questions — I reworked this quite a bit; it's now #1828, which replaces this PR.

A connection link opens in a **new tab/window** — there's no way to push params into an already-loaded Graph Explorer tab, so every link starts a fresh session. That removes the "switching could clear a session you care about" concern: there's nothing on the canvas to lose.

That's what simplifies the dialog question:

```
resolve(link):
  matches the ACTIVE connection   -> do nothing, clean up the URL
  matches an INACTIVE connection  -> switch to it (no prompt)
  matches nothing                 -> open the create form, prefilled but editable
  invalid params                  -> toast + ignore
```

@acarbonetto — I dropped the 3-option dialog. Switching to a connection you already have needs no prompt (same as clicking it in the connections list), since the session is fresh anyway. The only friction I kept is the create form for an *unknown* endpoint — the one case where a link introduces something unvetted. No "add but don't activate" mode; a link's whole intent is to connect.

Matching now also considers auth posture (IAM/region/service type), so a link asking for IAM won't silently reuse a plaintext connection to the same URL.

Closing this in favor of #1828.

PR #1856, opened 2026-06-19
refactor: migrate private members to ECMAScript # private fields

### LINKED ISSUE #1845: Migrate TypeScript `private` members to ECMAScript `#` private fields
## Description

Standardize class member privacy on ECMAScript `#` private fields instead of the TypeScript `private` keyword. `#` fields are enforced at runtime by the JS engine (not erased at compile time), aren't enumerable, and can't be reached via bracket access — so the privacy guarantee is real rather than type-check-only.

`src/utils/testing/persistence.ts` already uses `#` fields. This issue brings the rest of the codebase in line with that convention.

## Scope

Member declarations currently using `private` (≈7 across 3 files):

- `packages/graph-explorer/src/utils/testing/DbState.ts` — `private _activeSchema`
- `packages/graph-explorer/src/utils/rdf/PrefixLookup.ts` — `private readonly userMap`, `private readonly inferredMap`, `private constructor`
- `packages/graph-explorer/src/connector/LoggerConnector.ts` — `private readonly _baseUrl`, `private readonly _clientLogger`, `private _sendLog`

Also drop the `_` name-prefix convention on members that become `#` (the `#` already signals privacy).

## Caveats / out of scope

- **`private constructor`** (`PrefixLookup.ts`) has no `#` equivalent — JS has no private constructors. Leave it as `private`, or address the singleton pattern separately.
- **`readonly`** can't be combined with `#` fields in TypeScript. Migrating `private readonly` → `#` loses the compile-time readonly guarantee. Decide per-field whether runtime privacy outweighs losing `readonly`, or leave those as `private readonly`.
- Constructor parameter properties (`constructor(private x …)`) have no `#` shorthand and would need a manual field + assignment.

## Acceptance criteria

- [ ] `private`/`protected` member declarations migrated to `#` where a faithful equivalent exists
- [ ] Fields where `#` would lose `readonly` or hit the constructor caveats are either left as-is or explicitly converted with the tradeoff noted
- [ ] `pnpm checks` and `pnpm test` pass

Related: #1830

<audience-footer>

> [!IMPORTANT]

> [!TIP]

</audience-footer>

vishwakt:
## Description

Standardizes class member privacy on ECMAScript `#` private fields instead of the TypeScript `private` keyword, bringing the codebase in line with the convention already used in `src/utils/testing/persistence.ts`. `#` fields are enforced at runtime by the JS engine, aren't enumerable, and can't be reached via bracket access, so the privacy guarantee is real rather than type-check-only.

## Changes

Migrated the private members in three classes and dropped the now-redundant `_` name prefixes:

- `ServerLoggerConnector` (`LoggerConnector.ts`) — `#baseUrl`, `#clientLogger`, `#sendLog()`
- `PrefixLookup` (`PrefixLookup.ts`) — `#userMap`, `#inferredMap`
- `DbState` (`DbState.ts`) — `#activeSchema`

## Caveats from the issue

- **`private constructor` in `PrefixLookup` is left as-is.** JS has no private constructors, and the static `fromArray` factory depends on it.
- **The four `private readonly` fields lose their compile-time `readonly` guarantee.** TypeScript does not allow `readonly` on `#` fields, so converting them trades compile-time immutability for runtime privacy. All four are assigned exactly once in their constructor, so the practical risk is minimal. This is the tradeoff the issue's acceptance criteria explicitly allow noting.

## Verification

- `pnpm test` passes (1809 passed, 1 expected xfail)

Fixes #1845

### DIFF
--- a/packages/graph-explorer/src/connector/LoggerConnector.ts
+++ b/packages/graph-explorer/src/connector/LoggerConnector.ts
@@ -12,42 +12,42 @@ export interface LoggerConnector {
 
 /** Sends log messages to the server in the connection configuration. */
 export class ServerLoggerConnector implements LoggerConnector {
-  private readonly _baseUrl: string;
-  private readonly _clientLogger: ClientLoggerConnector;
+  #baseUrl: string;
+  #clientLogger: ClientLoggerConnector;
 
   constructor(connectionUrl: string) {
     const url = connectionUrl.replace(/\/$/, "");
-    this._baseUrl = `${url}/logger`;
-    this._clientLogger = new ClientLoggerConnector();
+    this.#baseUrl = `${url}/logger`;
+    this.#clientLogger = new ClientLoggerConnector();
   }
 
   public error(message: unknown) {
-    this._clientLogger.error(message);
-    return this._sendLog("error", message);
+    this.#clientLogger.error(message);
+    return this.#sendLog("error", message);
   }
 
   public warn(message: unknown) {
-    this._clientLogger.warn(message);
-    return this._sendLog("warn", message);
+    this.#clientLogger.warn(message);
+    return this.#sendLog("warn", message);
   }
 
   public info(message: unknown) {
-    this._clientLogger.info(message);
-    return this._sendLog("info", message);
+    this.#clientLogger.info(message);
+    return this.#sendLog("info", message);
   }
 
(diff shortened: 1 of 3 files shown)


kmcginnes (approved):
Thanks! Looks good to me.

Issue #1820, opened 2026-06-12
Concurrent edits in multiple browser tabs silently clobber shared persisted collections (styling, schema, connections, sessions)


kmcginnes:
## Description

Graph Explorer persists shared user data (User Preferences, Schema, Connections, Sessions) to IndexedDB via `atomWithLocalForage`. IndexedDB is shared across same-origin browser tabs, but the persistence layer has **no cross-tab synchronization** (no BroadcastChannel, no `storage` events, no Web Locks — verified by grep). Each atom reads its value **once at startup** into an in-memory Jotai atom, then writes the **whole value** back on every change.

For collections written read-modify-write — e.g. `set(atom, prev => { const next = [...prev]; next.push(x); return next; })` — a second tab's in-memory `prev` is stale. Persisting its whole collection **silently drops entries another tab added**. The last writer wins for the entire collection, not just the field it touched.

**Concrete failure:** Tab A styles vertex type X → persisted. Tab B (opened earlier, never saw X) styles type Y → writes its stale styling array → **type X's styling is silently lost**. No error, no warning; the user discovers it on next reload.

Scalar atoms (e.g. active connection) are unaffected — each write is a complete intended value.

## Environment

- OS: any
- Browser: any (Chrome, Firefox, Safari)
- Graph Explorer Version: 3.1.0 (current `main`)
- Graph Database & Version: N/A — client-side persistence bug, database-independent

## Steps to Reproduce

1. Open Graph Explorer in two browser tabs (same connection, same origin).
2. In Tab A, change the color of vertex type X. (persisted to IndexedDB)
3. In Tab B (which loaded before step 2), change the color of vertex type Y.
4. Reload either tab.
5. **Observed:** Type X's styling is gone — only Tab B's change survives.
6. **Expected:** Both type X and type Y retain their styling.

## Expected Behavior

Concurrent edits to shared persisted collections from different tabs should reconcile, not clobber. A change to one entry must not destroy a sibling entry written by another tab.

## Affected collections (32 write sites across 7 files)

| Collection | Severity | Sites | Notes |
|---|---|---|---|
| **User Preferences** (`userStylingAtom`) | **EXTREME** | 6 | All in `core/StateProvider/userPreferences.ts`. Highest-frequency user mutation (every color/icon/shape/line change). |
[...]

PR #1862, opened 2026-06-23
Cross-tab reconciliation for user styling and schema (WIP)

### LINKED ISSUE #1831: Diff-merge reconciliation for User Preferences (tracer bullet)
## Parent

#1820

## What to build

Implement the storage-layer reconciliation pipeline and wire it to **User Preferences** (`userStylingAtom`) only — the highest-frequency, highest-severity collection in #1820. This is the tracer bullet that builds and proves the entire mechanism end-to-end against one atom before widening to the rest.

The pipeline, hosted behind the `atomWithLocalForage` seam:
- **Per-key write queue**: serialize writes for a given storage key and coalesce rapid pending writes, so background persistence cannot race or interleave out of order.
- **Read-before-write + diff-merge**: on persist, re-read the current persisted value, determine which entries this tab changed (resulting value vs. this tab's previous in-memory value), and apply only those changed entries onto the freshly-read value. For User Preferences the merge unit is the per-type styling entry (keyed by vertex/edge `type`).
- **Error surfacing**: persistence failures are caught and reported (not silently swallowed as today).
- **Awaitable write**: the write returns a promise so callers that need to confirm persistence can await it.

In-memory reads stay synchronous (optimistic update on write), so no consumer changes are required. The updater runs once on in-memory state; reconciliation works on resulting values, never by re-running the updater (safe for non-idempotent updaters).

Outcome: the #1820 styling clobber scenario (Tab A styles type X, Tab B with stale memory styles type Y, reload → both survive) turns green. Resolves the EXTREME-severity row.

## Acceptance criteria

- [ ] The #1820 styling clobber failing test from the test-backend slice now passes
- [ ] Concurrent edits to different vertex/edge types from two tabs both survive a reload
- [ ] Out-of-order and rapid successive writes persist correctly (serialized + coalesced)
- [ ] Persistence failures are surfaced rather than silently dropped
- [ ] The write path exposes an awaitable result for callers that need persistence confirmation
- [ ] In-memory reads remain synchronous; no consumer call sites changed for this slice
- [ ] Reconciliation is implemented per the diff-merge ADR; no consumer-facing async reads introduced

## Blocked by

- #1829
- #1830

> [!IMPORTANT]

### LINKED ISSUE #1832: Extend diff-merge reconciliation to Schema, Connections, and Sessions
## Parent

#1820

## What to build

Extend the diff-merge reconciliation pipeline (built in the User Preferences tracer bullet) to the three Map-keyed shared collections: **Schema**, **Connections**, and **Sessions**. These are `Map<connectionId, ...>` collections where the natural merge unit is the per-connection entry, so the same per-key diff-merge applies with a Map-keyed strategy.

This slice closes the core of #1820 — the CRITICAL (Schema) and HIGH (Connections, Sessions) severity rows. Concrete scenarios that must reconcile:
- Create connection X in Tab A and connection Y in Tab B → both survive (Connections).
- A Schema Sync completing in one tab must not clobber a connection's schema written by another tab (Schema).
- Per-connection session data from concurrent tabs does not clobber sibling connections (Sessions).

Also handle the two synchronous read-after-write sites in the schema-sync query paths: they read the atom back immediately after writing in the same async flow, so they must await the now-promise-returning write before reading.

Known and accepted limitation (per ADR): two tabs editing the **same** connection's entry (e.g. both appending SPARQL prefixes to the same connection) still last-writer-wins, because that data is nested below the per-connection merge granularity. Cross-connection edits are protected; same-connection edits are not.

## Acceptance criteria

- [ ] Creating different connections concurrently in two tabs preserves both after reload
- [ ] A Schema Sync completing in one tab does not erase another connection's schema written by a second tab
- [ ] Concurrent session changes for different connections do not clobber each other
- [ ] The two synchronous read-after-write sites in the schema-sync paths await the write before reading and still behave correctly
- [ ] Map-keyed collections reconcile per connection id; failing two-tab tests for Schema, Connections, and Sessions pass
- [ ] Same-connection conflict remains last-writer-wins and is documented as accepted scope

## Blocked by

- #1831

> [!IMPORTANT]

kmcginnes:
> **🚧 Work in progress — for early feedback, not final review.** The implementation approach is still in flux and **likely to change** — I'm especially looking for a gut-check on the reconciliation design. Don't review for polish yet.

## What this is

Cross-tab diff-merge reconciliation for all four shared IndexedDB-backed collections — **User Preferences** (#1831) plus **Schema, Connections, and Sessions** (#1832), under parent #1820.

Two same-origin tabs share one IndexedDB but each holds its own in-memory copy of a collection. A blind whole-value write lets a tab with a stale copy silently drop entries another tab added. This reconciles at the `flush` seam: re-read live storage → merge this tab's per-key change → write.

## The design question 👀

There are **two reconcilers**, and the difference is driven entirely by the stored *shape*:

- **`reconcileMapByKey`** (schema, connections, sessions) — the stored value is already a `Map` keyed by the merge unit (a connection id), replaced wholesale per write. So it's a ~5-line per-key map merge keyed on **reference equality** — no deep compare, one collection. All three Map-shaped atoms share it.
- **`reconcileUserStyling`** (styling) — styling is stored as **arrays** of `{ type, … }` across **two** collections (`vertices`, `edges`), so it projects each array into a map, diffs with `lodash` `isEqual`, and spreads back.

The styling machinery only exists because of that array-of-objects shape. **Open question for the team:** is the styling complexity worth it, or should `UserStyling` move to a `Map`-shaped store so it reconciles as cheaply as the rest (at the cost of a persisted-shape migration + backward-compat handling)?

## Seam

All four atoms use one new factory, `atomWithReconciledLocalForage`, which puts the merge inside the `flush` thunk handed to the write queue. The plain `atomWithLocalForage` stays a blind whole-value write — atoms opt in per the per-key diff-merge reconciliation ADR. Scalar settings (user-layout, flags) stay plain.

**Accepted scope boundary:** cross-*entry* edits are protected (two tabs editing different connections/types both survive). Two tabs editing the *same* entry remain last-writer-wins — that's nested below the per-connection merge granularity and documented in the ADR.

## How to read
1. [atomWithLocalForage.ts](https://github.com/aws/graph-explorer/pull/1862/files#diff-91cfdf9fecb4c3a6a253a8515fcf2ff526baf0221b43a7c08fb5808ed0d943fe) — start here: the `atomWithReconciledLocalForage` seam + the generic `reconcileMapByKey`
2. [userPreferences.ts](https://github.com/aws/graph-explorer/pull/1862/files#diff-33d187c2920841f962bcf3cb1c06f79d48674699c7f9c6fb70b125761c9acab3) — the more complex `reconcileUserStyling`
3. [storageAtoms.ts](https://github.com/aws/graph-explorer/pull/1862/files#diff-2e6c435d81a94fe56fa689cd77b26189f05c931ae3c10ec77bc8a3900c1c1442) — which atoms opt in (styling, schema, configuration, graph-sessions)
4. [persistence.test.ts](https://github.com/aws/graph-explorer/pull/1862/files#diff-51157de97df3a9da81e5a22e834c77f6e6abc0153530005b4d3bd60a4404c6e8) — cross-tab regression tests for every reconciled collection, synchronized on `waitForIdle()`

## Status
All checks and the full suite (1882 tests) pass. The #1820 styling-clobber regression marker is flipped green.

- Part of #1820
- Closes #1831
- Closes #1832

### DIFF
--- a/.kiro/skills/testing/SKILL.md
+++ b/.kiro/skills/testing/SKILL.md
@@ -501,10 +501,10 @@ test("should handle errors gracefully", async () => {
 
 Because localForage runs against a real (fake-indexeddb) backend, tests can exercise the actual persistence path. `@/utils/testing` provides helpers for this, especially for cross-tab scenarios (two same-origin tabs share one IndexedDB but each holds its own in-memory cache — see the per-key diff-merge reconciliation ADR):
 
-- `openPersistenceTab(key, initialValue)` — simulates one browser tab: its own Jotai store and its own `atomWithLocalForage` instance, sharing the per-test IndexedDB. Open two tabs on the same key to express "Tab A persisted X; Tab B holds stale memory" situations.
+- `openPersistenceTab(key, initialValue, reconcile?)` — simulates one browser tab: its own Jotai store and its own `atomWithLocalForage` instance, sharing the per-test IndexedDB. Open two tabs on the same key to express "Tab A persisted X; Tab B holds stale memory" situations. Pass the `reconcile` the real atom is wired with (e.g. `reconcileUserStyling`) to exercise cross-tab reconciliation; omit it to model a plain whole-value-overwrite atom.
 - `tab.read()` — the tab's in-memory value.
-- `tab.write(value)` — writes through the tab; returns a promise that resolves once the value has landed in storage (`atomWithLocalForage`'s write returns its background persistence promise, which is what makes this awaitable).
-- `tab.flush()` — awaits the tab's most recent background write, so assertions don't need arbitrary timeouts.
+- `tab.write(value)` — writes through the tab; the in-memory value updates synchronously and persistence happens in the background through the shared write queue (the write returns `void`). Await `tab.flush()` to wait for it to land.
+- `tab.flush()` — resolves once the tab's outstanding background writes have settled (via `persistenceStatusStore.waitForIdle()`), so assertions don't need arbitrary timeouts.
 - `readPersistedValue(key)` — reads what actually landed in storage, bypassing any tab's in-memory cache. Use this to assert against durable state.
 
 ```typescript
@@ -512,7 +512,8 @@ import { openPersistenceTab, readPersistedValue } from "@/utils/testing";
 
 test("a later tab preloads what an earlier tab persisted", async () => {
   const tabA = await openPersistenceTab("user-styling", {});
-  await tabA.write({ vertices: [{ type: "Person", color: "#ff0000" }] });
+  tabA.write({ vertices: [{ type: "Person", color: "#ff0000" }] });
+  await tabA.flush();
 
   const tabB = await openPersistenceTab("user-styling", {});
   expect(tabB.read()).toEqual({
--- a/docs/adr/20260616-per-key-diff-merge-cross-tab-reconciliation.md
+++ b/docs/adr/20260616-per-key-diff-merge-cross-tab-reconciliation.md
@@ -19,7 +19,7 @@ Scalar atoms (e.g. active connection) are unaffected — each write is a complet
 Reconcile at the **storage layer**, inside the `atomWithLocalForage` write path, using a **per-key, diff-the-output merge**. On each persist:
 
 1. **Re-read** the current persisted value from IndexedDB (it may reflect writes by other tabs since this tab loaded).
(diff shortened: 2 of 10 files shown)


kmcginnes:
Replaced by #1869. The branch was renamed (`user-styling-cross-tab-reconciliation` → `cross-tab-map-atom-reconciliation`) because the scope grew beyond styling to cover all five Map-keyed storage atoms. GitHub doesn't allow re-pointing a PR's head ref, so this one is closed in favor of the new PR.

PR #1824, opened 2026-06-15
Add styling import / export / reset UI

jkemmererupgrade:
## Description

**Slice 2 of the 3-slice split of #1589** per @kmcginnes's review on that PR. This is the **Styling Import / Export / Reset UI**. Stacks on slice 1 (#1777) because it relies on the `lucide:` storage convention slice 1 introduces.

Adds three buttons to **Settings → General**:

- **Export Styling** — downloads the current node and edge styling as a pretty-printed JSON file (`graph-explorer-styling.json`, 2-space indent). Icons are written as symbolic references (`"iconUrl": "lucide:plane"`) so the file is human-readable and version-controllable.
- **Import Styling** — accepts a styling JSON, validates it via Zod, applies it as the active styling, and stores it as the reference baseline. The schema accepts a shorthand `"icon": ""` field that gets converted to `"iconUrl": "lucide:"` on import.
- **Reset All Styling** — resets every node and edge style. If a styling file has been imported this session, restores to those values; otherwise reverts to hardcoded defaults.

The per-type **Reset to Default** button on `NodeStyleDialog` now also restores to the imported baseline (when one exists), via a new `defaultStylingAtom` reference copy and a shared `mergeDefaultsIntoUserStyling` helper.

The shared `toJsonFileData` utility gets an optional `indent` parameter (default: compact, to preserve existing callers' behavior). Slice 2's exporter passes `indent: 2`.

### File format notes (slice 2 ↔ #1660)

Slice 2's exported format is intentionally a subset of what `#1660`'s `styles:` section will look like — same `{ vertices, edges }` record-of-types shape, same `lucide:` icon refs. When the unified config lands, a slice-2 export can drop straight into `#1660` under `styles:`.

### Deliberately NOT in this PR

- No `/defaultStyling` static route in `app.ts` — that's slice 3, folded into the unified config (#1660) per kmcginnes's direction.
- No auto-fetch in `AppStatusLoader.tsx` — slice 3.
- No `example/defaultStyling.json` or Docker-mount docs — slice 3.

## Validation

- `pnpm run check:types` — clean
- `pnpm run check:lint` — 0 errors / 0 warnings (oxlint)
- `pnpm run check:format` — clean (oxfmt)
- `pnpm test` — all pass
- `pnpm coverage` — all four thresholds met
[...]

Issue #1660, opened 2026-04-15
Unified server configuration file


kmcginnes:
## Description

Graph Explorer's configuration system has grown organically and is now a source of friction for users and contributors. Environment variables are inconsistently named (some prefixed `GRAPH_EXP_`, some `PROXY_SERVER_`, some unprefixed). Configuration is split across `.env` files, `config.json`, `defaultConnection.json`, Docker env vars, and shell scripts (`process-environment.sh`, `docker-entrypoint.sh`, `setup-ssl.sh`). The Docker image is based on Amazon Linux with manual Node installation and self-signed cert generation. Neptune Notebook is a special-cased build arg that changes base paths, ports, and log styles. The result is a system that is hard to configure, hard to debug, and hard to extend.

## Solution

Replace the entire configuration surface with a single JSON config file (`graph-explorer.config.json`) plus a small set of consistently-named environment variables (`GE_*` prefix). The config file becomes the single source of truth for connections, styles, and app-level settings. Environment variables handle deployment-specific overrides (port, TLS, log level). The Docker image is simplified to a `node:24-alpine` base with no shell scripts. Self-signed cert auto-generation is dropped in favor of user-provided certs. The server serves the client at `/` with all API routes under `/api/`. This is a breaking change requiring a new major version.

This work is blocked by the relative routes change (#1618), which eliminates the need for proxy URLs in connections and base path baking.

## User Stories

1. As a Docker deployer, I want to volume-mount a single config file to configure Graph Explorer, so that I don't need to manage dozens of environment variables
2. As a Docker deployer, I want a small, standard Docker image (node:24-alpine), so that my deployments are lighter and more secure
3. As a developer, I want `pnpm dev` to work with zero configuration, so that I can start contributing immediately
4. As a developer, I want to drop a `graph-explorer.config.json` in the project root to configure my local connections, so that setup is simple and consistent with production
5. As an ops engineer, I want environment variables to override config file values for port, TLS, and log settings, so that I can customize per-environment without editing files
6. As a Neptune Notebook deployer, I want to use the same Docker image with different config values, so that I don't need a special build
7. As a team lead, I want to define multiple named connections in the config file that are locked/managed, so that my team gets a consistent set of connections on every startup
8. As a team lead, I want to define default styles for nodes and edges in the config file, so that my team sees consistent visual defaults
9. As a user, I want to create my own connections through the UI alongside managed connections, so that I can explore additional databases
[...]

Issue #689, opened 2024-11-26
[Roadmap] Persisted and Shared User Settings


kmcginnes:
We envision the ability to share defined user settings with other users.

## Potential Scope

- Share node and edge styling (shape, color, icon, etc) settings
- Share SPARQL namespace settings
- Share connection and schema information

## Related Issues

- #573
- #325
- #173
- #112

## Disclaimer

This is a high level roadmap item. The details of the implementation and the described scope could change as the project evolves.

PR #1879, opened 2026-07-01
Add styling import/export with a shared file envelope

### LINKED ISSUE #573: Customization import / export or other customization sharing method
## Description

When working with a team, it's currently very tedious to replicate the visualisation customisations e.g. the node colours / shapes, the way edge names are displayed, the custom icons for node-type, etc.

It involves sharing icon files, sharing hex values for colours, manually replicating typography and case etc.

## Preferred Solution

It would be great if there was either an export / import button which allowed the export and import of all or partial customisation settings, or a save / commit / upload on the part of the user doing the customisation which would make that customisation available to anyone else with access to that Neptune instance / workbook.

## Additional Context

In schema-first graph databases (like TigerGraph), this is done at the schema level which is super useful for sharing customisations and enforcing visualisation style, but obviously doesn't allow the flexibility of letting each user set their own styles or customisations.

It would be great to have the richness of sharing styles / customisations, but in the flexible schema world of Neptune.

## Related Issues

- #112 
- #298 
- #111

kmcginnes:
## Description

Adds **styling load/save** to Settings → Styles: users can save their current styling configuration to a file and load one on another machine or browser to reproduce their visual setup.

The feature introduces a **three-layer styling cascade** — Custom → Shared → App Defaults — where loading writes only the *shared styles* layer, non-destructively to the user's own customizations. Along the way it extracts a reusable **shared file envelope** (`{ meta, data }` with kind + version guarding) that the graph file feature now also consumes.

Key design decisions:

- **Atomic load.** A file loads in full or not at all. Any invalid value rejects the whole file and reports every offending location; nothing is half-applied. A valid file that carries no recognized styles reports "No Styles Found" rather than a false success.
- **Forward compatible.** New entries and new optional fields from a future build load cleanly (unknown fields are stripped, not rejected); the format version is a single integer generation that bumps only on a breaking change.
- **Secure against untrusted files.** Loaded icons are allowlisted to Lucide references or base64 data URIs — remote URLs are rejected so loading never triggers an outbound request — and unknown fields are stripped so a file can't smuggle a raw icon URL past the allowlist.
- **Legible failures.** Load and save failures route through the shared `createDisplayError` helper, so a wrong-kind or too-new file reports a specific reason ("Invalid file", with the mismatch) rather than a generic error; a dismissed save picker is treated as a cancel, not a failure.

## Validation

- `pnpm test` passes — including a golden-file corpus proving every shipped file generation (integer and legacy `"1.0"` string) still loads, round-trip save→load tests, atomic-rejection cases, and component tests for the load dialog's every phase.

## How to read

1. [docs/features/settings.md](https://github.com/aws/graph-explorer/pull/1879/files#diff-befcb2c0fdbd93f1754bfdf784949e590caef624bd62d1b52ca2f4d065632a89) — start here; the user-facing behavior in prose (cascade, atomic load, reset actions).
2. [core/fileEnvelope/fileEnvelope.ts](https://github.com/aws/graph-explorer/pull/1879/files#diff-c63180029be679c2078eaad0bf39bc05b116077bb40edd04d565363a8068abff) — the shared envelope: version schema, read/write meta contract split, kind + generation guard.
3. [core/styling/stylingParser.ts](https://github.com/aws/graph-explorer/pull/1879/files#diff-1c2cdafafef1cae145218e5399fe8e8c881a6474c11ce28ee2fa6c859f350d00) — the styling payload contract: whole-file validation, icon allowlist, discriminated load issues.
4. [core/styling/useStylingImportExport.ts](https://github.com/aws/graph-explorer/pull/1879/files#diff-f3dade2b1da57c373086466cb1ad8c8f71db382c769f794c8dafff1d595fe63d) — the save merge and non-destructive load into the shared-styles layer.
5. [routes/Settings/LoadStylesButton.tsx](https://github.com/aws/graph-explorer/pull/1879/files#diff-03253f4f3da92815aaaebb8befc0c898cd428db0b36fe539949697d45a9b78c9) — the load dialog phase machine (conflicts / invalid / empty / complete / failed).
6. [core/StateProvider/userPreferences.ts](https://github.com/aws/graph-explorer/pull/1879/files#diff-33d187c2920841f962bcf3cb1c06f79d48674699c7f9c6fb70b125761c9acab3) — supporting change: the layered storage model the cascade reads from.

The `docs/adr/` entries capture the file-format and envelope decisions in depth if you want the rationale.

### Screenshots

**Settings → Styles.** The new page: save/load for style sharing, plus a danger zone for the two independent resets. The indicator shows how many types currently carry shared styles.

Styles settings page

#### Loading

| Conflict prompt | Success |
|---|---|
| Loading over existing shared styles lists the types that will be replaced and warns it can't be undone. New types are added alongside. | On success, the dialog reports exactly what was applied. |
| Replace existing shared styles confirmation | Styles loaded with counts |

**Atomic rejection.** A file with any invalid value loads nothing and lists every offending type and field so it can be fixed and reloaded.

Load failed, invalid values listed

#### Resetting

The two layers reset independently; each confirms before the irreversible change.

| Reset custom styles | Reset shared styles |
|---|---|
| Reset custom styles confirmation | Reset shared styles confirmation |

## Related Issues

- Closes #573
- Related to #689
- Related to #1589
- Related to #1660
- Builds on #1777
- Supersedes #1824
- Follow-up #1880 (adopt the new `typedEntries` helper across existing `Object.entries()` call sites)

### Check List


packages/graph-explorer/src/modules/GraphViewer/ImportGraphButton.tsx
```diff
@@ -25,7 +25,7 @@ export function ImportGraphButton() {
   return (
     <FileButton
       onChange={payload => payload && importGraph.mutate(payload)}
-      accept="application/json"
+      accept=".graph.json"
```
noel-improv: **[non-blocking]**:`.graph.json` filters the picker by filename suffix, so a renamed or older export greys out in the OS dialog. Maybe keep it `application/json` and let the save side own the naming? Similar to styles (`LoadStylesButton.tsx:165` → `.styles.json`) while connections deviated unless this was intentional

packages/graph-explorer/src/modules/GraphViewer/renderNode.tsx
```diff
@@ -33,7 +34,10 @@ const iconQueryOptions = (url: string) =>
       }
       logger.debug("Fetching icon", url);
       const response = await fetch(url);
-      return await response.text();
+      const text = await response.text();
```
noel-improv, line 37: **[non-blocking]**: Only touches custom SVG icons, lucide/default and raster paths return before this. Same legacy tension as the icon allowlist: an old custom icon leaning on `<foreignObject>`/`<style>`/external `<image href>` would render blank now (also `VertexIcon.tsx:53`). Might be worth testing against a real one or safe by assumption?

packages/graph-explorer/src/modules/GraphViewer/exportedGraph.ts
```diff
@@ -13,49 +15,49 @@ import {
...
+export const GRAPH_EXPORT_KIND = "graph-export";
+
+/**
+ * Format generation. A single integer that bumps only on a breaking change
+ * (renamed or removed fields). Additive changes are made as optional fields and
+ * do not bump it. This is both the version written to new files and the newest
+ * generation this build can read; files on disk from before the integer switch
+ * carry the legacy `"1.0"` string, which the envelope normalizes to `1`.
+ */
+export const GRAPH_EXPORT_VERSION = 1;
```
noel-improv, line 36: **[potentially-blocking]**: The doc above covers the read direction, a new build normalizes legacy `"1.0"` → `1`. The gap is the reverse: this writes integer `1`, and an older build's `version: z.literal("1.0")` has no matching normalize, so it rejects files this build exports (new→old fails). Is emitting `"1.0"` on write still on the table or is dropping old-build read-compat intended?

packages/graph-explorer/src/core/styling/stylingParser.ts
```diff
@@ -0,0 +1,282 @@
...
+// --- Per-entry zod schemas ---
+
+/**
+ * Imported icons must be either a Lucide reference or a base64-encoded data URI.
+ * Remote URLs are rejected so importing never triggers outbound requests.
+ */
+const safeIconValue = z
+  .string()
+  .regex(
+    /^(lucide:[a-z0-9-]+$|data:image\/(svg\+xml|png|jpeg|gif|webp);base64,)/,
```
noel-improv: **[potentially-blocking]** Thanks for clarifying, keeping this allowlist narrow is right, the concern is that it's paired with atomic reject. A legacy user who uploaded, say, a bmp icon back when the uploader took any `image/*` now has one value outside this set and the whole styles file fails to import, all their other styles included. Could a disallowed icon drop to default and salvage the rest instead of throwing? Worth confirming behavior.

noel-improv (commented):
Overall the work is comprehensive and looks good except a few potentially blocking items and few non-blockers.

packages/graph-explorer/src/modules/GraphViewer/ImportGraphButton.tsx
kmcginnes: Agreed. The suffix `accept` is only a hint browsers can ignore, so it filtered nothing while greying out renamed or older exports. Switched both the graph and styles loaders to `application/json` (matching connections). d800fcda

packages/graph-explorer/src/modules/GraphViewer/renderNode.tsx
kmcginnes: Added tests instead of leaving it to assumption (27f071a0). The behavior is gentler than a blank render. The sanitizer strips only the executable constructs (`<script>`, event-handler attributes, and `<foreignObject>`) and leaves the rest of the drawing intact, including inline `<style>` and `<image>` elements. So a legacy custom icon loses only the stripped pieces rather than going blank.

packages/graph-explorer/src/modules/GraphViewer/exportedGraph.ts
kmcginnes: Good catch!

Kept old-build read-compat. Graph export still writes `"1.0"` to disk, so older builds that validate `version` as the literal `"1.0"` can still import our files. On read, the envelope normalizes both `"1.0"` and `1` to the integer generation. Styling export is a new format with no older readers, so it writes the integer directly. d3daa018

packages/graph-explorer/src/core/styling/stylingParser.ts
kmcginnes: Confirmed and fixed the cause. The allowlist was narrower than the uploader: the uploader accepts any `image/*` and stores whatever subtype the browser reports, but the allowlist only permitted svg/png/jpeg/gif/webp, so a legacy bmp or ico would export fine and then fail to import. The allowlist now accepts any `image/*` base64 data URI (still rejecting remote URLs and non-image schemes), so real legacy icons pass. I also extracted `isAllowedIconValue` as one gate and applied it at the upload seam, so what you can upload and what import accepts can't drift again. 322c0da5

On atomic vs salvage: kept it atomic on purpose. A styling file is one coherent snapshot, so a clear all-or-nothing failure that reports every offending value beats silently half-applying it. Widening the allowlist removes the realistic case that would have tripped it. Documented the rationale in the ADR. 7e73065c

Issue #1876, opened 2026-06-30
Spike: deep-module restructuring of the storage-atom persistence layer


kmcginnes:
## Goal

The persistence stack under `src/core/StateProvider/` has a clean base (`writeQueue` → `createWriteThroughAtom`) but splays at the top into shallow, divergent surface. Investigate how to restructure it into **deep modules** — small interfaces, substantial hidden implementation — and produce a recommendation + sliced plan.

Three coupled pain points motivate this:

1. **Two divergent factories for one idea.** `atomWithLocalForage(key, default, reconcile?)` and `createSessionScopedAtom({key, default, codec})` both seed-before-return, write through the queue, and wrap `createWriteThroughAtom`. They differ only in *storage policy* (where the value lives, the cross-tab strategy, serialization). Callers must know which factory to pick, and the two have different parameter shapes for the same job.
2. **`storageAtoms.ts` is a god-list with an ordering footgun.** A positional `Promise.all` whose array order must match the destructure order (the exact "circular-dep regression" its own DEV NOTE warns about). Adding a key means editing three places, and per-tab vs. shared-vs-reconciled scope is invisible at the call site.
3. **Per-concept knowledge is scattered.** A single concept (e.g. graph-view-layout) has its key+default+codec in one file, wiring in `storageAtoms.ts`, migration in `migrateUserLayout.ts`, and hooks in another — no single definition of "the persisted entry."

Underlying all three: **cross-tab semantics are three unnamed mechanisms** (merge / isolate / overwrite), chosen implicitly by factory + optional arg. A reader can't answer "what happens when two tabs edit this?" without tracing code.

**Dependency note:** everything here is in-process or local-substitutable (fake-indexeddb, in-memory sessionStorage — both already injectable), so each option is deepenable by merging and testing *through the interface*; no network ports needed. The one genuine seam is the **storage backend** (localForage / sessionStorage / in-memory), which already has 2+ adapters.

### Option 1 — One `persistentAtom`, storage policy as a port

Collapse the two factories into one interface; make the *policy* a swappable strategy (a genuine port — three real adapters).

```ts
type StoragePolicy<T> = {
  seed(key: string, fallback: T): Promise<T>;   // how to read the initial value
  persist(key: string, value: T): void;         // how a change is written
};

function persistentAtom<T>(key: string, defaultValue: T, policy: StoragePolicy<T>)

// named adapters — the cross-tab semantics, finally named:
sharedValue<T>()                  // blind overwrite (scalars)
[...]

PR #1894, opened 2026-07-07
Scope graph-view and schema-view layouts to the browser tab

kmcginnes:
## Description

The graph-view and schema-view layout state (active sidebar tab, sidebar width, view toggles, details-auto-open preference) was persisted as a single localForage value shared across all browser tabs. Opening the app in a second tab clobbered the first tab's layout — last-writer-wins over the whole value.

This scopes both layouts to the browser tab using the same per-tab mechanism the active connection already uses: the live value lives in `sessionStorage` (survives this tab's reload, never leaks to other tabs), with a shared localForage breadcrumb that seeds a freshly-opened tab from the last-used layout on cold start.

**Core change:** the per-tab + breadcrumb "trick" that was inlined in `activeConnectionStorage.ts` is extracted into a generic `createSessionScopedAtom<T>`. `createActiveConfigurationAtom` is refactored onto it (so the subtle seed-and-claim logic lives in one place), and both layout atoms now route through it with zod-validated codecs.

Notable details:
- **zod-validated per-tab reads.** `sessionStorage` holds strings, so each layout has a codec. `deserialize` returns `null` for an absent value but *throws* on a corrupt/wrong-shape one; the seam (`createSessionScopedAtom`) catches, logs, and falls back to the breadcrumb then the default — so a stale or hand-edited value can never crash the top-level-await startup. Detecting corruption is kept separate from deciding what to do about it.
- **`Set` serialization.** Graph-view's `activeToggles` is a runtime `Set`; its codec serializes it as an array for JSON and rebuilds the `Set` on read. Schema-view is plain JSON.
- **No migration.** The breadcrumb keeps each concept's existing localForage key and shape, so existing stored layouts seed a fresh tab unchanged.

The new per-tab scope is recorded in an ADR alongside the two existing shared scopes (reconciled maps, blind-write scalars).

## Validation

- `pnpm test` passes (full suite, 2086 tests), including new coverage in `sessionScopedStorage.test.ts` (seed order, corrupt-value + `SecurityError` recovery, the `Set`-bearing cold-start claim, and cross-tab isolation/seeding through the real graph-view and schema-view codecs).
- Reviewer focus: confirm the seeding order (this tab's sessionStorage → shared breadcrumb → default), that a corrupt per-tab value recovers instead of crashing, and that graph-view's `activeToggles` `Set` round-trips across a write-then-cold-start sequence.

## How to read

1. [packages/graph-explorer/src/core/StateProvider/sessionScopedStorage.ts](https://github.com/aws/graph-explorer/pull/1894/files#diff-0ff70707c3bc1942183accc82210592683e754ed2599ee8bc5df0fc494f255d4) — start here; the extracted per-tab primitive (seed → claim → write-through) plus `parseSessionJson` and the corrupt-value recovery seam
2. [packages/graph-explorer/src/core/StateProvider/activeConnectionStorage.ts](https://github.com/aws/graph-explorer/pull/1894/files#diff-bff12db65182d96dffd13459187b5ca6d92337fb6d2ce46a4558ac816903e7d1) — the previously-inlined trick, now a thin wrapper over the primitive (behavior-preserving refactor)
3. [packages/graph-explorer/src/core/StateProvider/graphViewLayoutDefaults.ts](https://github.com/aws/graph-explorer/pull/1894/files#diff-87bd3cad4efcb73f39d8e41ab3eab9aff31bb45ee11223db234b754dde75c84f) — the graph-view codec; the `activeToggles` `Set` ↔ array serialization
4. [packages/graph-explorer/src/core/StateProvider/schemaViewLayoutDefaults.ts](https://github.com/aws/graph-explorer/pull/1894/files#diff-fbc4772bdbb4fe7324bc51ff976c8d5cb696c7a55aa7ac2ebe96e34fae5511a9) — the schema-view codec (plain JSON)
5. [packages/graph-explorer/src/core/StateProvider/storageAtoms.ts](https://github.com/aws/graph-explorer/pull/1894/files#diff-2e6c435d81a94fe56fa689cd77b26189f05c931ae3c10ec77bc8a3900c1c1442) — wires both layout atoms onto the primitive
6. [docs/adr/20260630-per-tab-session-scoped-storage-primitive.md](https://github.com/aws/graph-explorer/pull/1894/files#diff-cfa0454aee6e3505762ca9fed24e6d9ee37bf8b2a38e2d62566a1597f7d0613a) — the storage-scope decision and the three named scopes

## Related Issues

- Related to #1876 (spike to formalize the storage-scope primitives)

### Check List


packages/graph-explorer/src/core/StateProvider/sessionScopedStorage.ts
```diff
@@ -0,0 +1,173 @@
...
+}) {
+  let seedValue = readSessionSeed(sessionStorage, key, codec);
+  if (seedValue === null) {
+    // Cold start: seed from the shared breadcrumb and claim it into this tab's
+    // sessionStorage, so a later reload reads this value back rather than
+    // re-seeding from a breadcrumb another tab may have since moved.
+    const breadcrumb = await localForage.getItem<T>(key);
+    if (breadcrumb !== null) {
+      seedValue = transform ? transform(breadcrumb) : breadcrumb;
+      writeSession(sessionStorage, key, codec, seedValue);
```
mjuarros: The per-tab leg is hardened and the breadcrumb leg is trusted, so the self-heal guarantee holds on one side only.

`readSessionSeed` wraps `codec.deserialize` in try/catch, logs, and returns `null` so a bad per-tab value falls through — that is the documented behavior and the corrupt-value and `SecurityError` tests pin it. This branch has no equivalent. `localForage.getItem<T>(key)` asserts the breadcrumb as `T` without parsing it, then `transform` and `writeSession` both run outside any try.

That matters because of the interaction with two other lines. `writeSession` deliberately places `const serialized = codec.serialize(value)` outside its own try at line 160 (correct, for a genuine defect), and `graphViewLayoutCodec.serialize` spreads `[...layout.activeToggles]` at `graphViewLayoutDefaults.ts:100`. `transformGraphViewLayout` only touches `activeSidebarItem`, so a breadcrumb whose `activeToggles` is absent or non-iterable reaches that spread untouched, throws `TypeError`, and rejects this `createSessionScopedAtom` call — which `storageAtoms.ts` awaits inside a top-level `Promise.all` with no surrounding try. That is a blank app rather than a degraded one.

The ADR states the self-heal narrowly and correctly, scoping it to "a corrupt or stale **per-tab** value." The PR description generalizes it to "a stale or hand-edited value can never crash the top-level-await startup," and that word is what I am questioning.

Likelihood is bounded, which is why I read this as important rather than blocking: `activeToggles` has been a `Set` in the persisted shape since before the `user-layout` to `graph-view-layout` rename, and `migrateUserLayout` copies the legacy value verbatim, so a well-formed legacy breadcrumb round-trips fine. The reachable triggers are a hand-edited IndexedDB value — the same input class the per-tab leg was hardened against — and a restored config backup or rolled-back app version writing a different native shape, which is precisely the path "no migration" exists to protect.

Two ways out, and I don't have a strong preference:

1. Parse the breadcrumb through the same schema before claiming it, so both legs share one corruption seam and a throw falls through to `defaultValue`. Costs one parse per cold start plus a corrupt-breadcrumb test, and makes the general claim true.
2. Narrow the PR description to match what the ADR already says. Costs nothing, but leaves the hand-edit and rollback vectors able to take down boot.

Either is defensible for a non-critical view preference. Worth noting there is currently no test for a corrupt breadcrumb, only for a corrupt per-tab value.

CONTEXT.md
```diff
@@ -149,6 +169,9 @@ _Avoid_: Save-status indicator
 - **Neighbors** are **Vertices** one hop away from a given **Vertex**
 - **Styles** are scoped per **Vertex Type** (**Vertex Styles**) and **Edge Type** (**Edge Styles**)
 - The **Graph View**, **Data Table View**, and **Schema View** all render from the same **Session** and **Schema**
+- Each browser tab has its own **View Layout** per view, the same divergence as **Active Connection**
+- A **Layout** positions **Vertices** on the **Graph View** canvas and is not part of any **View Layout**
```
mjuarros: A question about how absolute you want this, because my in-flight #2144 collides with it and I am not assuming the glossary is the side that should bend.

This invariant, plus the `Layout` entry's `_Avoid_: View Layout (a different concept)` and the closed enumeration in `Schema View Layout` ("active sidebar panel, sidebar width, and the details-auto-open preference"), together say a Layout is never part of a View Layout. #2144 adds a required `layoutAlgorithm: LayoutName` field directly to `SchemaViewLayout` and defaults it in `transformSchemaViewLayout`. The moment it lands, a Layout is part of a View Layout and that enumeration is incomplete. Since this PR is sequenced first, the text would be merged as settled fact and then falsified.

I can see two resolutions and I think the choice is yours as the owner of the domain model:

1. **The invariant is what you meant, and #2144 is wrong.** Then `layoutAlgorithm` does not belong in `SchemaViewLayout` at all and I move it to its own atom in #2144. Costs me an atom and keeps the concept boundary sharp, which is a real benefit: it would mean "View Layout" stays purely UI chrome.
2. **The invariant is over-stated.** Then scoping it to the Graph View ("it is the algorithm, not part of the Graph View Layout's UI state") and opening the `Schema View Layout` list with "such as" costs one line here and saves a doc-correcting commit in #2144.

For what it is worth, #2145 does not conflict — its `Graph Arrangement` entry also treats Layout as the algorithm only, carrying `_Avoid_: Layout (only the algorithm)`. It does add a sixth Layout-adjacent term to this region of the glossary, so whichever way this goes, the six read better if they agree on where the algorithm lives.

Happy to take the work in #2144 either way. I just want the decision made here rather than discovered later.

CONTEXT.md
```diff
@@ -85,6 +85,26 @@ _Avoid_: Data Explorer (legacy route name)
...
+The View Layout for the Graph View — active sidebar panel, sidebar width, active content toggles, table-view height, and the details-auto-open preference.
+_Avoid_: Graph preferences, graph settings
+
+**Schema View Layout**:
+The View Layout for the Schema View — active sidebar panel, sidebar width, and the details-auto-open preference.
+_Avoid_: Schema preferences, schema settings
+
+**Storage Scope**:
+The cross-tab behavior a persisted atom picks at creation, so scope is a visible decision rather than a side effect of which factory was reached for. Three named scopes: **per-tab**, where tabs diverge and a fresh tab starts from the value most recently used; **shared-reconciled**, where a Map-keyed collection is merged per key across tabs; and **shared-blind-write**, where each write is the whole value. See the `per-tab-session-scoped-storage-primitive` ADR for which atoms use which, and `per-key-diff-merge-cross-tab-reconciliation` for the merge rule.
+_Avoid_: Persistence mode, storage strategy
```
mjuarros: Minor, and the term itself clearly earns its place — it names a decision the code really does make at atom creation.

The length is what I'd push back on. `CONTEXT-FORMAT.md` asks for "one or two sentences max" and "define what it IS, not what it does," and `docs/agents/domain.md` splits *what* into the glossary and *why* into ADRs. This entry enumerates all three scopes with their merge and divergence mechanics and points at two ADRs, which reads closer to an ADR abstract than a definition. The neighbouring `Layout` and `View Layout` entries are tight by comparison.

Something like: "The cross-tab persistence behavior a persisted atom picks at creation — per-tab (tabs diverge), shared-reconciled (merged per key), or shared-blind-write. See the storage ADRs for which atoms use which." The per-scope mechanics are already in the new ADR, so restating them here is the part that could go.

Entirely your call on which detail an agent needs inline versus behind a link — that is the judgement the glossary exists to encode, so I'm not going to be confident about it from outside.

docs/adr/20260921-per-tab-session-scoped-storage-primitive.md
```diff
@@ -0,0 +1,42 @@
...
+
+- **Per-tab** — `createSessionScopedAtom`. Live value in `sessionStorage`; shared localForage breadcrumb read once as the cold-start seed and claimed into `sessionStorage`. Tabs diverge. Backs Active Connection, graph-view layout, schema-view layout.
+- **Shared-reconciled** — `atomWithLocalForage` with `reconcileMapByKey`. Map-keyed collections genuinely shared across tabs, merged per key (`per-key-diff-merge`). Backs Connections, Schema, Vertex and Edge Styles, Sessions.
+- **Shared-blind-write** — `atomWithLocalForage` with no reconciler. Scalars where each write is the whole intended value and tabs need not diverge. Backs the boolean/number settings (e.g. `showDebugActions`).
+
+`createActiveConfigurationAtom` is refactored onto the primitive rather than left as a parallel implementation, so the seed-and-claim logic lives in exactly one place. Its own tests stay, narrowed to what the wrapper still owns: the empty-string-is-a-miss codec rule, the bare-string round trip, and the `resolveSessionStorage` fallback. They also stand as behavior-preservation evidence for the refactor.
```
mjuarros: Small factual correction: nothing was narrowed. `git diff bfc06442 HEAD -- packages/graph-explorer/src/core/StateProvider/activeConnectionStorage.test.ts` is empty — all ten tests are unchanged, including the multi-tab cases.

Worth fixing because the reality is stronger than the claim. Retaining the full pre-existing suite unmodified is better behavior-preservation evidence for the refactor than retaining a narrowed subset would be, and it is the thing that made me comfortable with the `createActiveConfigurationAtom` rewrite. Suggest rewording to say the tests were kept intact as behavior-preservation evidence.

packages/graph-explorer/src/core/StateProvider/sessionScopedStorage.ts
mjuarros, line 95: The per-tab leg is hardened and the breadcrumb leg is trusted, so the self-heal guarantee holds on one side only.

`readSessionSeed` wraps `codec.deserialize` in try/catch, logs, and returns `null` so a bad per-tab value falls through — that is the documented behavior and the corrupt-value and `SecurityError` tests pin it. This branch has no equivalent. `localForage.getItem<T>(key)` asserts the breadcrumb as `T` without parsing it, then `transform` and `writeSession` both run outside any try.

That matters because of the interaction with two other lines. `writeSession` deliberately places `const serialized = codec.serialize(value)` outside its own try at line 160 (correct, for a genuine defect), and `graphViewLayoutCodec.serialize` spreads `[...layout.activeToggles]` at `graphViewLayoutDefaults.ts:100`. `transformGraphViewLayout` only touches `activeSidebarItem`, so a breadcrumb whose `activeToggles` is absent or non-iterable reaches that spread untouched, throws `TypeError`, and rejects this `createSessionScopedAtom` call — which `storageAtoms.ts` awaits inside a top-level `Promise.all` with no surrounding try. That is a blank app rather than a degraded one.

The ADR states the self-heal narrowly and correctly, scoping it to "a corrupt or stale **per-tab** value." The PR description generalizes it to "a stale or hand-edited value can never crash the top-level-await startup," and that word is what I am questioning.

Likelihood is bounded, which is why I read this as important rather than blocking: `activeToggles` has been a `Set` in the persisted shape since before the `user-layout` to `graph-view-layout` rename, and `migrateUserLayout` copies the legacy value verbatim, so a well-formed legacy breadcrumb round-trips fine. The reachable triggers are a hand-edited IndexedDB value — the same input class the per-tab leg was hardened against — and a restored config backup or rolled-back app version writing a different native shape, which is precisely the path "no migration" exists to protect.

Two ways out, and I don't have a strong preference:

1. Parse the breadcrumb through the same schema before claiming it, so both legs share one corruption seam and a throw falls through to `defaultValue`. Costs one parse per cold start plus a corrupt-breadcrumb test, and makes the general claim true.
2. Narrow the PR description to match what the ADR already says. Costs nothing, but leaves the hand-edit and rollback vectors able to take down boot.

Either is defensible for a non-critical view preference. Worth noting there is currently no test for a corrupt breadcrumb, only for a corrupt per-tab value.

CONTEXT.md
mjuarros, line 173: A question about how absolute you want this, because my in-flight #2144 collides with it and I am not assuming the glossary is the side that should bend.

This invariant, plus the `Layout` entry's `_Avoid_: View Layout (a different concept)` and the closed enumeration in `Schema View Layout` ("active sidebar panel, sidebar width, and the details-auto-open preference"), together say a Layout is never part of a View Layout. #2144 adds a required `layoutAlgorithm: LayoutName` field directly to `SchemaViewLayout` and defaults it in `transformSchemaViewLayout`. The moment it lands, a Layout is part of a View Layout and that enumeration is incomplete. Since this PR is sequenced first, the text would be merged as settled fact and then falsified.

I can see two resolutions and I think the choice is yours as the owner of the domain model:

1. **The invariant is what you meant, and #2144 is wrong.** Then `layoutAlgorithm` does not belong in `SchemaViewLayout` at all and I move it to its own atom in #2144. Costs me an atom and keeps the concept boundary sharp, which is a real benefit: it would mean "View Layout" stays purely UI chrome.
2. **The invariant is over-stated.** Then scoping it to the Graph View ("it is the algorithm, not part of the Graph View Layout's UI state") and opening the `Schema View Layout` list with "such as" costs one line here and saves a doc-correcting commit in #2144.

For what it is worth, #2145 does not conflict — its `Graph Arrangement` entry also treats Layout as the algorithm only, carrying `_Avoid_: Layout (only the algorithm)`. It does add a sixth Layout-adjacent term to this region of the glossary, so whichever way this goes, the six read better if they agree on where the algorithm lives.

Happy to take the work in #2144 either way. I just want the decision made here rather than discovered later.

CONTEXT.md
mjuarros, line 106: Minor, and the term itself clearly earns its place — it names a decision the code really does make at atom creation.

The length is what I'd push back on. `CONTEXT-FORMAT.md` asks for "one or two sentences max" and "define what it IS, not what it does," and `docs/agents/domain.md` splits *what* into the glossary and *why* into ADRs. This entry enumerates all three scopes with their merge and divergence mechanics and points at two ADRs, which reads closer to an ADR abstract than a definition. The neighbouring `Layout` and `View Layout` entries are tight by comparison.

Something like: "The cross-tab persistence behavior a persisted atom picks at creation — per-tab (tabs diverge), shared-reconciled (merged per key), or shared-blind-write. See the storage ADRs for which atoms use which." The per-scope mechanics are already in the new ADR, so restating them here is the part that could go.

Entirely your call on which detail an agent needs inline versus behind a link — that is the judgement the glossary exists to encode, so I'm not going to be confident about it from outside.

docs/adr/20260921-per-tab-session-scoped-storage-primitive.md
mjuarros, line 23: Small factual correction: nothing was narrowed. `git diff bfc06442 HEAD -- packages/graph-explorer/src/core/StateProvider/activeConnectionStorage.test.ts` is empty — all ten tests are unchanged, including the multi-tab cases.

Worth fixing because the reality is stronger than the claim. Retaining the full pre-existing suite unmodified is better behavior-preservation evidence for the refactor than retaining a narrowed subset would be, and it is the thing that made me comfortable with the `createActiveConfigurationAtom` rewrite. Suggest rewording to say the tests were kept intact as behavior-preservation evidence.

packages/graph-explorer/src/core/StateProvider/sessionScopedStorage.test.ts
```diff
@@ -0,0 +1,492 @@
...
+    };
+    const atom = await createSessionScopedAtom<Counter>({
+      key: KEY,
+      defaultValue: { count: 0 },
+      codec: brokenCodec,
+      sessionStorage: createInMemorySessionStorage(),
+    });
+    const store = createStore();
+
+    expect(() => store.set(atom, { count: 5 })).toThrow(TypeError);
```
mjuarros, line 226: Small one, and it is the only one of the three new type-form assertions where your justifying comment does not reach.

`docs/agents/testing.md:79` asks for the instance form (`toThrow(new FooError(a, b))`) over `toThrow(FooError)`, because the type form passes even when the code built the error with the wrong data. The two in `graphViewLayoutDefaults.test.ts` and `schemaViewLayoutDefaults.test.ts` are correctly exempt and you said why inline: `SyntaxError` messages are V8-specific and `ZodError`'s `issues` shape moved between zod 3 and 4, so pinning an instance there would couple those tests to engine and library versions.

This one is different. The `TypeError` is thrown by the test's own `brokenCodec` at line 214 with a fixed message, so there is no version coupling to avoid and the instance form is available:

```ts
expect(() => store.set(atom, { count: 5 })).toThrow(
  new TypeError("activeToggles is not iterable"),
);
```

Strictly stronger here: it would catch a regression where `writeSession` swallowed the codec's error and rethrew a different `TypeError`, which is exactly the laundering this test exists to prevent. As written, any `TypeError` from anywhere in the setter satisfies it.

Context rather than criticism: these three are the repo's first type-form assertions. The other eight `toThrow` sites all use the instance form.

Issue #1785, opened 2026-05-27
[Feature Request] Property-based node styling (dynamic icons/colors from property values)


iowusu:
## Description

There is currently no way to visually differentiate nodes based on their property values. When working with graphs that contain heterogeneous data for example, fraud detection graphs where nodes represent people, accounts, and devices all nodes of the same type look identical regardless of their property values. This makes it difficult to spot meaningful patterns at a glance.

It would be valuable to configure node icons and/or colors to change dynamically based on a node's property values (e.g., a gender property mapping to distinct icons, or a risk_score property mapping to a color gradient

## Preferred Solution

No preference — open to any approach the community sees fit.

## Related Issues

Related to [#1774](https://github.com/aws/graph-explorer/issues/1774) — adds the Lucide icon library that would supply icons for property-based mappings
Related to [#691](https://github.com/aws/graph-explorer/issues/691) — roadmap item for improved node/edge styling (shape, color, icons)
Related to [#313](https://github.com/aws/graph-explorer/issues/313) — feature request for simpler node customization options

> [!TIP]
> Please use a 👍 reaction to provide a +1/vote. This helps the community and maintainers prioritize this request.

kmcginnes:
Marking as `internal` and `needs-triage` as I will need to come up with a user experience that makes sense and some sort of rules engine that won't expose any security issues (such as if we just evaluated raw JS).

PR #1915, opened 2026-07-09
Add conditional styling for vertex and edge types

jkemmererupgrade:
## Description

Adds an optional **conditional style** per vertex and edge type. When a rendered entity's attribute satisfies a condition, an alternate style is layered on top of the base style — useful for flagging states (e.g. a `Person` node with `known_bad = true`) without defining a separate type.

- **One condition per type.** The conditional pane in each style dialog reuses the base style controls and overrides only the fields you change; type-level fields (display name/description) stay in the base pane.
- **Condition** = attribute + operator + value. Operators: `=`, `!=`, `>`, `<`, `>=`, `<=`, and **`matches`** — a `*`-wildcard text pattern (e.g. `Jo*`), where every other character is matched literally so regex metacharacters carry no special meaning. The text operators (`=`, `!=`, `matches`) have a **case-sensitivity toggle**, case-sensitive by default (conditions saved before the flag existed keep the original case-sensitive behavior). Comparison is otherwise **type-aware** and evaluated in JavaScript at render time: numeric when both sides are numbers, chronological when both parse as dates (handles ISO **and** non-ISO formats like `M/D/YYYY`), lexicographic otherwise. Matching is surfaced to Cytoscape via a `conditionMet` flag rather than Cytoscape's value-selector operators, which compare lexicographically / via `parseFloat` and mishandle dates.
- **Precedence.** Selection, focus dimming, and hidden states always take visual priority over a conditional style (Cytoscape's order-based cascade).
- **Import / export / reset.** Conditional styles ride the existing shared-file-envelope format and the styling storage, so they are saved, imported, and reset with no file-format version bump. In the selective import modal, a type that carries a condition shows a **separate, independently selectable "conditional" tile** right after its base tile (Base → When-met previews + the condition), so you can import a base style with or without its condition.

Reintegrated onto the current styling architecture after the node/edge styling sidebar was combined (#1914) and the styles cascade collapsed to a single user layer (#1981).

## Validation

- `pnpm checks` passes with no errors; `pnpm test` passes (2385 tests); `pnpm coverage` clears all thresholds without changing them.
- A full **headless-Cytoscape pipeline test** drives condition evaluation + selector generation + the applied stylesheet and asserts only the matching entity receives the conditional color — covering numeric, ISO date, non-ISO (`M/D/YYYY`) date, boolean, and wildcard (`matches`) conditions.
- Manually validated in a locally built Docker image against a live graph: styling only matching entities, correct non-ISO date comparison, wildcard matching, dialogs, and the selective import tiles.

## Related Issues

Related to #1785 (property-based node styling). This is a partial step: it covers the single-condition case — differentiating entities that meet one condition (e.g. `known_bad = true`, or `risk_score > 90` → red) — which handles the fraud use case the issue leads with. It does **not** yet cover the issue's other examples: mapping one property to several distinct icons (e.g. gender → distinct icons) or mapping a numeric property to a color gradient. Those would be follow-on work.

### Check List

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/features/graph-view.md
+++ b/docs/features/graph-view.md
@@ -83,6 +83,7 @@ On the **Nodes** tab, each node type can be customized in a variety of ways.
 - **Display description attribute** allows you to choose the attribute on the node that is used to describe the node in search
 - **Icon** can be picked from the built-in Lucide library via the **Browse** button, or uploaded as a custom SVG/raster image.
 - **Colors and borders** can be customized to visually distinguish from other node types
+- **Conditional style** applies a different appearance to individual nodes of the type when they meet a condition (see [Conditional Styling](#conditional-styling))
 
 On the **Edges** tab, each edge type can be customized in a variety of ways.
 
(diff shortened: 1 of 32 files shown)


jkemmererupgrade:
Hey @kmcginnes, Thank you for merging the node-styling! I refactored the simple conditional styling PR I had to use the latest and put it up as a direct PR and closed mine. Style import and Export as you have it works and brings along the conditional styling as well.

jkemmererupgrade:
Applying conditional styling to a node:
Screenshot 2026-07-09 at 2 29 16 PM

jkemmererupgrade:
Updated to include the conditional styling as tiles in the import panel, fixed conflicts.

kmcginnes:
Hi @jkemmererupgrade, thanks for this, and for reintegrating it onto the styling changes that landed underneath you. Being upfront about which part of [[Feature Request] Property-based node styling #1785](https://github.com/aws/graph-explorer/issues/1785) this covers and which part it doesn't is helpful.

This is a sizable addition to how styling works, so I want to think through how conditional styles fit with where we're taking the styling system before I get into the details. It's also taking me longer than I'd like to get back to community PRs, so expect this one to take some time.

PR #1930, opened 2026-07-13
Add per-type label font size to node and edge styling

danielvanza:
## Summary

- Font size was a hardcoded global constant (7px, with a 6px minimum-zoomed size) applied only through the generic `node`/`edge` Cytoscape selectors, so every vertex/edge type rendered labels at the same size.
- Adds `fontSize` and `minZoomedFontSize` to the vertex/edge style cascade (`VertexVisualStyle`/`EdgeVisualStyle`), with the existing 7/6 values as app defaults, so the fields resolve through the existing app-default → shared → user cascade.
- Emits `font-size`/`min-zoomed-font-size` from the per-type Cytoscape selectors, and exposes the two values as inputs in the node and edge style dialogs.
- The fields flow through the existing styling import/export as additive optional fields — no `STYLING_EXPORT_VERSION` bump needed.
- Also drops the node badge's fixed `maxWidth` truncation, since it assumed the old fixed font size and would clip short labels once a larger custom size is set.

## Test plan

- [x] `pnpm checks` (types, lint, format)
- [x] `pnpm test` — all green except one pre-existing, unrelated flaky test on `main` (`safeSessionStorage.test.ts`, confirmed failing on a clean checkout of `main` as well)
- [x] New/updated unit tests: style cascade defaults + overrides, per-type Cytoscape selector emission (vertex and edge), styling import/export round-trip, badge truncation behavior
- [x] Manual verification: opening a node/edge Customize dialog shows the new font-size inputs; changing them resizes that type's labels; Reset restores 7/6; export/import round-trips the values

### DIFF
--- a/docs/features/graph-view.md
+++ b/docs/features/graph-view.md
@@ -83,6 +83,7 @@ On the **Nodes** tab, each node type can be customized in a variety of ways.
 - **Display description attribute** allows you to choose the attribute on the node that is used to describe the node in search
 - **Icon** can be picked from the built-in Lucide library via the **Browse** button, or uploaded as a custom SVG/raster image.
 - **Colors and borders** can be customized to visually distinguish from other node types
+- **Label font size** and **minimum zoomed font size** control how large the node's label renders and how small it can shrink before it hides while zooming out
 
 On the **Edges** tab, each edge type can be customized in a variety of ways.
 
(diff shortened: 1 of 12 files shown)


kmcginnes:
Hi @danielvanza, thanks for the contribution, and welcome. Per-type font size fits the existing style cascade cleanly, and routing it through the app-default to user layers rather than special-casing it was the right instinct.

Fair warning that it's taking me longer than I'd like to get back to reviewing community PRs, so this may sit for a bit. I see you have a few open now, and I will work through them.

PR #1931, opened 2026-07-13
Add Schema-view vertex type hiding

danielvanza:
## Summary

- The Schema view rendered every discovered vertex type with no way to exclude noisy ones — e.g. a database's own generic base label (Gremlin's default `Vertex` label) showing up as a node type on every schema graph.
- Adds a per-connection, persisted store of hidden vertex types (`hiddenSchemaTypesAtom` + `useHiddenSchemaTypes()`), mirroring the existing schema-atom pattern, and an eye toggle on each row of the Styles panel's Nodes tab when opened from the Schema view.
- Hiding a type also drops its schema-graph edges, since they lose an endpoint — no separate edge-hiding logic needed.
- Hidden types stay listed (dimmed) so they can be restored, and the hidden set is scoped to the Schema view only: the main Graph view, canvas filtering, and discovery are untouched. The Styles panel is shared between the Graph and Schema views, so the toggle is gated behind an opt-in prop passed only from the Schema view's sidebar, and the toggle itself lives in a Schema-view-only wrapper around the shared row rather than inside it.
- Also includes a small, related fix: `createVertex()` used to take the first label in a multi-labeled vertex's type list as its primary (styling/display) type. Gremlin/TinkerPop's default base label, `vertex`, commonly appears first alongside a more specific label (e.g. `["vertex", "sqsqueue"]`), which meant such nodes were styled and named after the meaningless generic label instead of their specific one. This makes hiding the generic type from the Schema view more useful, since the same generic label no longer clobbers a node's display name/style on the main canvas.

## Test plan

- [x] `pnpm checks` (types, lint, format)
- [x] `pnpm test` — all green except one pre-existing, unrelated flaky test on `main` (`safeSessionStorage.test.ts`, confirmed failing on a clean checkout of `main` as well)
- [x] New unit tests: hidden-types atom (toggle, per-connection isolation, empty-set cleanup), Schema graph data filtering (nodes and their edges), sidebar toggle wiring, and a regression test confirming the main Graph view's Styles panel never shows the toggle
- [x] Manual verification: toggling a type off in the Schema view's Nodes tab removes it (and its connections) from the schema graph only; the row stays listed, dimmed, and restores the type when toggled back on; hidden state persists per connection across reload

### DIFF
--- a/docs/features/schema-view.md
+++ b/docs/features/schema-view.md
@@ -17,7 +17,7 @@ You can open the Schema view by clicking "Schema" in the navigation bar.
 The sidebar has two panels:
 
 - **Details** — shows properties and connections for the selected node type or edge connection
-- **Styles** — customize colors and icons for node types and edge types, split across a Nodes tab and an Edges tab
+- **Styles** — customize colors and icons for node types and edge types, split across a Nodes tab and an Edges tab. On the Nodes tab, the eye icon next to each node type hides it (and its connections) from the Schema view only — the main Graph view and data discovery are unaffected. Hidden types stay listed, dimmed, so they can be shown again; the hidden set persists per connection.
 
 Click the active tab icon to collapse the sidebar to just the icon strip. Click any tab icon to reopen it. Both the active tab and sidebar width are remembered across sessions.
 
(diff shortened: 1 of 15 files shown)


kmcginnes:
Hi @danielvanza, thanks for this one as well. Hiding a database's generic base label from the Schema view solves a real annoyance, and I like that you kept the hidden set scoped to the Schema view instead of letting it leak into the Graph view.

Same caveat as on [#1930](https://github.com/aws/graph-explorer/pull/1930): getting back around to community PRs is slow going right now, so this may sit a while.

PR #1979, opened 2026-07-20
fix: show a friendly Access denied message for 403 responses

### LINKED ISSUE #1977: 403 errors show an unfriendly "Network Response 403" message
## Description

When the database or proxy returns a 403 Forbidden, the user sees the generic `Network Response 403` title with whatever raw message the server returns, which is often an opaque JSON blob or empty. There is no dedicated branch in `createDisplayError` for 403, so it falls through to the catch-all `NetworkError` path.

A 403 can be caused by several things (credentials, permissions, proxy configuration), so the message should stay generic and point the user at what they can reasonably check.

## Environment

- All environments, browser-based

## Steps to Reproduce

1. Connect to a database or proxy that will reject requests with 403
2. Trigger any request (e.g. schema sync)

## Expected Behavior

A friendly, generic message such as:

> **Access denied**
> The server refused this request. Check that your connection credentials and permissions are correct, then try again.

## Actual Behavior

Title says "Network Response 403" and the message is either the raw server response or the generic "An error occurred. Please try again."

---

> [!IMPORTANT]

> [!TIP]

vishwakt:
Closes #1977

403 responses fell through to the catch-all `NetworkError` branch, showing "Network Response 403" with the raw server response. Added a dedicated 403 branch with the copy from the issue. The message stays generic since a 403 can mean credentials, permissions, or proxy configuration. The raw response is still available via Error Details.

Tested against a local stub returning a Neptune-style `AccessDeniedException` 403, triggered via schema sync.

**Before:**

before-403

**After:**

after-403

### DIFF
--- a/packages/graph-explorer/src/utils/createDisplayError.test.ts
+++ b/packages/graph-explorer/src/utils/createDisplayError.test.ts
@@ -178,6 +178,30 @@ describe("createDisplayError", () => {
     });
   });
 
+  it("Should handle access denied error", () => {
+    const result = createDisplayError(
+      new NetworkError("Network error", 403, null),
+    );
+    expect(result).toStrictEqual({
+      title: "Access denied",
+      message:
+        "The server refused this request. Check that your connection credentials and permissions are correct, then try again.",
+    });
+  });
+
+  it("Should handle access denied error ignoring raw server data", () => {
+    const result = createDisplayError(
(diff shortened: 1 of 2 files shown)


vishwakt:
hello @kmcginnes would you be able to spare some time to review this PR? Thank you!

Issue #2020, opened 2026-07-30
Consolidate and harden query text construction per Query Language


kmcginnes:
Query text for each supported Query Language is currently assembled by roughly twenty independent template functions, each deciding for itself how to wrap interpolated values in delimiters. A single shared helper covers only one character, and it is applied inconsistently — equivalent values are handled in one template and not in its sibling.

Improve and consolidate this logic by introducing a per-language module that owns how a value is placed into query text, so that the rules live in one place per Query Language rather than at every call site. Gremlin, openCypher, and SPARQL each have different quoting and delimiting rules, so a single shared helper cannot be correct for all three.

Expected benefits:

- Attribute names and type labels containing ordinary punctuation (spaces, quotes, backticks, dots) no longer produce malformed queries
- A clear message when an attribute cannot be used in a query, instead of an opaque database error
- One reviewable location per Query Language for how query text is built

Detailed design and task breakdown are tracked by the maintainers.

<audience-footer>

> [!IMPORTANT]
> Internal only — this issue is maintained by the core team and is not accepting external contributions.

</audience-footer>

PR #2027, opened 2026-07-30
Add Query Fragment modules for each query language

kmcginnes:
## Description

Introduces a Query Fragment module for each query language (`gremlin`, `openCypher`, `sparql`) that constructs pieces of query text carrying their own delimiters, so templates will not need to write quotes, backticks, or angle brackets around interpolated values.

This is a purely additive first step — the modules exist and are fully tested, but nothing calls them yet. It relocates the delimiter-and-escaping responsibility that today lives scattered across ~20 templates (via `escapeString` and the three per-connector `idParam` helpers) into one place per language. Behavior is preserved exactly; migrating the templates onto these modules is separate follow-up work.

Each module exposes only the positions its language supports: Gremlin and openCypher get `stringLiteral`, `identifier`, and `id`; SPARQL gets `stringLiteral` and `iri` (RDF has no bare identifiers, and an entity ID is itself an IRI). Fragments are minted in one place, `connector/queryFragment.ts`, so every construction path is auditable. The old `escapeString` and `idParam` helpers are marked deprecated and will be removed as the templates migrate.

## How to read

1. [connector/queryFragment.ts](https://github.com/aws/graph-explorer/pull/2027/files#diff-b3c44237a969c9dd5b5165eaa53360161d312b9db8214308248da44db7c8567b) — the `QueryFragment` type and the single mint point
2. [connector/gremlin/fragments.ts](https://github.com/aws/graph-explorer/pull/2027/files#diff-1d089533c25397d7fa71d381aa6345baad8801c33bca2ac12c8e9121e2256124) — representative module; openCypher mirrors it
3. [connector/sparql/fragments.ts](https://github.com/aws/graph-explorer/pull/2027/files#diff-f0b80fabd10e05efa4dc35727ef0ffcde7228d9f789996d857a295b97dde3edb) — differs: `iri` instead of `identifier`/`id`
4. [docs/agents/connectors.md](https://github.com/aws/graph-explorer/pull/2027/files#diff-aa67cd30579825aed72a294739be26b064de612e061e5875d25d40fd5290cb1f) — guidance pointing new query code at the modules

## Validation

- `pnpm checks` passes (types, lint, format)
- `pnpm test` passes — 30 new tests across the three fragment modules, assertions chosen to remain stable as the escaping rules are refined later
- No existing template or test was modified; no production code path reaches the new modules yet

## Related Issues

- Related to #2020

### Check List

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -35,6 +35,10 @@ _Avoid_: Seed connection
 The graph query protocol a Connection uses — one of Gremlin, openCypher, or SPARQL. Determines which Explorer is instantiated and implies the graph type (Gremlin/openCypher → property graph, SPARQL → RDF).
 _Avoid_: Query engine (internal code name `queryEngine`, but UI says "Query Language")
 
+**Query Fragment**:
+A piece of query text that carries its own delimiters, so a template interpolating it never writes quotes around it. Every value reaching a query — a string literal, an identifier such as a Property key or type label, or an IRI — becomes a Query Fragment first. Each Query Language has its own fragments; one language's fragment is not valid in another.
+_Avoid_: Query Parameter Value (implies parameterization, which no supported backend offers for identifiers), Query Literal (an identifier is not a literal), Fragment (unqualified — the connector code already calls a Vertex or Edge returned without its attributes a "fragment", which is a result rather than query text)
+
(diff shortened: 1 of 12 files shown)


kmcginnes:

PR #2032, opened 2026-07-31
Migrate openCypher templates to Query Fragment constructors

kmcginnes:
## Description

Routes value-interpolating openCypher templates through the per-language Query Fragment object added in #2027, instead of hand-writing quotes and backticks. Sibling of #2030 (the Gremlin migration).

- **Templates migrated** — `oneHopTemplate`, `keywordSearchTemplate`, `edgeConnectionsTemplate`, `edgesSchemaTemplate`, `verticesSchemaTemplate`, `vertexTypeCountTemplate`, plus the `vertexDetails` / `edgeDetails` / `neighborCounts` ID lists — now call `fragment.string(...)`, `fragment.identifier(...)`, and `fragment.id(...)`.
- **Deprecated `openCypher/idParam.ts` removed** now that nothing calls it.
- **Undelimited identifier positions are intentionally left unchanged** — the property accesses `v.${attr}` / `tgt.${name}` and the label positions `v:${type}` / `tgt:${type}` are still raw. Delimiting them changes the generated text for every name, so they move with the later rule-correction step, not this migration. This is documented in the `identifier` constructor.
- **`hasSearchTerm` guard restructured** to a narrowing form (`searchTerm ? fragment.string(searchTerm) : undefined`), since `Boolean(x)` doesn't narrow and the constructor takes a required string.
- **Behaviour is preserved** for values without a double quote. String and Date criterion values and the keyword search term now escape one consistently at the fragment boundary; each such change has a new test. Non-string criterion values are coerced with `String(...)` to preserve prior tolerance — the broader `Criterion.value: any` cleanup is tracked separately in #2031.

## How to read

1. [vertexDetails.ts](https://github.com/aws/graph-explorer/pull/2032/files#diff-677c3e38668ae13be9da35b19412c2d3755ed5fb13b3b4d769c2a2a53ee07fbe) — the simplest shape: an ID list moves from `idParam` to `fragment.id`
2. [keywordSearchTemplate.ts](https://github.com/aws/graph-explorer/pull/2032/files#diff-7a332287e3b15f4ff0645f60afbef031c5bae47cd9036e7adaf882d288ecde40) — the `hasSearchTerm` guard restructured to a narrowing form; vertex types and the search term move to fragments, while `v.${attr}` property positions stay raw for the later ticket
[...]

PR #2030, opened 2026-07-31
Migrate Gremlin templates to Query Fragment constructors

kmcginnes:
## Description

Routes every value-interpolating Gremlin template through the per-language Query Fragment object added in #2027, instead of hand-writing quote characters. This puts all Gremlin quoting/escaping behind one boundary.

- **All nine Gremlin query builders migrated** — `oneHopTemplate`, `keywordSearchTemplate`, `edgeConnectionsTemplate`, `edgesSchemaTemplate`, `verticesSchemaTemplate`, `vertexTypeCountTemplate`, plus the `vertexDetails` / `edgeDetails` / `neighborCounts` ID lists — now call `fragment.string(...)`, `fragment.identifier(...)`, and `fragment.id(...)`. No template writes its own quotes around an interpolated value.
- **Deprecated `gremlin/idParam.ts` removed** now that nothing calls it.
- **Behaviour is preserved** for values without a double quote. Values that contain one are now escaped consistently at the fragment boundary; every such change in generated text has a new quote-bearing test. (Broader character-set handling is deliberately left to the later rule-correction step.)

## How to read

1. [vertexDetails.ts](https://github.com/aws/graph-explorer/pull/2030/files#diff-34e2069acaf4e520a7f3ffa1aa8053ca8c50702c8ec2895a35bd1a982c796805) — the simplest shape: an ID list moves from `idParam` to `fragment.id`
2. [oneHopTemplate.ts](https://github.com/aws/graph-explorer/pull/2030/files#diff-88dac7bd19897e26fd2e354ba783714a944af45efe03374def286129634f7140) — the richest migration: ids, labels, criterion names, and String values all route through `fragment.*`; numeric/datetime values stay unquoted
3. [keywordSearchTemplate.ts](https://github.com/aws/graph-explorer/pull/2030/files#diff-a85e651228422134a4c7adf427e738d121f37a02aa7dfcdb479c4d1c18be5142) — vertex types, attribute names, and the search term move to fragments; the `has(id,...)` keyword position stays as-is

## Validation

`pnpm checks` passes (lint, format, types) and `pnpm test` passes (209 files, 2365 tests). New tests assert the exact escaped output at each site whose generated text changed.

## Related Issues

- Related to #2020

### Check List

- [x] I confirm that my contribution is made under the terms of the Apache 2.0 license.
- [x] I have verified `pnpm checks` passes with no errors.
- [x] I have verified `pnpm test` passes with no failures.
- [x] I have covered new added functionality with unit tests if necessary.
[...]

PR #2033, opened 2026-07-31
Migrate SPARQL templates to Query Fragment constructors

kmcginnes:
## Description

Routes value-interpolating SPARQL templates and filter helpers through the per-language Query Fragment object added in #2027, instead of the deprecated `idParam` helper and manual escaping. Sibling of #2030 (Gremlin) and #2032 (openCypher).

- **Templates and helpers migrated** — `oneHopNeighborsTemplate`, `keywordSearchTemplate`, `edgeConnectionsTemplate`, `predicatesByClassTemplate`, `classWithCountsTemplates`, `blankNodeOneHopNeighborsTemplate`, `filterHelpers`, plus the `vertexDetails` / `neighborCounts` ID lists — now build IRIs with `fragment.iri(...)` and the keyword search literal with `fragment.string(...)`.
- **Deprecated `sparql/idParam.ts` removed** now that nothing calls it.
- **Output is byte-identical.** `idParam` and `fragment.iri` both emit `<value>`; the search term's escaping matches the previous `escapeString` plus manual quotes. This migration adds no IRI content validation — that is a behaviour change and belongs with the later rule-correction step.
- **Left for the rule-correction step:** the two case-insensitive `regex(str(...), "...", "i")` substring filters keep their current structure and raw object interpolation. Converting those to string containment moves with that later work.
- One test assertion that pinned the exact non-string-ID error message is relaxed to match the reason (`/must be a URI/`), since the message now comes from the shared constructor.

## How to read

1. [filterHelpers.ts](https://github.com/aws/graph-explorer/pull/2033/files#diff-4dff80413b713d8e365e8686a764e9ac122d2d895d0c4b3218622ca16ef76b3d) — shared helpers: subject-class and excluded-vertex IRIs move to `fragment.iri`
2. [oneHopNeighborsTemplate.ts](https://github.com/aws/graph-explorer/pull/2033/files#diff-6b494fad3960c6decd5f2a717b41206dc30cdebfeb33e20e34703ef982533139) — the resource/type/predicate IRIs route through `fragment.iri`; the `c.object` regex position stays raw for the later rule-correction step
3. [keywordSearchTemplate.ts](https://github.com/aws/graph-explorer/pull/2033/files#diff-875e2b688a54339ae5f622d28341b29795b2261314937258e866fc2d912bfedf) — predicate IRIs to `fragment.iri` and the search literal to `fragment.string` (byte-identical to the old `escapeString` + quotes); the `regex(...)` structure is left for later

## Validation

`pnpm checks` passes (lint, format, types) and `pnpm test` passes (209 files, 2357 tests). This is a byte-identical migration, so it relies on the existing SPARQL template tests as its net rather than adding new ones.

## Related Issues

- Related to #2020

### Check List


kmcginnes:

PR #2034, opened 2026-07-31
Remove the shared escape helper

kmcginnes:
## Description

The contract step of the Query Fragment migration. With every query template now building its own delimited text through the per-language Query Fragment objects (#2030, #2032, #2033), the old shared single-character escape helper (`utils/escapeString`) has no callers left in query construction and is removed.

It had exactly one remaining **non-query** caller: the graph-export import path in `exportedGraph.ts`, which sanitizes double quotes in imported property-graph entity IDs at the import boundary. That behaviour is preserved by inlining it as a module-private `escapeDoubleQuotes` helper (byte-identical to the old logic) before deleting the shared helper, its barrel export in `utils/index.ts`, and its test.

- **Non-query caller handled by inlining**, not left dangling — behaviour is unchanged for the trimmed, non-empty entity IDs that reach it.
- **`utils/escapeString.ts`, its test, and the `@/utils` export removed.**
- **Byte-identical.** The inlined expression is character-for-character the old helper minus an empty-string guard that this caller never exercised (upstream `typeof` check + empty-ID filter guarantee only non-empty strings arrive).

## How to read

1. [exportedGraph.ts](https://github.com/aws/graph-explorer/pull/2034/files#diff-fff15adbd73a2138f9b414ab81ac9c59d68c997255303862d7701d3f189706a4) — the one behavioural anchor: the import-boundary sanitization now lives here as `escapeDoubleQuotes`. The rest of the diff is deletions (helper, test, barrel export) and the consuming test dropping its import.

## Validation

`pnpm checks` passes (lint, format, types) and `pnpm test` passes (208 files, 2367 tests — three fewer than before, matching the removed `escapeString` tests). This is a byte-identical relocation, so it relies on the existing `exportedGraph` tests as its net; the double-quote test was updated to assert the escaped output inline rather than via the deleted helper.

## Related Issues

- Related to #2020

### Check List


kmcginnes:

PR #2045, opened 2026-08-04
Require every SPARQL attribute filter to match

kmcginnes:
## What

On RDF/SPARQL connections, neighbor-expansion attribute filters were joined with `||`, so adding a second filter **widened** the results instead of narrowing them. Gremlin (`and(...)`) and openCypher (`AND`) have conjoined since their first commits, and this connector's own in-memory blank-node path has required every filter to match since 2023.

Each filter now gets its own `FILTER EXISTS`, so the filters conjoin. A filter is satisfied when *any* value of that attribute contains the value, which matters in RDF where a subject may repeat a predicate.

## Why it happened

Every filter shared one `?neighbor ?pValue ?object` triple pattern. Swapping `||` for `&&` on that shape requires a single `?pValue` to equal two attribute IRIs at once, so it matches nothing — verified, 0 rows. Conjunction needs the pattern restructured, so the disjunction is best read as the shape someone could get working rather than a decision. It has been present since the initial port commit `8a6eac7d` (2022-11-28); no commit, PR, or issue explains it, and no conjunction was ever attempted.

## Impact, measured

Against a purpose-built 1.19M-triple fixture, expanding a 50,000-member hub:

| filters | correct | was returned | error |
| --- | --- | --- | --- |
| 2 | 29 | 4,455 | 154x |
| 3 | 9 | 20,395 | 2,266x |

The error grew with each filter added, and nothing ever failed — users got wrong results quickly enough not to suspect them. Single-filter expansion was unaffected, which is why it survived. Dropping the unbound pattern also removes the dominant query cost, so the same hub expansion is now **~3x faster** (6.01s to 1.94s, using the query the code actually generates).

## Why `FILTER EXISTS` and not the faster form

One bound triple pattern per filter is measurably faster on hub expansion, but it regresses the **common** case — an ordinary low-degree node went 18ms to ~790ms, because Neptune applies the filters first and scans the whole attribute index. The only fix is `hint:joinOrder`, a Blazegraph/Neptune extension that is syntactically an ordinary triple pattern: other engines try to match it, find nothing, and **return zero rows with no error**. Verified against a live third-party store — 3 rows without the hint, 0 with it. Graph Explorer ships a BlazeGraph guide and supports arbitrary SPARQL 1.1 endpoints, so that was disqualifying. Full reasoning and the rejected alternatives are in the ADR (`docs/adr/20260804-sparql-attribute-filters-conjoin-via-filter-exists.md`).

## Second commit: two consequences of the conjunction

- A filter row with a **blank value** now means "the attribute must be present" rather than "no constraint", so blank rows are dropped when the request is built.
- An expansion returning nothing reported "No more neighbors to expand", which claims the node is exhausted. With a conjunction the likely cause is that the filters excluded everything, so the message now says so when the request carried filters.

## Notes for the reviewer

- No `Changelog.md` entry — that is a release-time change. Suggested wording for the next release section: *Multiple attribute filters on an RDF/SPARQL neighbor expansion now all have to match. Previously they were combined with OR, so adding a second filter widened the results instead of narrowing them. Gremlin and openCypher were already correct. Expect fewer — and correct — results from multi-filter expansions on RDF connections.*
- `CONTEXT.md` already asserted "every filter must match"; this makes that true rather than needing correction.
- **Ordering / known issue, handled separately:** moving the attribute name into a triple pattern changes what an unvalidated IRI can reach — a `>` in a predicate IRI previously broke out into a boolean expression, where the grammar rejects triple patterns, and now breaks out into a graph pattern where `GRAPH`/`SERVICE` are legal (both confirmed against Neptune). Validating the `fragment.iri` position is a **separate in-flight change and should land first**. The predicate name comes from schema discovery, not user input, so reaching it needs a poisoned predicate IRI in the store. Noted in the ADR's Consequences. This PR carries no `fragments.ts` / `queryFragment.ts` changes.
- Also unchanged, tracked separately: raw interpolation of the filter *value* into the regex, and the unescaped `new RegExp` in the blank-node path.
- Test coverage was mutation-checked: reverting to either disjunctive form fails the new tests.


kmcginnes:
Closing as no longer viable as-is. While this branch was open, main independently landed the two things this PR depended on or overlapped with: #2054 (forbid unrepresentable IRIs in `fragment.iri` — supersedes this PR's IRI-validation work and resolves the injection ordering concern) and #2058 (switch the SPARQL attribute filter to `CONTAINS`/`LCASE` substring matching, with a test forbidding `regex(`). This PR's `FILTER EXISTS` blocks were written with `regex(...)`, which now conflicts with #2058 both textually and semantically. The core contribution — the OR→AND conjunction bug fix — is still unfixed on main and remains worth doing; it will be re-approached fresh on top of the new `CONTAINS`/`LCASE` template rather than rebased. Superseding PR to follow.

Issue #2060, opened 2026-08-06
Misleading "No more neighbors to expand" message when filters exclude all neighbors


kmcginnes:
## Description

When expanding a node's neighbors with attribute filters, if the filters match nothing, the app shows the toast **"No more neighbors to expand"** — which implies the node is fully expanded. The truth is the opposite: the node has neighbors, but the applied filters excluded all of them. A user would reasonably conclude there is nothing left and stop, when relaxing the filter would reveal neighbors.

This is connector-agnostic — the toast lives in the shared expand hook (`useExpandNode`), so it affects Gremlin, openCypher, and SPARQL alike.

## Environment

- OS: N/A
- Browser: N/A
- Graph Explorer Version: current `main`
- Graph Database & Version: any (Gremlin / openCypher / SPARQL)

## Steps to Reproduce

1. Open the expand-neighbors sidebar for a node that has neighbors.
2. Add an attribute filter whose value matches none of them (e.g. a value no neighbor's property contains).
3. Expand.
4. See the toast "No more neighbors to expand".

## Expected Behavior

When the expansion request carried attribute filters and returned no results, the message should attribute the empty result to the filters — e.g. **"No neighbors matched your filters"** — rather than implying the node is exhausted. When no filters were applied, keep "No more neighbors to expand".

## Fix Path

The toast fires from two shared sites in `packages/graph-explorer/src/hooks/useExpandNode.tsx` (line 97, single-node `expandNode`; line 169, multi-node `expandNodes`), based purely on an empty result set with no inspection of the request. At both sites, check whether the request carried non-empty `attributeFilters`; if so, show the filter-specific message. Today only the single-node sidebar path supplies filters, but writing it at both sites keeps them correct if the multi path ever gains filters.

<audience-footer>

> [!IMPORTANT]
> If you are interested in working on this issue, please leave a comment.

> [!TIP]
> Please use a 👍 reaction to provide a +1/vote. This helps the community and maintainers prioritize this request.

</audience-footer>

Issue #2039, opened 2026-08-04
Investigate a shared attribute-match constructor per Query Language


kmcginnes:
## Goal

Attribute filtering and Keyword Search each build "does this attribute value match this user string" independently in every Query Language, so the same idea is implemented six times:

- `connector/gremlin/fetchNeighbors/oneHopTemplate.ts` — `attributeFilterTemplate`
- `connector/gremlin/keywordSearch/keywordSearchTemplate.ts` — inline in the `.or(…)` map
- `connector/openCypher/fetchNeighbors/oneHopTemplate.ts` — `attributeFilterTemplate`
- `connector/openCypher/keywordSearch/keywordSearchTemplate.ts` — inline in the `OR` map
- `connector/sparql/fetchNeighbors/oneHopNeighborsTemplate.ts` — `createFilterTemplate`
- `connector/sparql/keywordSearch/keywordSearchTemplate.ts` — `getFilterObject`

Within SPARQL the same `regex(str(…), …, "i")` expression is written out twice, in `createFilterTemplate` and `getFilterObject`. That duplication is why the semantics drifted apart in the first place (see the divergence table in #2038).

Note the two SPARQL call sites are not structurally parallel to the others: keyword search pairs one search term against a set of predicates (`FILTER (?pValue IN (…))` plus a single value filter), whereas expansion pairs each predicate with its own value. Any shared seam has to accommodate both shapes or deliberately serve only one.

Investigate introducing one attribute-match constructor per Query Language that both features call, alongside the existing `fragment` modules from #2020.

## Expected Outcome

A recommendation covering:

- Whether the seam belongs at the *predicate* (one attribute compared to one value) or higher at the whole where-clause. Note that Keyword Search carries per-language special cases that neighbor expansion has no use for — the node-ID pseudo-attribute, missing-type handling, and the all-attributes fan-out — which argues for the narrower seam.
- How the node-ID case is modelled. Today it travels as a sentinel string (`SEARCH_TOKENS.NODE_ID`, value `"__id"`) inside `searchByAttributes: string[]`, so "which kind of string is this" is re-derived in every template. A discriminated union — an attribute name versus the ID — would make it explicit. SPARQL cannot search by ID at all, and today the only thing preventing a malformed query is that the UI withholds the option (`useKeywordSearch.ts`); the constructor could reject it at the boundary via the existing `InvalidFragmentValueError` instead.
[...]

Issue #2038, opened 2026-08-04
Decide the intended attribute match semantics across Query Languages


kmcginnes:
## Goal

Neighbor expansion and Keyword Search both let a user type a value that an attribute must match, but the four code paths that implement the match disagree on what "match" means. Decide what the single intended semantic is, and how close each Query Language can get to it.

The divergences, as of `main`:

| | Gremlin | openCypher | SPARQL (query) | SPARQL (stored blank node, in memory) |
|---|---|---|---|---|
| Neighbor expansion | `containing(…)` — case-sensitive | `CONTAINS` — case-sensitive | `regex(str(?object), …, "i")` — case-insensitive | `RegExp(…, "gi")` — case-insensitive |
| Keyword Search, partial | `containing(…)` — case-sensitive | `CONTAINS` — case-sensitive | `regex(str(?value), …, "i")` — case-insensitive | n/a |
| Keyword Search, exact | `has(k, v)` | `= v` | `?value = v` — also full RDF term equality, so a language-tagged or typed literal will not match a plain one | n/a |

So the same typed value silently means different things depending on which database a user is connected to, and nothing records that as intentional.

Note the two SPARQL paths use regular-expression matching where the other languages use substring matching, so a value containing punctuation behaves differently there too.

The hard constraint that makes this a spike rather than a task: Graph Explorer's documented floor is TinkerPop 3.6.2 (Neptune 1.2.1.0), where the only substring predicate is `containing(…)`. `toLower()` arrives in 3.7.1, and Neptune forbids lambdas in every version — so a case-insensitive substring match is not straightforwardly reachable on the oldest supported Gremlin. Any convergence proposal has to say what happens there.

## Expected Outcome

- A decision on the intended match semantic for attribute filtering and Keyword Search partial match — case-sensitive or case-insensitive.
- A per-Query-Language conformance table showing what is achievable at each supported version floor, and what the fallback is where the intended semantic is unreachable.
- A recommendation on whether the remaining divergence should be declared and documented (e.g. an ADR plus a note in the contributor guide) or worked around.
- If a workaround is proposed for Gremlin at 3.6.2, an assessment of its query cost — index usage and whether it forces a scan.
- Whether SPARQL exact match should stay RDF term equality or compare on string value.
- A tasked-out follow-up, if the decision implies code changes.

## Related Issues

- Originated from #2031
- Related to #2020

<audience-footer>

> [!IMPORTANT]
[...]

PR #2061, opened 2026-08-06
Conjoin SPARQL neighbor attribute filters with AND instead of OR

kmcginnes:
## Description

On RDF/SPARQL connections, adding multiple attribute filters while expanding a node's neighbors combined them with **OR**, so each added filter *widened* the results — the opposite of the "Filter to narrow results" UI intent, and inconsistent with Gremlin and openCypher, which already AND their filters.

The template built one shared triple pattern `?neighbor ?pValue ?object` and OR-joined the per-filter conditions. This replaces that with **one `FILTER EXISTS` block per filter, conjoined by juxtaposition (AND)** — every filter must match. Each filter is existential over its own bound predicate, which is correct for RDF where a subject may repeat a predicate. The shared unbound pattern is dropped entirely (it was both the OR-widening bug and the dominant query cost).

Scope is *purely* OR→AND. IRI safety (`fragment.iri` throws on forbidden characters, #2054) and case-insensitive substring matching (`CONTAINS`/`LCASE`, #2058) already landed on `main` and are unchanged here.

## Validation

- New template tests pin the AND shape and guard against regression: multi-filter → two adjacent `FILTER EXISTS` blocks + `not.toContain("?pValue")`; single filter; two filters on the same attribute (existential); no filters → no `FILTER EXISTS`. The blank-node template inherits the change and has a matching test.
- `pnpm checks` and `pnpm test` (2600 tests) pass.
- Verified live against a 1.19M-triple Neptune 1.4.7.0 fixture: the emitted query returns the ground-truth **29** neighbors for a 50k-member hub with 2 filters, where the old OR form returned **4,455** (154× over-return). The verbatim emitted query also parses and executes with no syntax error.

## How to read

1. [oneHopNeighborsTemplate.ts](https://github.com/aws/graph-explorer/pull/2061/files#diff-6b494fad3960c6decd5f2a717b41206dc30cdebfeb33e20e34703ef982533139) — **core**: the OR→AND rewrite in `getFilterTemplate`
2. [oneHopNeighborsTemplate.test.ts](https://github.com/aws/graph-explorer/pull/2061/files#diff-2f11e9fbee5129fc086fd1fe93affd05612bffd88cd5985c2422de9d01474143) — **core**: its regression tests
3. [oneHopNeighborsBlankNodesIdsTemplate.test.ts](https://github.com/aws/graph-explorer/pull/2061/files#diff-1fccf59e80b8c5b3dd9c82454fa9ede9f20e1f049adcc38c2f6d701397f52630) — blank-node template test (inherits the fix via `findNeighborsUsingFilters`)
4. [ADR](https://github.com/aws/graph-explorer/pull/2061/files#diff-495b1142baec7c854ed1d3905bad7f59a4b2eca8cd7561a2339a35f81cb85219) — why `FILTER EXISTS` over the two rejected alternatives (naive `&&`; bound-pattern-per-filter needing a non-portable `hint:`)
5. connector-guidance update (`docs/agents/connectors.md`)

## Related Issues

- Related to #2060 (misleading "No more neighbors to expand" toast when filters exclude everything — a UX consequence surfaced by this change, tracked separately)
- Related to #2038 (deciding attribute-match *semantics* across query languages)
- Related to #2039 (shared attribute-match constructor tech debt)

### Check List


docs/features/graph-view.md
```diff
@@ -65,7 +65,7 @@ The Search UI provides two powerful ways to search and interact with your graph
 Provides fine grained control over neighbor expansions
 
 - Filter by node label (or rdf:type for RDF databases)
-- Filter by attribute value
+- Filter by attribute value; when you add more than one attribute filter, all of them must match
```
kmcginnes: unnecessary doc change. We can leave it ambiguous as this is the expected behavior

CONTEXT.md
```diff
@@ -50,7 +50,7 @@ Vertices directly connected to a given vertex (one hop away). Users "expand neig
 _Avoid_: Connections (ambiguous with Connection)
 
 **Attribute Filter**:
-A named property paired with a value that a neighbor's property must contain for that neighbor to be expanded. Only string properties are filterable, and every filter must match. Users add these when expanding neighbors to narrow the results.
+A named property paired with a value that a neighbor's property must contain for that neighbor to be expanded. Only string properties are filterable, and every filter must match; a filter matches when any one of a property's values contains the value. Users add these when expanding neighbors to narrow the results.
```
kmcginnes: This also seems unnecessary and the expected behavior

docs/features/graph-view.md
kmcginnes: Reverted — dropped the graph-view.md line.

CONTEXT.md
kmcginnes: Reverted — CONTEXT.md is back to its `main` wording (no change).

PR #2063, opened 2026-08-07
Split composite labels once at the Gremlin ingestion boundary

kmcginnes:
## Description

Amazon Neptune joins a vertex's multiple labels with the reserved `::` delimiter and surfaces the composite as a single string on the Gremlin wire. Graph Explorer's domain model holds only single labels, so the split belongs **once, at ingestion** — not scattered across sites, and not re-applied to values that are already single labels.

Previously the `::` split was inlined at eight connector sites with two inconsistent handlings (`.split("::").filter(Boolean)` at some, bare `.split("::")` at others), including sites that were re-splitting Graph Explorer's own already-single-label domain ids.

This PR:

- Adds a shared Gremlin `splitLabel` helper that owns the `::` convention and **keeps the whole string when any segment is empty** (`foo::`, `::foo`, `a::::b`, `::`), rather than silently resolving `foo::` to `foo` or emitting an empty label.
- Routes the **five genuine ingestion sites** (raw Neptune wire labels) through it: `mapApiVertex`, `fetchEdgeConnections`, `verticesSchemaTemplate`, `edgesSchemaTemplate`, `neighborCounts`.
- **Deletes** the split at the **three sites that operated on already-single-label domain ids** (`gremlin/fetchNeighbors/oneHopTemplate`, `keywordSearch/keywordSearchTemplate`, `openCypher/fetchNeighbors/oneHopTemplate`). Re-splitting these was a no-op on real data and wrong for a generic TinkerPop backend where `::` is an ordinary label character.
- `mapApiVertex` maps an empty element label to no types (`[]`), preserving prior behavior for that boundary without reintroducing the `foo:: → foo` bug.

The rationale, the live Neptune/TinkerPop evidence, and the known limitation (the split is unconditional and only correct for Neptune) are recorded in the ADR. Gremlin-only: openCypher returns labels as an array and never carries `::`.

## How to read

1. [packages/graph-explorer/src/connector/gremlin/splitLabel.ts](https://github.com/aws/graph-explorer/pull/2063/files#diff-bdaf9057861676f5fdff033211fbdbe8035c39348b6d435706218c6d8a54e2e5) — start here; the helper and the keep-whole-on-empty rule.
2. [docs/adr/20260807-split-composite-labels-once-at-gremlin-ingestion.md](https://github.com/aws/graph-explorer/pull/2063/files#diff-35ae6a3a35c4026e1d72da76382e4d42b3fb01d65a03acf0ed4a402d91db0c91) — why the split lives at ingestion, and the empty-segment decision.
3. The five ingestion call sites, then the three delete sites (with their tests flipped to verbatim assertions).

## Validation

- `pnpm test` passes (2609 tests). `splitLabel` has full awkward-input coverage; `mapApiVertex` gains composite-split and empty-label cases; the three delete-site tests assert verbatim behavior.

## Related Issues

- Related to #2020

### Check List

### DIFF
--- a/packages/graph-explorer/src/connector/gremlin/splitLabel.test.ts
+++ b/packages/graph-explorer/src/connector/gremlin/splitLabel.test.ts
@@ -1,0 +1,35 @@
+import { splitLabel } from "./splitLabel";
+
+describe("splitLabel", () => {
+  it("returns a single-element array for a plain label", () => {
+    expect(splitLabel("foo")).toEqual(["foo"]);
+  });
+
+  it("splits a composite label on the :: delimiter", () => {
+    expect(splitLabel("foo::bar")).toEqual(["foo", "bar"]);
+  });
+
+  it("splits a composite label with more than two segments", () => {
+    expect(splitLabel("foo::bar::baz")).toEqual(["foo", "bar", "baz"]);
+  });
+
+  it("keeps a trailing :: verbatim rather than dropping the empty segment", () => {
+    expect(splitLabel("foo::")).toEqual(["foo::"]);
+  });
+
+  it("keeps a leading :: verbatim rather than dropping the empty segment", () => {
+    expect(splitLabel("::foo")).toEqual(["::foo"]);
+  });
+
+  it("keeps a doubled :: verbatim rather than dropping the empty segment", () => {
+    expect(splitLabel("a::::b")).toEqual(["a::::b"]);
+  });
+
+  it("keeps a lone :: verbatim", () => {
+    expect(splitLabel("::")).toEqual(["::"]);
+  });
+
+  it("returns the empty string unchanged", () => {
+    expect(splitLabel("")).toEqual([""]);
+  });
+});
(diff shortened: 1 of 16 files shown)


packages/graph-explorer/src/connector/gremlin/splitLabel.test.ts
```diff
@@ -0,0 +1,35 @@
...
+    expect(splitLabel("foo::")).toEqual(["foo::"]);
+  });
+
+  it("keeps a leading :: verbatim rather than dropping the empty segment", () => {
+    expect(splitLabel("::foo")).toEqual(["::foo"]);
+  });
+
+  it("keeps a doubled :: verbatim rather than dropping the empty segment", () => {
+    expect(splitLabel("a::::b")).toEqual(["a::::b"]);
+  });
```
kmcginnes, line 26: We should also add tests for other counts of `:` in the string. Neptune's delimiter is precisely two colons. So `foo:::bar` and `foo:bar` is not split.

PR #2119, opened 2026-08-14
Resolve graph style data once per type instead of once per element

kmcginnes:
## Description

Style data varies only by type, but was computed once per element. Within a render pass N nodes of a type now cost one `vertexStyleData` call rather than N.

The memoizing closure lives in a plain `create*Resolver` factory rather than inside the hook because the React Compiler lint rules reject a hook that returns a closure mutating its own captured cache. That constraint is non-obvious — a two-line hook forwarding to a factory reads like a pointless wrapper — so it is recorded in `docs/agents/react.md` in a later PR in this stack.

The vertex resolver takes the styles whose icons are in scope, because the canvas needs only the types it draws while the schema view needs every type. That parameter is what the next PR uses to cut the canvas down from every schema type.

Also here:

- `VertexStyleLookup` / `EdgeStyleLookup` name the shape of the two style atoms, so the contract is explicit rather than inferred at each call site.
- `useAllEdgeStyles` is deleted — it has no callers now that `useGraphStyles` no longer consumes edge styles.
- The style-context regression test asserts the exact selector set instead of `toBeLessThanOrEqual(6)`, which would have passed with three extra selectors.

## How to read

1. `styleDataResolvers.ts` — new; the resolvers and the factories.
2. `renderedEntities.ts` / `useSchemaGraphData.ts` — call sites.
3. `graphStyles.ts` — the lookup types and the dead-code deletion.

## Stack

> [!NOTE]
> A later PR in this stack (`fuse-canvas-vertex-passes`) **replaces** this resolver layer with an eager per-type map, after review found the lazy memoization bought nothing when every caller knows its key set up front. This PR is kept as its own step because the per-type win it introduces is independent of that restructuring and the intermediate state is green. If you would rather not land the resolver layer at all, merge this and the fusing PR together.

### DIFF
--- a/packages/graph-explorer/src/components/LabelPreview.test.tsx
+++ b/packages/graph-explorer/src/components/LabelPreview.test.tsx
@@ -20,9 +20,10 @@ function renderLabel(style: LabelVisualStyle, scale = 2) {
 
 describe("LabelPreview", () => {
   describe("text color follows label darkness for contrast", () => {
+    // Casing follows `labelTextColorFor`, the helper the canvas shares.
     it("uses white text on a dark label color", () => {
       const el = renderLabel(labelStyle({ labelColor: "#1d2531" }));
-      expect(el.style.color).toBe("#ffffff");
+      expect(el.style.color).toBe("#FFFFFF");
     });
 
     it("uses black text on a light label color", () => {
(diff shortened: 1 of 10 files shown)


kmcginnes:

Issue #2116, opened 2026-08-14
Unbounded atomFamily growth: families keyed on freshly allocated arrays and objects are never released


kmcginnes:
`atomFamily` from `jotai-family` caches one atom per parameter identity and never evicts unless `remove` or `setShouldRemove` is called. Nothing under `packages/graph-explorer/src` calls either — verified, zero matches. That is fine for a family keyed on a stable branded ID, where the key set is bounded by the data. It is not fine for a family keyed on a value that is freshly allocated on every recomputation: each recomputation interns a new entry that can never be reached again, and each entry retains whatever its derived atom computed. Retained memory then grows with the number of state mutations rather than with the size of the data.

Neighbor expansion is the app's primary interaction and it mutates `nodesAtom`/`edgesAtom` on every expand, so this grows during ordinary use. It is not a crash and there is no recovery path short of closing the tab.

## Why it happens

`useAllNeighbors` is the clearest case. It derives the key array from the node map, so the array is a new object on every `nodesAtom` change:

```ts
// packages/graph-explorer/src/core/StateProvider/neighbors.ts:98-103
export function useAllNeighbors() {
  const vertices = useAtomValue(nodesAtom);
  const vertexIds = useMemo(() => vertices.keys().toArray(), [vertices]);
  ...
  const fetchedNeighbors = useAtomValue(allFetchedNeighborsSelector(vertexIds));
```

The `useMemo` stabilises the array across renders but not across store mutations, which is exactly the axis that matters here. Every expand interns a fresh entry in `allFetchedNeighborsSelector`, and that entry holds a `Map<VertexId, Vertex[]>` over the whole graph as it stood at that moment.

## Sites

All keyed on a value allocated fresh per recomputation:

1. `core/StateProvider/neighbors.ts:229` — `allFetchedNeighborsSelector = atomFamily((ids: VertexId[]) => …)`. **Highest priority**: same shape and severity as the one already fixed, in the neighbor path, driven by the same interaction, and each retained entry is proportional to the whole graph.
2. `core/StateProvider/displayEdge.ts:67` — `displayEdgeSelector = atomFamily((edge: Edge) => …)`, keyed on object identity. A re-derived `Edge` interns a second entry for an edge that already had one. Reached from three call sites, including the map over the full edge set at `displayEdge.ts:129`.
[...]

PR #2120, opened 2026-08-14
Share one visible-vertex computation and scope canvas icons to drawn types

kmcginnes:
## Description

Two related fixes to how much work a render does.

**The vertex pipeline ran twice per render.** `GraphViewer` calls `useRenderedVertices()` directly *and* `useRenderedEdges()`, which called `useRenderedVertices()` internally to build its set of valid endpoints. Every hook has its own render-scoped memo, so the whole pipeline — icon resolution included — ran once per call site. The filter predicate is now a derived atom both pipelines read, so the store computes it once.

**Canvas style and icon resolution covered every type in the schema.** `useBackgroundImageMap(useAllVertexStyles())` iterated all of them on every render. On the stress schema that is **10,044 vertex types resolved to draw 3**. `canvasVertexStylesAtom` scopes it to the types actually on the canvas. The schema view keeps `useAllVertexStyles`, since drawing every type is its job.

## Measurement

Chrome DevTools against a live 10k-type Neptune schema, expanding a 244-neighbour airport with limit 75 (76 nodes / 150 edges). Matched 10s windows from the click, ~81k samples each:

| self time in window | before | after this PR | after scoping (this PR) |
| --- | --: | --: | --: |
| `useBackgroundImageMap` | 59.7ms | ~29ms | **0.0ms** |
| style work in `renderedEntities` | 5.2ms | ~0.7ms | ~0.3ms |

The 59.7 → 29ms halving is the double-call fix, and matches the predicted 50% exactly. The drop to 0.0ms is the scoping — the function no longer appears in the profile at all.

**What this does not fix.** Total busy JS and INP were within run-to-run noise, and the dominant expansion cost remains cytoscape rendering plus the fcose layout (~390-470ms long task), which this does not touch. This removes schema-scale style work from the render path; it is not a fix for the reported expansion choppiness.

## How to read

1. `renderedEntities.ts` — `visibleVertexIdsAtom` and `canvasVertexStylesAtom`.
2. `displayVertex.ts` — one selector exported so the atoms can read it.

> [!NOTE]
> The comment on `visibleVertexIdsAtom` records that it still recomputes on a style change, because `displayVerticesInCanvasSelector` resolves display labels through the style atoms. Decoupling that is tracked in #2116.

## Stack

Both atoms are fused into one in `fuse-canvas-vertex-passes` later in this stack.


kmcginnes:

PR #2122, opened 2026-08-14
Always set every style data field so a stale value cannot be stranded

kmcginnes:
## Description

**A node could keep displaying a stale icon forever, with no error.**

`cy.json({ elements })` *merges* element data: `ele.data(obj)` adds and overwrites keys but never deletes ones missing from the new object. Both of the optional style data fields were therefore unclearable once applied. Concrete path: a user replaces a type's icon, the new icon fails to load, `iconRegistry` retries three times and gives up, the type drops out of the background-image map, `__iconUrl` goes absent — and the node silently keeps rendering the **previous** icon while its colour updates around it. The user gets no signal the upload failed and reasonably concludes it worked.

This is a regression from the `data()` migration in #2112, not preserved behaviour: the old stylesheet path set `background-image` unconditionally, so `undefined` cleared it.

The fix is to make the fields total rather than to gate on presence. `ge_iconUrl` carries `"none"` (cytoscape's no-image value) when a type has no icon, and `ge_lineDashPattern` carries cytoscape's default `[6, 3]` for solid lines, which ignore it. Both then live on the base `node` / `edge` rule and always reflect current state, so there are **no gated selectors left** — the stylesheet is two rules.

Note that adding `background-image: none` to the base rule alone would *not* have fixed this: the gated selector keeps matching while the key is still present, and overrides it.

Two side benefits: the icon gate moves out of the generic `components/Graph` component next to the producer that feeds it, and `__iconUrl` is renamed `ge_iconUrl` to match its siblings. `ge_lineDashPattern` had the same defect latently — harmless today because cytoscape ignores the pattern for solid lines, but the same shape.

## How to read

1. `graphElementStyleData.ts` — the `ALWAYS_SET` rationale and the two now-total fields.
2. `useGraphStyles.ts` — mappers moved onto the base rules, gated selector deleted.
3. `useManageStyles.ts` — the hardcoded gate deleted from the generic component.
4. The ADR gains the merge-semantics constraint, since it is the reason for the design.


kmcginnes:

PR #2124, opened 2026-08-14
Resolve canvas vertices, ids, and style scope in one pass

kmcginnes:
## Description

Three passes over the canvas vertices become one, and the resolver layer from #2119 collapses.

The visible-ids atom iterated every vertex; the style-scope atom iterated them again re-testing membership; `useRenderedVertices` iterated a third time re-testing it once more. Two passes existed only to re-discover a decision the first had already made. `canvasVerticesAtom` returns the drawn vertices, their ids, and the styles of the types they draw from a single loop.

That also removes the reason a missing icon used to be **silent**. Style data was assembled from two inputs of different totality — scalar fields from a lookup that answers for any type, the icon from a partial scope list — so `resolveStyleData(type)` always succeeded and always looked plausible, and a scope miss was indistinguishable from a type that genuinely has no icon. Now a drawn vertex's `primaryType` is in `stylesByType` **by construction** (same loop), the style data derives entirely from that scoped list, and a type absent from it throws rather than quietly drawing a node without its icon.

The memoizing resolver layer goes with it: five exported symbols, two factories and two cache closures replaced by one hook returning a plain per-type map. The lazy memoization bought nothing because every caller knows its full key set up front — and the resolver *function* shape is what forced the total-lookup dependency, since a function must answer for any argument where a map need not.

The edge half is inlined rather than kept as a hook: the drawn edge types are only knowable while filtering, so producing them up front would need the extra pass this PR removes. Each consumer fills a per-render map on first sight of a type inside its existing loop — still one `Color` parse per type rather than per edge.

Iteration order is unchanged; rendered vertices stay in canvas insertion order.

## How to read

1. `renderedEntities.ts` — `canvasVerticesAtom`, then the two hooks.
2. `styleDataResolvers.ts` — down to a single hook.
3. `useSchemaGraphData.ts` — mirrors the same invariant assertion.

### DIFF
--- a/packages/graph-explorer/src/core/StateProvider/graphStyles.ts
+++ b/packages/graph-explorer/src/core/StateProvider/graphStyles.ts
@@ -270,7 +270,7 @@ export function resolveEdgeStyle(
 
 /** Returns an array of vertex styles based on the known vertex types in the schema.
  * For the schema view, which draws every type; the canvas scopes itself to the
- * types it draws via `canvasVertexStylesAtom`. Always includes an entry for
+ * types it draws via `canvasVerticesAtom`. Always includes an entry for
  * `LABELS.MISSING_TYPE` so blank nodes (assigned that synthetic type at runtime)
  * are styled rather than skipped.
  */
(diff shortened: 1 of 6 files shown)


kmcginnes:

PR #2125, opened 2026-08-14
Enforce the style data producer/consumer lockstep structurally

kmcginnes:
## Description

The ADR from #2112 records this as a Consequence:

> **Producer and consumer must stay in lockstep.** A new per-type style property must be added in two places together — the `ge_*` field + its producer, and the matching `data(…)` mapper. Editing one side alone silently drops the style.

Nothing enforced it, and the migration had already dropped assertions for `ge_labelBackgroundOpacity`, `ge_labelBackgroundColor`, `ge_labelBorderWidth`, `ge_labelBorderColor`, `ge_labelBorderStyle` and the arrow-colour / distance mappings — asserted nowhere.

This closes the loop by comparing **key sets** rather than listing field names, so it cannot rot the way a hand-written list does. Both directions:

- every `ge_*` key the producers emit appears as a `data(...)` mapper somewhere in the stylesheet
- every `ge_*` mapper corresponds to a key a producer emits

Non-`ge_*` mappers (`displayName`) come from the rendered entity rather than these producers, so they're excluded by prefix, not by name. It walks every rule rather than hardcoding `node` / `edge`, so a legitimately re-added gated selector is still covered.

Verified by deliberate probe: adding a bogus mapper and a bogus producer field each failed the expected direction, then both probes were reverted.

## How to read

One new `describe` block at the end of `useGraphStyles.test.tsx`.

### DIFF
--- a/packages/graph-explorer/src/modules/GraphViewer/useGraphStyles.test.tsx
+++ b/packages/graph-explorer/src/modules/GraphViewer/useGraphStyles.test.tsx
@@ -1,6 +1,14 @@
 // @vitest-environment happy-dom
 import { describe, expect, it } from "vitest";
 
+import {
+  createEdgeType,
+  createVertexType,
+  edgeStyleData,
+  type EdgeStyle,
+  vertexStyleData,
+  type VertexStyle,
+} from "@/core";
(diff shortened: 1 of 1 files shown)


kmcginnes:

PR #2126, opened 2026-08-14
Extract the rendered entity ID codec out of renderedEntities

kmcginnes:
## Description

`renderedEntities.ts` had accumulated four unrelated concerns: a pure string ID codec, the canvas visibility atom, the style scope, and the cytoscape render hooks. Only the last is what the module name describes.

The codec has no coupling to the rest — no React, no Jotai, no state — and was exactly the unit the first four test blocks covered, so it moves out whole along with them. 263 lines down to 193. Pure move: no behaviour, signature, or exported name changes, and the barrel keeps every symbol public at the same path.

No deep import paths needed updating; every consumer already imported through `@/core`.

## How to read

`renderedEntityIds.ts` and its test are the moved code. The diff on `renderedEntities.ts` is deletions plus one import.

### DIFF
--- a/packages/graph-explorer/src/core/StateProvider/index.ts
+++ b/packages/graph-explorer/src/core/StateProvider/index.ts
@@ -10,6 +10,7 @@ export * from "./featureFlags";
 export * from "./neighbors";
 export * from "./nodes";
 export * from "./renderedEntities";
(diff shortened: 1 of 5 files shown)


kmcginnes:

Issue #2086, opened 2026-08-12
Virtualized searchable Combobox for label-scale selects (10k+ node/edge labels)


kmcginnes:
## Problem Statement

When a graph has 10,000+ distinct vertex-type labels and 10,000+ edge-type labels, Graph Explorer's UI locks up. The worst offenders are the node-type/attribute pick-lists rendered as plain Radix `<Select>` dropdowns: opening one mounts every option at once, freezing the browser. Because keyword search and the Data Explorer node-type picker are core, unavoidable flows, users on large schemas effectively cannot use the product.

## Solution

Introduce a single reusable, virtualized, searchable **Combobox** component built on Base UI (`@base-ui/react`) with `@tanstack/react-virtual` for windowing, and swap it into the three label-scale select surfaces. Users get a type-to-filter input backed by a windowed result list that stays instant regardless of how many labels exist. Existing small-list selects keep using the current `SelectField`/Radix `Select` untouched.

## User Stories

1. As a user of a graph with 10,000+ vertex labels, I want the keyword-search node-type picker to open instantly, so that the app doesn't freeze when I start a search.
2. As a user, I want to type into the node-type picker to filter labels, so that I can find a type without scrolling thousands of entries.
3. As a user, I want the filtered results to appear immediately as I type, so that searching feels responsive.
4. As a user, I want to pick an attribute to search by from a large attribute list without the UI locking up, so that I can scope my keyword search.
5. As a user, I want the Data Explorer node-type picker to open and filter instantly on a large schema, so that I can browse entities of a chosen type.
6. As a user, I want to select a label with the keyboard (arrow keys + Enter), so that I can operate the picker without a mouse.
7. As a keyboard/screen-reader user, I want the combobox to announce roles, active option, and position correctly, so that it's accessible even though the list is virtualized.
8. As a user, I want to see a clear "no results" state when my filter matches nothing, so that I know the query was applied.
9. As a user, I want the currently selected label shown in the closed control, so that I know my current selection.
10. As a user, I want the filter input to reset when I reopen the picker, so that I always start from the full list.
11. As a user, I want a disabled picker to look and behave as disabled, so that I understand when selection isn't available.
12. As a user with a normal (small) schema, I want the existing selects to keep working exactly as before, so that nothing regresses for me.
13. As a developer, I want one Combobox component encapsulating virtualization and filtering, so that the rest of the app is shielded from those library decisions.
[...]

Issue #1680, opened 2026-04-17
Virtualize UI components that render all node types


kmcginnes:
## Description

Several UI components render all node types into the DOM without virtualization. With 70k+ types, opening a dropdown or viewing the schema graph creates tens of thousands of DOM nodes and freezes the browser.

### 1. Replace Radix Select with virtualized combobox for node type dropdowns

The search sidebar node type dropdown (`FilterSearchTabContent.tsx`) and the Data Explorer type selector (`DataExplorer.tsx`) use Radix Select which renders all items to the DOM when opened.

Replace with a searchable combobox using Radix Popover + react-virtuoso that only renders visible items and supports type-ahead filtering.

### 2. Paginate or cluster the schema graph view (`useSchemaGraphData.ts`)

The schema graph creates a Cytoscape node for every vertex type in the schema. With 70k types, Cytoscape cannot lay out or render this many nodes. Consider pagination, clustering by prefix/namespace, or a search-driven exploration mode.

## Related Issues

- Part of #1677

Issue #1677, opened 2026-04-17
Support graphs with very large numbers of node types (70k+)


kmcginnes:
## Description

Graph Explorer does not handle schemas with very large numbers of node types (70k+) gracefully. Multiple areas of the codebase assume a small-to-moderate number of types and perform operations that scale poorly — O(n²) merge loops, per-type Cytoscape style selectors, per-type TanStack Query instances, non-virtualized dropdowns, and thousands of sequential network requests during schema sync.

The result is that the React app locks up or becomes unresponsive when connected to a graph with tens of thousands of node types.

## High-Level Plan

Break the work into four phases, each independently shippable:

1. **Eliminate O(n²) in the state layer** — Index arrays into Maps in `mergeConfiguration()`, convert preference lookups from `.find()` to Map-based, replace `cloneDeep` with shallow spread.

2. **Scope graph styles to canvas types only** — Change `useGraphStyles`, `useBackgroundImageMap`, and the Legend to only process types that are actually present on the graph canvas, not all schema types.

3. **Virtualize UI for large type lists** — Replace non-virtualized Radix Select dropdowns with a searchable, virtualized combobox for node type selection. *(In progress on `large-schema-performance` branch.)*

4. **Improve schema sync for large schemas** — Add progress reporting and cancellation, increase batch concurrency, and consider lazy attribute fetching.

---

> [!IMPORTANT]
> If you are interested in working on this issue, please leave a comment.

> [!TIP]
> Please use a 👍 reaction to provide a +1/vote. This helps the community and maintainers prioritize this request.

kmcginnes:
## Architectural Context for the Styled Schema Redesign (#1725–#1729)

A detailed plan is saved in the repo at `plans/styled-schema-merge-redesign.md`.

### Key findings from the design session

**The true schema shape is `createGraphSchema()`, not `VertexTypeConfig`**

`VertexTypeConfig` extends `VertexPreferencesStorageModel`, making it look like schema carries style data. In reality, the database only provides `type` and `attributes` per vertex/edge type. The function `createGraphSchema()` in `schema.ts` already strips this down to the pure form — `{ type, attributes }` stored in Maps. The style fields on `VertexTypeConfig` are only populated post-merge by `mergeConfiguration()`. The persisted schema in IndexedDB does not contain style data.

**`displayNameAttribute` and `longDisplayNameAttribute` are preferences, not schema**

[...]

PR #2131, opened 2026-08-18
Virtualized Combobox component and keyword-search node-type picker migration

### LINKED ISSUE #2087: Virtualized searchable Combobox component + keyword-search node-type picker
## Parent

- Part of #2086

## What to build

A reusable, virtualized, searchable **Combobox** component, wired into the keyword-search node-type picker so it works end-to-end on a large schema. Adds `@base-ui/react` and `@tanstack/react-virtual` (imported via subpath exports to preserve tree-shaking). The Combobox is single-select, controlled (`value`/`onValueChange`), takes `{label, value}` items, and supports placeholder, disabled, and a "no results" empty state. Accessibility comes from Base UI (W3C APG combobox pattern); windowing from TanStack Virtual using Base UI's virtualized pattern (`virtualized` prop + `useFilteredItems()` + `onItemHighlighted` → `scrollToIndex`). Filtering starts with the default path plus `useDeferredValue()` on the query.

On a graph with 10,000+ vertex labels, opening the keyword-search node-type picker is instant, filters as the user types, and selects with mouse or keyboard.

## Acceptance criteria

- [ ] Combobox component exists, styled to the design system, single-select and controlled
- [ ] Base UI + TanStack Virtual added via subpath imports; new deps confined behind the component
- [ ] Type-to-filter narrows the list; selection calls `onValueChange` with the correct value; placeholder, disabled, and empty state work; arrow+Enter keyboard selection works
- [ ] Component unit tests cover filter/select/keyboard and a 10,000-item interactivity guardrail (filter to a late item and select it) without asserting DOM node counts
- [ ] Keyword-search node-type select uses the Combobox, preserving its existing Jotai-atom-controlled wiring
- [ ] `pnpm checks` and `pnpm test` pass

## Blocked by

- None — can start immediately.

### LINKED ISSUE #2088: Memoize useSearchableAttributes to stop per-render scan-and-sort
## Parent

- Part of #2086

## What to build

Stop `useSearchableAttributes` from recomputing a full scan-and-sort of searchable attributes on every render. Memoize it (via `useMemo` keyed on its inputs, or convert it to a Jotai `atomFamily` keyed by vertex type) so it caches like the other display-config selectors. No user-visible UI change on its own; this removes the per-keystroke hot-path cost the attribute picker depends on.

## Acceptance criteria

- [ ] `useSearchableAttributes` no longer allocates/scans/sorts on every render when inputs are unchanged
- [ ] Hook test on a large schema asserts the returned array is referentially stable across re-renders with unchanged inputs
- [ ] `pnpm checks` and `pnpm test` pass

## Blocked by

- None — can start immediately.

### LINKED ISSUE #2089: Swap keyword-search attribute picker to the virtualized Combobox
## Parent

- Part of #2086

## What to build

Swap the keyword-search **attribute** select from the raw Radix `Select` to the virtualized searchable Combobox. On a graph whose selected vertex type has a large searchable-attribute set, the attribute picker opens instantly, filters as the user types, and selects, feeling instant because the attribute derivation is now memoized.

## Acceptance criteria

- [ ] Keyword-search attribute select uses the Combobox, preserving its existing `selectedAttributeAtom`-controlled wiring
- [ ] Instant open + filter + select on a large attribute set
- [ ] `pnpm checks` and `pnpm test` pass

## Blocked by

- #2087 (Combobox component)
- #2088 (useSearchableAttributes memoization)

### LINKED ISSUE #2090: Swap Data Explorer node-type picker to the virtualized Combobox
## Parent

- Part of #2086

## What to build

Swap the Data Explorer **node-type picker** from `SelectField` to the virtualized searchable Combobox, preserving its router-driven selection wiring. On a graph with 10,000+ vertex labels, the picker opens and filters instantly.

## Acceptance criteria

- [ ] Data Explorer node-type picker uses the Combobox; route navigation on selection is unchanged
- [ ] Instant open + filter + select on a 10,000-label schema
- [ ] `DataExplorer.test.tsx` exercises the new component and passes
- [ ] `SelectField` and its other callers remain untouched
- [ ] `pnpm checks` and `pnpm test` pass

## Blocked by

- #2087 (Combobox component)

mjuarros:
## Description

SelectField renders every option as a DOM node unconditionally. With a schema of 10,000+ vertex/edge types, this froze the UI for several seconds every time the Data Explorer node-type picker, the Search Sidebar's node-type filter, or its attribute filter was opened.

This replaces those three pickers with a new Combobox component built on Base UI's accessible combobox primitives plus @tanstack/react-virtual for windowed rendering, so only the visible options (~20) are ever mounted regardless of how many types the schema has. Type-to-filter narrows the list as you type. Every other SelectField usage in the app is untouched — they're all either fixed-size enums or scoped to a single vertex/edge type's own attributes, so they were never at risk of this.

## How to read

1. `packages/graph-explorer/src/components/Combobox.tsx` — the new component; start here
2. `packages/graph-explorer/src/routes/DataExplorer/DataExplorer.tsx`, `packages/graph-explorer/src/modules/SearchSidebar/FilterSearchTabContent.tsx`, `packages/graph-explorer/src/modules/SearchSidebar/useKeywordSearch.ts` — the three sites swapped from SelectField to Combobox
3. `packages/graph-explorer/src/core/StateProvider/displayTypeConfigs.ts` — a small pre-existing memoization bug in useSearchableAttributes fixed along the way (found while restoring a referential-stability guarantee post-migration); tangential to the main change
4. `packages/graph-explorer/src/components/Combobox.test.tsx`, `packages/graph-explorer/src/modules/SearchSidebar/useKeywordSearch.test.ts` — tests, including a 10,000-item scale guardrail

## Validation

Manually validated against a synthetic 10,000-vertex-type / 10,000-edge-type dataset: the Data Explorer and Search Sidebar pickers previously froze the page for several seconds on open; they now open instantly regardless of schema size. Confirmed with an isolated real-Chromium test harness that open latency stays flat (~30ms) across 10, 500, and 10,000 items — a virtualizer-wiring bug that briefly regressed this was caught the same way and fixed before landing.

`pnpm checks` and `pnpm test` pass (219 test files, 2696 tests).

## Related Issues

- Closes #2087
- Closes #2089
- Closes #2090
- Related to #2086
- Related to #1680
- Related to #1677

### Check List

Delay combobox implementation with no text type option
delayedCombobox

New virtualized combobox with text type option
virtualizedCombobox

- Closes #2088


pnpm-workspace.yaml
```diff
@@ -11,6 +11,7 @@ overrides:
   flatted@<3.4.2: ">=3.4.2"
   form-data@>=4.0.0 <4.0.6: ">=4.0.6"
   minimatch@<10.2.3: ">=10.2.3"
+  nanoid@<3.3.18: ">=3.3.18"
```
noel-improv: Can you confirm this override is intended as unbounded? The lockfile was already on 3.3.18, which is outside `<3.3.18`, but it still moves to nanoid 6.0.1 (ESM-only, `engines.node: ^22 || ^24 || >=26`, no `require` export condition) and postcss loads `nanoid/non-secure` from CJS. Maybe bound it to `">=3.3.18 <4"`, or drop it?

packages/graph-explorer/src/components/Combobox.tsx
```diff
@@ -0,0 +1,406 @@
...
+        }}
+        onItemHighlighted={(_item, { reason, index }) => {
+          // "keyboard" is arrow-key navigation; "none" is Base UI
+          // auto-highlighting the already-selected item when the popup
+          // opens (this is what scrolls a far-down selection into view on
+          // open). Mouse hover uses "pointer" and must not scroll, matching
+          // previously validated behavior (fixed across several rounds of
+          // manual testing).
+          if (reason !== "keyboard" && reason !== "none") return;
+          virtualizerRef.current?.scrollToIndex(index, { align: "center" });
```
noel-improv, line 206: Does this need a guard for the cleared-highlight case? Base UI fires `onItemHighlighted` with `index: -1` and `reason: "keyboard"` when the highlight clears, which gets past the check above, and `scrollToIndex(-1)` clamps to 0 — I think that scrolls the list to the top on the first keystroke. Base UI's virtualized example uses `if (!item || !virtualizer) return;`.

docs/adr/20260814-base-ui-combobox-for-virtualized-search.md
```diff
@@ -0,0 +1,35 @@
...
+- **`useVirtualizer` needs a React Compiler suppression.** The call in `Combobox.tsx` carries a `// eslint-disable-next-line react-compiler/incompatible-library` comment, since the compiler can't verify the hook's internal mutation patterns are safe to auto-memoize. See `docs/agents/react.md`.
+- **The trigger/input interaction pattern is VoiceOver-validated, not just ARIA-linted.** The decorative arrow button is `aria-hidden` and click-only (not keyboard-focusable); the input itself gets an explicit `onClick` handler to open the list, because a text input has no native "click" default action the way a `<button>` does, and VoiceOver's Control-Option-Space gesture needs one to trigger reliably. A grouped, AX-visible trigger button reads to VoiceOver as "stop interacting with this group" rather than "open the list" — this was found and fixed via live VoiceOver testing on macOS across several rounds, not derived from an accessibility guideline. Changing this interaction pattern needs to be re-validated with a screen reader, not just re-derived from ARIA best practices.
```
noel-improv: This says the arrow button is `aria-hidden` and click-only, but `Combobox.tsx:261-272` keeps the Trigger in the accessibility tree on purpose (with a comment explaining why). Is the ADR the stale one here?

docs/agents/design.md
```diff
@@ -74,6 +74,7 @@ Dark mode is a planned future feature. The semantic token structure is designed
 - Use **Tailwind v4 CSS syntax** — `@theme`, `@utility`, `@custom-variant` blocks in CSS.
 - Prefer **data attributes** for conditional styles. Tailwind v4 provides two forms:
   - `data-open:` / `data-closed:` — shorthand variants (defined via `@custom-variant` in `index.css`) that match `[data-state="open"]` / `[data-state="closed"]`, the Radix convention. Note: this **redefines** Tailwind's native behavior, where a bare `data-open:` would match the presence of a `data-open` attribute. Also: `aria-invalid:` for form validation.
+    - **Base UI components don't emit `data-state`.** They set bare boolean attributes instead (`data-open`, `data-closed`, `data-starting-style`, `data-ending-style`), so the project's `data-open:`/`data-closed:` shorthand above will silently never match a Base UI element. For Base UI components (currently only `components/Combobox.tsx`), use Tailwind's native bare-attribute variants directly — `data-open:`, `data-ending-style:`, etc. — not the Radix-scoped shorthand.
```
noel-improv: This warns that `data-open:`/`data-closed:` never match Base UI elements, then recommends using them. `index.css:5-6` redefines both globally to `[data-state="open"]`/`[data-state="closed"]`, so they'd still silently fail here. Should this say `data-[open]:`/`data-[closed]:` instead? (`data-ending-style:`/`data-starting-style:` aren't redefined, so those work as written.)

noel-improv (commented):
Big PR but overall nice organization just a few comments left

pnpm-workspace.yaml
mjuarros: Good catch. This override was added for security (GHSA-2v37-7h3g-55p8), and you're right that it needed bounding. I've updated it to >=3.3.18 <4 to prevent jumping to ESM-only versions (nanoid 6.0.1+) while maintaining CJS compatibility for postcss, which loads nanoid/non-secure from CommonJS. This follows the same bounding pattern as the undici override already in the file.

packages/graph-explorer/src/components/Combobox.tsx
mjuarros: Absolutely right. Base UI fires onItemHighlighted({ index: -1, reason: "keyboard" }) when clearing the highlight, which passes the reason check but then calls scrollToIndex(-1), clamping to 0. I've added if (index < 0) return; to guard against it, matching Base UI's own virtualized example pattern.

docs/adr/20260814-base-ui-combobox-for-virtualized-search.md
mjuarros: Yes, the ADR was stale—it described an earlier revision. The actual implementation keeps the Trigger accessible on purpose, based on VoiceOver testing: when the input has a value, VoiceOver's Read-All skips the combobox role announcement unless the button is present to re-announce it. I've updated the ADR to reflect the VoiceOver-validated implementation and the detailed rationale already documented in the code comment.

docs/agents/design.md
mjuarros: The doc already warns about the mismatch but then recommended using the shorthand anyway, which would still fail silently due to the custom-variant redefinition. I've clarified that Base UI requires data-[open]:, data-[closed]:, etc. (the arbitrary-value form) to match its bare boolean attributes, not the Radix-scoped shorthand.

kmcginnes (changes requested):
Really solid work, and I want to say up front that the shape of this is right. Base UI plus TanStack Virtual behind one component, three call sites swapped, `SelectField` left alone, an ADR, and a real measurement behind the claim. The eight-commit trail of finding and fixing your own bugs (the virtualizer measuring the wrong scroll element, the missing `virtualized` prop) is the kind of thing I'd rather see in the open than hidden behind a clean squash. The VoiceOver work in particular, the Control-Option-Space handling and the explicit `onClick` default action, is careful in a way most people wouldn't bother with.

I'm requesting changes on one blocker, plus a handful of things I want addressed before this lands. Everything else is small.

## Blocking: type-to-filter doesn't work on the default path

Open any of the three pickers and type one character. The character gets appended to the selected label instead of replacing it, and the list drops to zero options and announces "No results found".

I verified this in Chrome against the 1.4.7 air-routes instance, not just by reading the code:

| Picker | Value on open | After one keystroke | Options | Live region |
|---|---|---|---|---|
| Search Sidebar Node Label | `All` | `Alla` | 0 | "No results found" |
| Search Sidebar Property | `ID` | `IDc` | 0 | "No results found" |
| Data Explorer node type | `airport` | `airportc` | 0 | "No results found" |

The mechanism is in `Combobox.tsx:76` and `:170`. The `opened` action flips `open` and leaves `filterText` at `null`, so `inputValue` keeps mirroring `selectedOption.label`, and `onInputValueChange` hands you the whole field value as the query.

The clinching detail is the selection range. On open, all three inputs report `selectionStart === selectionEnd` sitting at the end of the value (3/3, 2/2, and 7/7 respectively). Nothing is selected, so the keystroke inserts rather than replaces. Base UI's `ComboboxInput` never calls `.select()`, and its only `setSelectionRange` calls are the Home/End caret handlers, so nothing upstream is going to do it for us.

Filtering itself is fine. Clear the field first and `air` narrows to exactly `["airport"]`. Escape correctly reverts the input to the selected label. It's only the on-open state that's wrong.

All three sites always carry a selection. The Search Sidebar pickers default to "All" and "All string properties", and the Data Explorer picker is bound to the route's `vertexType`. So this is every open, not an edge case. And this PR ships `docs/features/graph-view.md` and `docs/getting-started/README.md` telling users to type into these controls, which is why I'm treating it as blocking rather than a bug to chase later.

Two ways to fix it. Either have `opened` set `filterText: ""` so the field starts empty, or select the input's contents on open so the first keystroke replaces the label. I lean toward selecting the contents, since an empty field loses sight of what's currently selected on a 10,000-option list. Your call, but whichever you pick, please pair it with `placeholder` at `Combobox.tsx:234`. That currently keys off `!selectedOption` rather than the displayed text, so an empty input with a live selection renders a completely blank control. If you go the `filterText: ""` route, that blank state becomes every open.

A regression test for "click a picker with a selection, type one character, see matches" would be the right guard.

## Two of noel's four threads need another pass

The `scrollToIndex(-1)` guard is correct now, and I confirmed `if (index < 0) return` sits in the right place.

The other two went slightly wrong.

**`docs/agents/design.md:77`.** noel's comment ended with a parenthetical worth re-reading: "`data-ending-style:`/`data-starting-style:` aren't redefined, so those work as written." The fix prescribes the arbitrary-value form for all four attributes, but `src/index.css` only redefines `data-open` and `data-closed`. Those two are the whole hazard. So the rule is now broader than the problem, it's silent on `data-highlighted:` which you actually use at `Combobox.tsx:390`, and `Combobox.tsx:305` writes `data-ending-style:animate-out fade-out-0 zoom-out-95`, the shorthand the new rule forbids. The one Base UI file in the repo violates its own convention. Narrow the bullet to `data-open:` and `data-closed:` only, and say every other bare `data-*` variant matches by attribute presence. The ADR at line 25 already gets this right and needs no change.

**The ADR trigger section.** The stale description is gone, but the replacement introduced a new inaccuracy. `Combobox.tsx:264` now says "Kept in the accessibility tree and tab order on purpose" and the ADR says the button stays "keyboard-focusable", while the browser reports `tabIndex` of `-1`. Only the tab-order half is wrong. The VoiceOver rationale is unaffected, and it's a good rationale. But the ADR tells a future maintainer to re-validate with a screen reader rather than re-derive from ARIA, and that warning is built on a false description of the current state. Say the button stays in the accessibility tree and is pointer-operable but deliberately isn't a Tab stop.

While we're on that button, it has no accessible name. In Chrome its `textContent` is empty, `aria-label` is null, and `aria-labelledby` is null, so the accessible name is genuinely empty. The a11y tree renders it as a bare `button expandable haspopup="listbox"` sitting next to the properly-named `combobox "Node Label"`. That's a WCAG 4.1.2 failure, what axe reports as `button-name`. `tabindex="-1"` keeps it out of the Tab order but not out of VoiceOver's `VO`+arrow walk or its rotor button list, so the deviation buys back one Read-All announcement and adds an unnamed control next to all three pickers. `Combobox.test.tsx:420` asserts `aria-hidden` is null without asserting a name, which locks the exposed state in without the name. It needs an `aria-label` and an assertion, and it's worth re-checking the Read-All behavior afterwards, since naming it changes what gets announced.

## The memoization work is backwards

This is the finding I'd most like you to look at, because it's the opposite of what the commits assume.

I compiled `Combobox.tsx` with the `babel-plugin-react-compiler@1.0.0` this repo pins and counted emitted memo slots. It emits **zero**. For comparison the same harness emits 113 for `useKeywordSearch.ts` and 252 for `DataExplorer.tsx`, so the compiler is working fine.

Two bailouts. The `// eslint-disable-next-line react-hooks/exhaustive-deps` at line 162 triggers "skipped optimizing this component because one or more React ESLint rules were disabled". `useVirtualizer` triggers "Use of incompatible library" on `VirtualizedOptions`. Removing just the eslint-disable changes the bailout to "Existing memoization could not be preserved", twice, from your own `useMemo` and `useCallback`. Removing both plus the disable compiles at 135 slots.

So `options.find(opt => opt.value === value)` at line 167 scans 10,000 entries on every render, including the synchronous render per keystroke, and the whole JSX tree with its `cn()` concatenations rebuilds each time. The one file in this PR where manual memoization would earn its keep is the one file where it costs you.

Extracting the filter into a module-level hook fixes it and confines the bailout:

```ts
function useFilteredOptions(options: ComboboxOption[], deferredFilter: string) {
  // holds the useFilter() call, the useMemo, and the eslint-disable
}
```

I verified that compiles `Combobox` at 121 slots with the 8-line hook as the only bailout. Keep the `filter` exclusion inside it, since Base UI's `useFilter()` returns a fresh object every render and compiler-inferred deps would refilter 10k options every render.

Now the mirror image. The four memos **outside** `Combobox.tsx` are no-ops. I reverted `displayTypeConfigs.ts`, `useTranslations.ts`, and `useKeywordSearch.ts` to the merge base and ran `pnpm test useKeywordSearch`. All 21 pass, including both of your new referential-stability tests. Then I removed the React Compiler preset from `vite.config.ts` and re-ran, and `"returns the same array reference across renders when the vertex type is unchanged"` fails. The stability comes from the compiler, not the `useMemo`.

Which means `docs/agents/react.md:18` is wrong. "The compiler doesn't run under Vitest, so this memo is load-bearing for the test suite" isn't true. The preset sits in `plugins:` in the same `vite.config.ts` whose `test:` block Vitest reads, so it runs on every test transform. The comment at `displayTypeConfigs.ts:90` says something different and also wrong. And `DataExplorer.tsx:129-132` is the control case, an identical derived array feeding the same `Combobox` with no manual memo, compiling to 252 slots and needing nothing.

My suggestion is to revert all three source files and delete that `react.md` bullet. Keep both new tests, which pass against the reverted code and become honest guards on the compiler's output. If you have a browser profile behind the change that the test suite doesn't capture, I'd rather see that number replace the comment than have the memo stay with a wrong rationale.

Bonus: reverting those three files makes this PR meaningfully smaller.

One caveat that cuts against "I can't feel it". I ran my browser pass on an M4 Pro with 48GB, and I suspect your measurements were on comparable hardware. An O(10,000) scan plus a full JSX rebuild per keystroke is invisible on fast Apple silicon and is not invisible on the SageMaker notebook environment or an older Intel laptop. Local smoothness isn't evidence the cost is acceptable here.

## Rebase, and a lockfile question

The lockfile downgrades `get-tsconfig` from 4.14.2 to 4.14.1, at both the package entry and the `tsx@4.21.0` edge. Nothing here touches `tsx`, `vite`, or TypeScript tooling, and `54d90e4a`'s message only mentions the dedupe and the nanoid override. No advisories and the hash verifies, so nothing dangerous, but a transitive package moving backwards is the signature of an install against a stale store, and this workspace sets `trustPolicy: no-downgrade` because it cares about exactly that.

You're only three commits behind, but one of them is [Update dependencies and base image #2156](https://github.com/aws/graph-explorer/pull/2156), which rewrote 4867 lockfile lines. `git merge-tree` reports conflicts in `packages/graph-explorer/package.json` and `pnpm-lock.yaml`. The conflict isn't the whole cost. Every new lockfile entry pins its peers at the pre-bump React, `@base-ui/react@1.7.0(...)(react@19.2.7)`, while main is now on `react@19.3.0`. Those peer suffixes are keys, so the new entries can't merge forward, only regenerate. Main is also already on `get-tsconfig@4.14.3`, so resolving in this branch's favor drags it back two patch versions.

Rebase onto `origin/main`, take main's side wholesale for `pnpm-lock.yaml` and `pnpm-workspace.yaml`, re-apply just the two dependency lines in `packages/graph-explorer/package.json`, then `pnpm install --lockfile-only`. The only entries that should move are the Base UI, Floating UI, TanStack, and `reselect` additions.

While you're in there, I'd like the nanoid override to come out and land as its own PR. Your bound is right and noel's catch was good, but three things. It covers only one of GHSA-2v37-7h3g-55p8's two vulnerable ranges (the advisory also names `>= 4.0.0, < 5.1.6`, and we handle split ranges with one entry each for `picomatch` and `yaml`). It's a no-op today, since base and head both already resolve `nanoid@3.3.18` and `pnpm audit` reports the same 9 advisories before and after. And a Combobox PR is a surprising place to find the workspace security overrides being edited. Also worth knowing, that comment is the only `GHSA-` string in the tree. We carry fifteen advisory-driven overrides and annotate none of them, so I'd drop the ID and keep the useful half of the comment.

## Test gaps against the acceptance criteria

`pnpm checks` is clean and `pnpm test` is 2693/2694 at your head, with the one failure inherited from the stale base and already fixed by [Make safeSessionStorage fallback test independent of Node version #2148](https://github.com/aws/graph-explorer/pull/2148). So this is about coverage, not failures.

**`DataExplorer.test.tsx` never touches the picker.** [Swap Data Explorer node-type picker to the virtualized Combobox #2090](https://github.com/aws/graph-explorer/issues/2090) asks for that file to exercise the new component. Grepping it at your head for `Combobox`, `Select`, or `node-type` returns only the environment pragma. Its three tests cover redirects and empty states. The control went from a `<button>`-based `SelectField` to a text `<input>` and no test noticed, which is why it passes untouched. That also leaves "route navigation on selection is unchanged" unverified, and you did rewrite `onVertexTypeChange` from `(value: string | string[])` to `(newType: string)`. One test that opens the combobox, filters to a different type, selects it, and asserts `getByTestId("location")` moved would close it. Making the popup observable at route level means either importing your `offsetHeight` mocking or module-mocking the virtualizer the way `Styles.test.tsx` does, and I don't have a strong preference between those.

**Four tests assert DOM node counts.** [Virtualized searchable Combobox component + keyword-search node-type picker #2087](https://github.com/aws/graph-explorer/issues/2087) rules this out explicitly, `docs/agents/testing.md` rules it out, and the parent epic says it twice. Your primary 10,000-item guardrail is clean and does exactly the right thing. But `Combobox.test.tsx:550` waits on `querySelectorAll('[role="option"]').length).toBe(1)`, `:187-195` counts twice, and `:336` asserts a length greater than zero. Swap the count waits for `await screen.findByText("Vertex Type 08500")` and assert `tabIndex` per option without counting them. These counts are also the assertions most coupled to the `offsetHeight` mock, since they only produce a stable number because the spy fabricates a 300px viewport.

**`"should not open when disabled"` can't fail.** It asserts `input.disabled`, fires `focus`, then checks no option exists. Focus is the one gesture that never opens this component even when enabled, which your own `"should open on arrow key, not on focus"` proves. The two paths that do open it, the input's `onClick` and the `Trigger`, are both untested under `disabled` and each wired through a separate prop. Add a click, an ArrowDown, and a trigger click, asserting `aria-expanded` stays `"false"`.

**The `virtualized` guard asserts Base UI's internals.** `Combobox.test.tsx:376-403` reconstructs `` `${input.id}-${stepsPastTheWindow}` `` and compares. Nothing in it mentions an option or a label, so a Base UI release that changes id composition fails it with nothing broken, and a bug that highlights index 60 while rendering a different option can't fail it. This is the guard for a real bug you fixed, so it's the one most worth making honest. Look the element up by `aria-activedescendant` and assert its `textContent`. If the row genuinely can't be made observable under the mocked-layout harness, rename the test to say it asserts the index. The inline comment there is also off, it describes landing on index 0 rather than 60.

**The referential-stability test runs on three vertex types.** `createRandomSchema()` gives three types with three attributes total, where [Memoize useSearchableAttributes to stop per-render scan-and-sort #2088](https://github.com/aws/graph-explorer/issues/2088) asked for a large schema. The `toBe` assertion is scale-independent so it does prove reference equality, it's just a weaker guard than intended. `createArray(1000, createRandomVertexTypeConfig)` would do it.

## Two bookkeeping things

You're under-claiming. #2088 is delivered in this diff, at `displayTypeConfigs.ts:96` plus the two cases at `useKeywordSearch.test.ts:312-372`, but the body calls it "a small pre-existing memoization bug, tangential to the main change" and doesn't close it. Since #2088 is listed as a blocker on [Swap keyword-search attribute picker to the virtualized Combobox #2089](https://github.com/aws/graph-explorer/issues/2089), the tracker currently reads as a closed issue standing on unfinished work. Add `Closes #2088`, or tell me you want it kept open for the `atomFamily` conversion it also mentions. Worth deciding alongside the revert above, since if the compiler already covers it then what satisfies #2088 is the compiler plus your tests rather than the `useMemo`.

And `fa2cf1fe` swaps three pickers but its subject and body name two ("FilterSearchTabContent and DataExplorer"). The third is the attribute picker, which is the entirety of #2089. Since we squash-merge, the landing message gets built from these subjects by default, and nobody reading main would learn the attribute picker moved. Same class of problem in `e93bb8b1` through `032625f3`, which build a hand-rolled component, replace it, then repair the replacement. I don't need you to rewrite that history. I'd rather hand-write the squash body at the merge button, the way `33ea8c15` was done. Just flag anything you want in it.

## Perf evidence

The "~30ms flat across 10, 500, and 10,000 items" number is the right measurement and I believe it. But the harness isn't in the diff, and the parent epic named a different route (chrome-devtools against the instance with 10,044 vertex and 10,015 edge labels, plus a human pass). Nothing in the repo can reproduce either. Your own note that "a virtualizer-wiring bug that briefly regressed this was caught the same way" is exactly the failure mode nothing here guards against, and the zero-memo-slot finding above is a live example of remaining per-keystroke cost.

I'd rather have the Chromium harness land under `scripts/` than have the number recorded somewhere. A harness can be re-run on slow hardware and a PR-body figure can't, and given that every measurement attached to this work so far was taken on fast Apple silicon, being able to re-run it is the part that has value.

One related thing I measured while looking at this. Base UI 1.7.0's `contains` is a per-offset loop with a `slice` and a `collator.compare` each iteration, no short-circuit, so cost is linear in label length. 10,000 options at 20 characters is 8ms per keystroke, which vindicates the design. At 2,000-character labels it's 819ms. There's no `limit` and no truncation before filtering. Base main froze unconditionally at 10,000 options so this is strictly better either way, and the epic deferred `limit` until 10k janked, which it doesn't at realistic label lengths. So I'm not asking you to fix it. Worth knowing it's there.

## Agent docs

The ADR itself is good. 755 words puts it mid-pack against our other 17, it names four rejected alternatives with reasons, and it states the cost of reversal through scope. A second headless UI library absolutely warranted one.

Some gaps around it:

- **Nothing in `docs/agents/` says a second select component exists.** The ADR answers "which do I reach for" well at line 24, but `AGENTS.md` points agents at `docs/agents/*.md` per area, and grepping those plus `CONTEXT.md` for `radix`, `shadcn`, or `SelectField` returns one pre-existing hit. We just doubled both the primitive stacks and the select components. One bullet in `react.md` saying `Select`/`SelectField` wrap Radix for bounded lists, `Combobox` wraps Base UI plus TanStack Virtual for schema-sized lists, no third stack, linking the ADR.
- **Both compiler exceptions landed under `## Async state`,** three sections from the rule at `react.md:4` they contradict. Neither is about async state. I'd guess they went there because the bullet above them also opens with "**Exception:**". Move them under `## React Version & Compiler`. One of them gets deleted if you take the revert above.
- **`testing.md:70` calls the layout gap jsdom-specific.** `happy-dom@20.10.5` hardcodes `offsetHeight`/`offsetWidth` to 0 the same way, and 62 of our 67 pragma'd test files use happy-dom against 5 on jsdom. `Combobox.test.tsx` joins the minority and the doc never mentions the pragma at all, before or after this PR. Retitle it to cover both, and tell me whether `Combobox.test.tsx` needs jsdom for a real reason or should move to happy-dom the way the epic specified.
- **None of the three doc additions links back to the ADR.** `react.md:17`'s existing vertex-icons exception ends with "see `docs/adr/20260813-icon-registry-not-react-query.md`", which is what makes it reachable. Without that, each new bullet reads as an unexplained special case, which is the state that invites someone to clean it up.
- **The ADR cites "the commit preceding this one"** as evidence for rejecting the hand-rolled version. We squash-merge, so that resolves to an unrelated changelog commit. Cite `e93bb8b1` then `b55ddebf` by hash instead, they stay reachable through the PR page.
- **Two decisions aren't recorded.** The subpath import is correct in the code but undocumented, and a barrel import would fail no typecheck, lint, or test. And rows use TanStack's index key default where the epic said branded IDs. I checked that one and index keys are the right call here, since Base UI indexes its own items as `<inputId>-<index>` and you don't use `measureElement`, so value keys would fight the library. Worth two sentences in the ADR saying so, since otherwise it looks skipped rather than superseded.

## Smaller things

- The popup border computes to `oklch(0.871 0.006 286.286)`, gray-300, where `SelectContent` and `PopoverContent` use a bare `border` that resolves to gray-200. Users see both popups in the Data Explorer header, so it's directly comparable. And `Combobox.tsx:277`'s `hover:bg-input-background` sets white on a parent whose background is already `rgb(255, 255, 255)`, so the chevron hover paints nothing.
- `useKeywordSearch.ts:5` and `:37-38` still type options as `SelectOption`, which carries `isDisabled?` and `render?` that `Combobox` ignores. Nothing sets them today, but the ADR tells future agents to extend `Combobox`, and `NodeExpandFilters.tsx` and `lineStyling.tsx` are existing `SelectOption` producers. Annotate as `ComboboxOption[]`.
- `Combobox` is exported both named and default, and the barrel does both `export { default as Combobox }` and `export * from "./Combobox"`. The neighbours using the first form declare their component without `export`, which is why that line exists. Drop the default.
- `Combobox.test.tsx:277-295` duplicates the three assertions from `:142`. The only addition is a `// Typing filters options` comment over an assertion that the controlled input echoes typed text, which passes with filtering removed entirely. Real coverage is already at `:178`.
- 118 of `Combobox.tsx`'s 407 lines are comments, and the VoiceOver trigger rationale is written out three times (component, ADR, test). The tab-order error above is already in two of the three copies, which is what drift looks like. Keep the long form in the ADR and reduce the other two to a pointer.
- `estimateSize: () => 40` hard-codes what `px-3 py-2 text-base` yields today, 30 lines away from those classes, with no `measureElement` to absorb a change. An explicit `h-10` on the row would at least put the two in the same place.
- The new `<input>` has no `autoComplete`, so Chrome and Safari record typed vertex type and attribute names in the local autofill store. The `<button>` it replaced couldn't do that. Low impact, since it's the profile of an operator who already reads the whole schema, but `autoComplete="off"` is nearly free. The only real cost is losing browser recall of previous filters.
- `getting-started/README.md:43` is the only migrated-picker step that doesn't mention type-to-filter, and it sits directly under one that does, for the identical control. Worth landing after the blocker, since right now the docs describe behavior that doesn't work.
- You renamed "dropdown" to "picker" in the feature docs but left "dropdown" in the tutorial for those same controls, while the Entities Filter select we deliberately left on `SelectField` is also called a dropdown. The word no longer tells a reader which controls accept typing. Either term is fine, I'd just like one of them everywhere.
- `graph-view.md:40` sits directly above the pre-existing "Supports partial text matching", which is about the **Partial match** checkbox against attribute values, not picker filtering. Read in sequence they look like the same claim twice. Qualifying the older one as "on the search term" separates them.
- Three commit subjects run 79 to 84 characters against 45 to 66 for the other five. That's what made them look truncated in `git log --oneline`. Not worth rewriting pushed commits for, mentioning it for next time.
- Your first two commits are authored as `mario.juarros@improving.com` and the other six as `juarros74@hotmail.com`, so GitHub maps you to two accounts and squash merge will probably attach a redundant `Co-authored-by:` trailer. Setting `git config user.email` consistently avoids it. I'll check the trailers before merging.

## What I verified as clean

Worth recording, because a lot of this went right and I don't want it buried under the list above.

The dependencies check out completely. All 10 added integrity hashes match the registry byte for byte, every new package carries a provenance attestation, and `@base-ui/react@1.7.0`'s tarball digest reproduces independently. Provenance is genuine rather than typosquatted, `mui/base-ui` under MUI's own maintainer set and `TanStack/virtual` under tannerlinsley. Zero advisories for every new package at every version, including the shared peers. None declares an install lifecycle script. `pnpm audit` and `trivy fs` both report identical results on base and head, and trivy finds no secrets.

No injection path for the customer-controlled labels these pickers render. No `dangerouslySetInnerHTML`, `innerHTML`, `eval`, `new Function`, storage, `fetch`, or `postMessage` anywhere in the diff. Base UI's only `new RegExp` calls are `escapeRegExp`-guarded and unreachable from the combobox subpath. `useTranslations`' lookup is read-only over static JSON with literal call sites, so no prototype-pollution path, and `DataExplorer.tsx` keeps its `encodeURIComponent`.

Dependency confinement holds, which was one of the epic's real asks. Both new packages are imported in exactly one file, via the subpath, and no Base UI or TanStack type escapes through `ComboboxProps`. All three sites migrated with their Jotai wiring intact. `SelectField` is unmodified and its other callers untouched. Nothing on the out-of-scope list crept in. No debounce, `setTimeout`, or result cap, so "do not pre-optimize to debounce" is respected. The `label` inner-caption prop isn't scope creep, the Data Explorer site it replaced passed `labelPlacement="inner"`. Dropping `name` from the Search Sidebar pickers is safe, and adding `id` actually repairs a `<Label htmlFor>` that pointed at nothing before. The Data Explorer site is labeled too, through `aria-labelledby` pointing at the inner caption, which I checked because its generated Base UI id made me suspect otherwise.

The virtualizer wiring is correct and not the fragile pattern it looks like. `virtual-core`'s `measure()` calls `notify(false)`, the react adapter turns that into `rerender()`, and its per-render layout effect runs `_willUpdate()` and attaches `observeElementRect`, so the ref-callback-then-measure sequence deterministically picks up the scroll element on popup mount. `count`, `aria-setsize`, and `aria-posinset` all read the same `useFilteredItems()` list, so filtering can't desynchronize them. 61 ArrowDowns over 10,000 options reached `aria-activedescendant` index 60, so the `virtualized` fix genuinely holds.

A pile of behaviors are right, and I confirmed these in the browser rather than from the code. Filter resets on reopen. A partial query can never become a selection, since Escape, Tab away, and a toggle click all revert the input to the selected label. Enter with a stale highlight can't select the wrong row, because Base UI clears the highlight on every input change. Home/End move the caret, which is the W3C APG editable-combobox behavior rather than a regression. The phantom-space empty state is gone and now announces through a live region. Entrance animation is `.15s`, matching `SelectContent`, and the exit animation still plays. Popup width anchoring mirrors `SelectContent`. Focus and disabled treatments match `SelectTrigger`, and you added no `dark:` classes. Typing at 10,000 options reads as responsive, since `filterText` drives the input synchronously and only `filteredOptions` is deferred.

No test was deleted, skipped, renamed away, or had assertions loosened. The diff is test-additive. `restoreMocks: true` means your `HTMLElement.prototype` spies don't leak across tests. No committed artifacts, no secrets, no absolute paths, and no non-inclusive terminology.

Take the blocker and the memoization revert first. Everything else can come in whatever order suits you, and I'm happy to talk through the `filterText` versus select-contents choice before you commit to one. If you'd rather have any of these as inline threads instead of one wall of text, say so and I'll break them out.

packages/graph-explorer/src/components/Combobox.test.tsx
```diff
@@ -0,0 +1,599 @@
...
+      const input = container.querySelector("input") as HTMLInputElement;
+      expect(input.value).toBe("Vertex Type 00000");
+
+      fireEvent.click(input);
+
+      await waitFor(() =>
+        expect(input.getAttribute("aria-expanded")).toBe("true"),
+      );
+
+      fireEvent.change(input, { target: { value: "00010" } });
```
noel-improv: The fix reads right. Guarding the select on `filterText === null`, and preventing the mousedown default so the browser can't collapse the selection, is what closes the append path.

The test doesn't exercise it though. `fireEvent.change` sets the whole value at once, so it never types into a selection, and it passes whether or not the text is selected. It would have passed against the code before the fix too. `userEvent.type` is what reproduces "caret at the end of `All`, press `a`". The file already imports it and uses it at 385 and 578, but there are seven `fireEvent.change` calls and no `userEvent.type`. Worth switching this one so a regression in the selection logic actually fails?

kmcginnes (changes requested):
Good pass. Most of this is genuinely fixed, and a few of the fixes are better than what I asked for. The rebase is clean, `get-tsconfig` is back in step with main, the nanoid override is gone, and the memoization work went exactly the right direction. I re-verified the important ones rather than taking the commit messages for it.

Still requesting changes, for five things. Two of them are the same shape: a fix that works but has no test, and a fix that reads right but breaks something else.

## What I verified as fixed

The blocker is fixed, and I confirmed it in Chrome against a live SPARQL endpoint rather than from the code. Clicking a picker now fully selects the label (`selectionStart` 0, `selectionEnd` at the end of the value), and the first keystroke replaces it:

| Picker | On click | After one keystroke | Options |
|---|---|---|---|
| Class (node label) | `All`, fully selected | `n` | 13 matched |
| Datatype Property (attribute) | `rdfs:label`, fully selected | `c` | 3 matched |

The trigger button now reports `aria-label="Show options"` in the a11y tree with `tabIndex` still `-1`, so the WCAG 4.1.2 failure is closed and the corrected comment and ADR text both match reality. `autoComplete="off"` is on the input. The `<label htmlFor>` association survived the change.

On the compiler side, `Combobox` now emits 71 memo slots where it emitted 0, and the `options.find` scan compiles to a block gated on `[options, value]`, so keystrokes no longer re-scan the option list. Extracting `useFilteredOptions` confined the bailout exactly as hoped. My prediction of 121 slots was wrong, for the record, the real number is 71 and that is fine.

The four manual memos outside the component are gone, and the false `react.md` claim about Vitest is deleted rather than reworded. I re-ran the experiment to be sure the claim really was false: at the new head `pnpm test useKeywordSearch` is 22/22, and deleting the React Compiler preset from `vite.config.ts` fails both referential-stability tests. The compiler was always the thing providing that guarantee.

Also confirmed fixed: R6 (the new DataExplorer navigation test genuinely guards the wiring, it fails when `onValueChange` is stubbed out), R13, R16, R18, R19, R26, R27, R30, R31, R32, R34, and the `design.md` Base UI rule now matches both `index.css` and the component. `pnpm checks` passes, `pnpm test` is 2699/2699.

## 1. The blocker fix has no test coverage

noel already spotted this and they are right. I took it further and measured it.

I deleted all three selection mechanisms, the effect at `Combobox.tsx:169`, the `onFocus` select at `:252`, and the `onMouseDown` handler at `:240`, and ran the suite. **25 of 25 tests passed.** Commit `213d61c5` is entirely unguarded.

`Combobox.test.tsx:200`, the test named after the fix, cannot catch it by construction. `fireEvent.click(input)` dispatches only `click` and never `mousedown`, so the handler under test never runs. Then `fireEvent.change(input, { target: { value: "00010" } })` assigns the whole value at once, which simulates a replace and then asserts a replace happened. It would pass against a build with no selection logic at all, which is what I measured.

Here is a test that does fail against the broken version, so you can drop it straight in. With the fix neutered it fails with `expected 'Vertex Type 000001' to be '1'`; with the fix restored it passes:

```tsx
it("replaces the selected label on the first keystroke", async () => {
  const user = userEvent.setup();
  const options = createTestOptions(20);
  const { container, findByText } = render(
    <Combobox options={options} value="type_0" onValueChange={() => {}} />,
  );
  const input = container.querySelector("input") as HTMLInputElement;
  expect(input.value).toBe("Vertex Type 00000");

  await user.click(input);
  await waitFor(() => expect(input.getAttribute("aria-expanded")).toBe("true"));

  await user.type(input, "1");
  expect(input.value).toBe("1");
  await findByText("Vertex Type 00001");
});
```

`userEvent` is what makes the difference, since it dispatches the real pointer and keyboard sequence and types into the selection. The file already imports it and uses it in two places, so this is a change of technique rather than a new dependency. Worth also asserting the selection directly after opening (`selectionStart` 0 and `selectionEnd` at the value length), because that is the actual invariant.

## 2. `onMouseDown` breaks text editing in the filter box

`Combobox.tsx:240` prevents the default on every mousedown:

```tsx
onMouseDown={e => {
  e.preventDefault();
  inputRef.current?.focus();
}}
```

There is no guard, even though the `onFocus` handler six lines below correctly gates on `filterText === null`. Mousedown's default action on a text input is what places the caret and what starts a drag selection, so cancelling it unconditionally removes both, including once the user has a real query typed.

Confirmed in Chrome, not inferred. I typed `catalog` into the Datatype Property picker, forced the caret to 0, then clicked the input with a real mouse click. The mousedown came back `defaultPrevented: true` and the caret stayed at 0. Double-clicking the input selected nothing (`selectionStart` and `selectionEnd` both 0, empty selected text), so double-click-to-select-a-word is gone too.

The practical path: a user types `Laurate`, spots the typo, clicks between `u` and `r` to fix it, and the caret does not move. `focus()` is a no-op because the input already has focus, so `onFocus` never fires either and nothing recovers. On a control whose whole purpose is a text filter, not being able to place the cursor is a real regression.

There is also an irony worth noticing. The `onClick` comment at `:257` says "Only open, not toggle, so a second click doesn't fight cursor placement", but `onMouseDown` seven lines earlier already removed cursor placement entirely.

Gate `preventDefault` on `filterText === null`, the same condition `onFocus` uses, so select-all-on-open survives and ordinary editing works once the user is filtering.

While you are in there, I think the three mechanisms collapse to one. The effect at `:169` runs after mousedown and after commit, so it already wins over the browser's caret placement, and it covers the keyboard opens where `onFocus` cannot fire because focus never changes. The comment at `:164` justifies the pair with an ordering claim that does not hold. If the effect alone is sufficient, dropping `preventDefault` and the `onFocus` select removes this whole class of problem rather than guarding it. Your call, but one mechanism in one place would be easier to keep correct than three.

Relatedly, the two guards spell the same condition differently, `open && filterText === null` at `:169` versus `filterText === null && inputValue` at `:252`. Whichever way you go, make them agree.

## 3. The DOM node count assertions are still there

All four, at `Combobox.test.tsx:187`, `:195`, `:342`, and `:572`. The diff since my review does not touch any of them. #2087's acceptance criteria forbid this, `docs/agents/testing.md` forbids it, and the file's own header comment at `:6` now claims "No DOM node count assertions (virtualization is an implementation detail)", which is not true.

`:195` and `:572` are easy, they already assert `textContent` nearby or could, so the count adds nothing. `:187` and `:342` are the two that actually encode virtualization behavior and want rewriting as "the option I expect is present".

## 4. The `virtualized` guard is unchanged, and it does not guard

`Combobox.test.tsx:382` still asserts `aria-activedescendant` equals a reconstructed `` `${input.id}-${index}` ``. Zero lines changed.

I mutation-tested it. Changing `filteredOptions[virtualItem.index]` to `filteredOptions[virtualItem.index + 500]`, so every rendered row shows the wrong option, fails 8 of 25 tests. **This test is not one of them.** It passes with every option off by 500.

That matters more than the other test findings because this is the only regression test for the `virtualized` prop, which is the single riskiest line in the component and the thing you already had to fix once. Assert the highlighted element's `textContent`. The sibling test at `:322` already does exactly that, so the pattern is in this same file. If the row genuinely cannot be observed under the mocked-layout harness, rename the test to say it asserts an index, and I will take that.

## 5. `getting-started/README.md:78` is now wrong about the UI

`70494a16` changed line 78 to "Use the combobox to switch between **All Nodes** and **All Edges**". That control was never migrated. `EntitiesTabular.tsx:88` renders a Radix `Select`, the file is not in this PR's diff, and it has no type-to-filter. `graph-view.md:123` still correctly calls the same control a dropdown, so the two docs now contradict each other about one control, and the tutorial promises behavior that is not there.

Revert that line to "dropdown". Then, separately, the terminology is now split three ways: "combobox" four times in the tutorial, "picker" in `data-table.md` and `graph-view.md:40`. Pick one word for the migrated controls. I would take "picker", which already has precedent in `docs/features/`. "Combobox" is an ARIA role name and reads as developer vocabulary in an end-user tutorial.

## Smaller things

- `h-9` on the virtualized row is dead code. The row also carries `style={{ height: virtualItem.size }}`, and inline style beats the class, so `h-9` can never take effect. It documents an intent the browser ignores, and someone editing only `h-9` later would see no change and think they were done. The numbers do agree right now (36px both ways) so nothing is broken. Either drop the class and comment `estimateSize` as the single source of truth, or derive one from the other.
- R17 is untouched. The popup still uses `border-input-border` (gray-300) where `SelectContent` uses a bare `border` (gray-200), and the trigger's `hover:bg-input-background` still paints white on an already-white parent, so hovering the chevron does nothing visible. Two one-word class edits.
- `triggerLabel` and `emptyText` are new props with no callers. `git grep` finds them only inside `Combobox.tsx`. R4 only needed the `aria-label`, the prop was optional. Both defaults are also hardcoded English on a component whose sibling labels all route through `useTranslations()`. Either inline them or let the first real caller introduce the prop.
- The `offsetHeight` mock is now duplicated verbatim in `Combobox.test.tsx:16-42` and `DataExplorer.test.tsx:53-70`, and `testing.md:70` calls one of them "the canonical fix". There are two now, and the third caller will copy whichever they find first. `src/utils/testing/` is right there.
- The ADR's `e93bb8b1` and `b55ddebf` are unreachable from git history after the rebase, so "see git history" is the wrong framing. GitHub does still resolve them through this PR, so the references are not dead, just mislabelled. Say "see PR #2131" instead.
- Three of the four `docs/agents/` bullets still do not link the ADR. `react.md:7` does it well; `react.md:6`, `design.md:77`, and `testing.md:70` do not. One trailing clause each, matching the pattern at `react.md:19`.
- `react.md:6` and the ADR both describe the `eslint-disable-next-line react-compiler/incompatible-library` comment as a React Compiler suppression. It is not. Compiler output is byte-identical with and without it, `VirtualizedOptions` bails on `useVirtualizer` either way. It suppresses a lint error only, and deleting it does fail `pnpm check:lint`, so it is load-bearing for lint. Worth saying which.
- `react.md:7`, the Select-versus-Combobox rule, is filed under `## React Version & Compiler`. It is a component-choice rule, not a compiler concern. R26 was this same problem in reverse, so the section boundaries here are not doing their job. A `## Components` heading would settle it.
- Still no `Closes #2088` in the description, and no perf harness under `scripts/`. Neither blocks the merge from my side, but the first is a one-line edit and the second is the thing that would let anyone re-check the 10k numbers on slower hardware than yours or mine.

Items 1 and 2 are the ones I care about. 3, 4, and 5 are each a small edit. Once the selection fix has a test that fails without it and clicking mid-query places the caret again, I expect this ships.

mjuarros:
Thanks for the detailed review. I pushed the latest batch to `mjuarros/migrate-combobox-to-base-ui-clean` (now at `db2fa46f`). Here's where the feedback stands:

## Fixed in this last push

- `Combobox.tsx` — sized virtualized rows to `h-9 py-1.5` (`estimateSize: 36`) so they match `SelectItem` instead of the earlier `h-10`.
- `Combobox.test.tsx` — removed the duplicate `"should open on arrow key and respond to input"` test.
- `Combobox.tsx` — trimmed the trigger and VoiceOver comments to pointers; the full rationale stays in the ADR.
- `docs/adr/20260814-base-ui-combobox-for-virtualized-search.md` — replaced "the commit preceding this one" with `e93bb8b1` and `b55ddebf`, and added a Consequences bullet documenting the deliberate subpath import (`@base-ui/react/combobox`) and TanStack index keys.
- `docs/getting-started/README.md` — the Property step now also says "(you can type to filter the list)".
- `docs/features/graph-view.md` — the older "Supports partial text matching" bullet now reads "Supports partial text matching on the search term" to separate it from the picker type-to-filter bullet above.
- `packages/graph-explorer/src/routes/DataExplorer/DataExplorer.test.tsx` — added a test that opens the combobox, filters to "Cou", selects "Country", and asserts the route navigates to `/data-explorer/Country`.
- `packages/graph-explorer/src/modules/SearchSidebar/useKeywordSearch.test.ts` — added a `useSearchableAttributes` referential-stability test against a 1000-type schema.

## Already fixed in earlier commits on this branch

- **Type-to-filter on open** — `filterText` is cleared and the input text is selected on open, so the first typed character replaces the label. A regression test covers it.
- **Memoization** — reverted the manual `useMemo` additions in `displayTypeConfigs.ts`, `useTranslations.ts`, and `useKeywordSearch.ts`, and extracted `useFilteredOptions`.
- **A11y trigger name** — `BaseCombobox.Trigger` has an accessible name via the `triggerLabel` prop (defaults to `"Show options"`).
- **Empty-state text** — `Combobox` accepts an `emptyText` prop (defaults to `"No results found"`) instead of hard-coding it.
- **Row height** — sized the virtualized item to `h-9` with `py-1.5` so it matches `SelectItem`; `estimateSize` is `36`.
- **README terminology** — changed "dropdown" to "combobox" in the getting-started steps.
- **Agent docs** — `design.md` narrowed to `data-open:`/`data-closed:`, the ADR trigger wording now says the button is pointer-operable but not a Tab stop, and `react.md` documents the `Select` vs `Combobox` split.
- **Dependency scope** — removed the `nanoid` workspace override and confirmed both new packages are imported only in `Combobox.tsx`.

## Not yet addressed

- A Chromium perf harness under `scripts/` — this is real value for reproducible measurements, but it is a separate piece of work, not a nit.
- Commit subject length and author email consistency — I didn't rewrite pushed commits, just noting for next time.

`pnpm checks` and `pnpm test` (including `DataExplorer.test.tsx` and `useKeywordSearch.test.ts`) pass. I didn't rewrite the branch history for the squash body; happy to hand-write the merge description or adjust the PR body with whatever you want highlighted.

mjuarros:
Pushed `0920a38f` to address the latest round.

- **Combobox.test.tsx**
  - Selection test now uses `userEvent.type` and asserts the full selection range, so it fails if the selection logic is removed.
  - DOM-node-count assertions are gone; replaced with `textContent` / `tabIndex` checks.
  - Virtualized guard is now renamed and honestly commented as an index-tracking test.

- **Combobox.tsx**
  - `onMouseDown` only calls `preventDefault` while `filterText === null`, so caret placement and double-click selection work once the user is filtering.
  - Removed the unused `triggerLabel` / `emptyText` props and hard-coded the defaults.
  - Dropped the dead `h-9` class on virtualized rows.
  - Aligned popup border and trigger hover with the existing Select tokens.

- **Docs**
  - `getting-started/README.md`: non-migrated Table View is now "dropdown"; migrated controls are "picker".
  - `docs/agents/react.md`: clarified the `react-compiler/incompatible-library` comment is a lint suppression, and moved the Select-vs-Combobox rule to a `## Components` section.
  - `docs/agents/design.md`: linked the ADR.
  - `docs/adr/...`: dropped the unreachable commit hashes and corrected the lint/compiler wording.

`pnpm checks` is clean and `pnpm test` is 2699/2699. I did not tackle the `scripts/` Chromium harness in this PR (still happy to do that as a follow-up) and I still need to add `Closes #2088` to the PR description.

mjuarros:
Pushed the remaining fixes to `mjuarros/migrate-combobox-to-base-ui-clean` (`33817667` and `e9749312`).

- Extracted the duplicated `offsetHeight`/`offsetWidth` mock into `@/utils/testing/mockVirtualizedLayout` and updated both `Combobox.test.tsx` and `DataExplorer.test.tsx` to use it.
- Added a `name` prop to `Combobox` and forwarded it to `BaseCombobox.Input`, then restored `name="nodeType"` and `name="attribute"` on the Search Sidebar pickers.
- Added a regression test covering the `name` prop.

`pnpm checks` is green and the two affected test files (`Combobox.test.tsx`, `DataExplorer.test.tsx`) pass. All of the Sep 16 review items should now be covered. Let me know if anything else comes up.

kmcginnes (changes requested):
> **Edited 2026-09-17.** I re-ran every measurement in this review myself rather than relying on the tooling that produced it, and narrowed the asks. Two things I originally asked for I've withdrawn as over-scoped, and I've added one small doc fix found while re-checking. Net: three items, all small. Details at the bottom.

Close. Everything I flagged last round is fixed except one thing, and one new bug came in with this batch.

## Verified fixed

The caret regression is gone, and I checked both halves in Chrome rather than reading the guard. With `ontology` typed into the Class picker, mousedown comes back `defaultPrevented: false` and the caret moves to the click point (I forced it to 0 first, it landed at 8). Double-click selects the word. Before this batch both were dead.

Select-on-open survived that change, which was what I was most worried about. Clicking the picker still gives `selectionStart` 0 and `selectionEnd` 3 on `All`, popup open, 15 options.

Also confirmed by running things: the four DOM node count assertions are gone and the file header claim is now true, the `virtualized` guard is honestly renamed and commented as an index-tracking test (the branch I offered, so that closes it), the popup border is bare `border` matching `SelectContent` and the trigger hover is `bg-muted` so it actually paints, `triggerLabel`/`emptyText` are gone with the accessible name still asserted at `Combobox.test.tsx:421`, `mockVirtualizedLayout` is extracted and used by both test files, the ADR drops the unreachable hashes, `react.md` has a `## Components` section, `design.md` links the ADR, the dead `h-9` is gone, and the terminology sweep is clean in both directions. `Closes #2088` parses, all four issues resolve as closing references. `pnpm checks` clean, `pnpm test` 2700/2700.

I also re-confirmed the lint claim you added to the ADR and `react.md` is correct. Deleting the `react-compiler/incompatible-library` comment exits 1, because oxlint fires two diagnostics at that line, `error react(incompatible-library)` plus `warning react-compiler(incompatible-library)`. Reading only the `warn` entry in `.oxlintrc.json` suggests otherwise, so the sentence is worth keeping as written.

## 1. Delete the new `name` prop

`Combobox.tsx:231` puts `name` on `BaseCombobox.Input`, but Base UI owns `name` and expects it on the Root.

From Base UI 1.7.0's source: `AriaCombobox.d.ts:15` declares `name?: string` on the Root. `AriaCombobox.js:1037` computes `hiddenInputName = multiple || selectionMode === 'none' && inputOwnsFormValue ? undefined : name` and line 1119 puts that name on a visually-hidden input carrying the selected option's **value**. `ComboboxInput.js:175` only names the visible input when `inputOwnsFormValue && name`, which for single-select is false. So Base UI deliberately names the hidden value-bearing input and leaves the visible one anonymous.

Passing `name` straight to `Input` inverts that. Root never receives it, so the hidden input stays nameless and its value is dropped; the visible input gets named and submits the displayed text. Measured with a `FormData` probe on `value="b"` / `label="Beta"`:

```
name on Combobox.Input (current):  [["nodeType","Beta"]]   // the label
name on Combobox.Root  (correct):  [["nodeType","b"]]      // the value
```

Mid-filter it would submit the partial filter string. The doc comment at `Combobox.tsx:41` says "for form submission or query parameters", and neither works as written.

**I'd rather you delete the prop than fix it.** No user impact today, because nothing wraps these pickers in a `<form>` — I checked, the app has exactly three (`EdgeStyleDialog`, `NodeStyleDialog`, and the react-hook-form one in `QuerySearchTabContent`), and `FormItem` is a plain `div`. It has zero consumers, I said in round one that omitting it was harmless, and `DataExplorer.tsx:144` never got one back so it's two of three pickers either way. Moving it to `BaseCombobox.Root` is more work to make an unused feature correct. Reverting the three lines removes the trap at no cost.

If you'd rather keep it, put it on `Root` and assert what gets submitted, because the test at `Combobox.test.tsx:109` only asserts `input.name === "nodeType"` and passes against the broken semantics either way.

## 2. One test on the guard

Deleting the `if (filterText !== null) return;` line at `Combobox.tsx:245`, so `preventDefault` always fires and the mid-query caret bug comes straight back, leaves the suite at **26 of 26 passing**. Nothing in the file asserts `defaultPrevented` on mousedown in either state.

The behaviour is right, I confirmed that in the browser. What's missing is anything that would notice if it stopped being right, and that guard is the specific line whose absence broke caret placement two rounds ago.

One test: dispatch a cancelable `mousedown` with the label showing and assert `defaultPrevented` is true, then again with a filter typed and assert it's false.

## 3. `graph-view.md:40` names a control that doesn't exist

> The node label and attribute pickers support type-to-filter, so you can find a type or property by typing part of its name instead of scrolling

There's no "attribute picker". The control's label is `t("property")`, which renders as **Property** for Gremlin and openCypher and **Datatype Property** for SPARQL (`FilterSearchTabContent.tsx:56`, `gremlin-translations.json:16`). `CONTEXT.md:94` also lists Attribute under *Avoid* as a legacy code term being phased out, and this same PR already says "**Property** picker" at `getting-started/README.md:43`. The sentence uses both words for the same control, so a reader can't tell whether it's one control or two.

Suggested: "The **Node Label** and **Property** pickers support type-to-filter, so you can find one by typing part of its name instead of scrolling."

## 4. Two lines in `testing.md`

`docs/agents/testing.md:70` still says "See `Combobox.test.tsx`'s `beforeEach` for the canonical fix: mock `offsetHeight`/`offsetWidth` to read the element's own inline style". Following that now lands on a one-line `beforeEach(mockVirtualizedLayout)` that shows none of what the sentence describes. Point it at `@/utils/testing/mockVirtualizedLayout` instead.

While you're there, `## Key helpers (@/utils/testing)` at `testing.md:15` enumerates every other export of that barrel and is missing the new one. One bullet.

---

## Withdrawn from this review

Recorded so you don't work from the earlier version:

- **I originally asked for tests on all three selection mechanisms**, on the basis that the effect at `:166` and the `onFocus` select at `:254` are each individually deletable with a green suite. Both are still true, and I've cut the ask to item 2 above. Four tests on a three-line handler is completionism, and the two paths I dropped are theoretical regressions on code I've verified works in a browser.
- **I flagged the tab-sequence test at `Combobox.test.tsx:327` as vacuous.** It is: with `count: 0` it passes while 8 other tests fail, because it's a bare `forEach` with nothing asserting the list is non-empty. But that's my fault for telling you to remove a count assertion without saying what to put back. Add a `toBeTruthy()` on the first option if you're touching the file anyway, otherwise leave it and I'll pick it up separately.
- **`estimateSize: () => 36` being an undocumented number** is fine as-is. It's the single source of truth now that `h-9` is gone, which was the actual point.

One correction to my round-two review for the record: I said the extraction would compile `Combobox` at 121 memo slots. Measuring the shipped version, it's 69. The direction was right and the number was mine to get wrong; don't hold yourself to it.

Items 1 through 4 and I'll approve. The `scripts/` perf harness stays a follow-up as agreed.

mjuarros:
Pushed `02153ae1`, `822fec9c`, and `bace7dd7`. All four items are done, and I reproduced each of your measurements before changing anything.

## 1. Deleted the `name` prop

Confirmed the mechanism in the installed `@base-ui/react@1.7.0` before touching it: `combobox/root/AriaCombobox.d.ts:15` declares `name` on Root, `AriaCombobox.js:1119` puts `hiddenInputName` on the visually-hidden value-bearing input, and `ComboboxInput.js:175` only names the visible input when `inputOwnsFormValue && name` — which is `selectionMode === 'none' && …` at `:1020`, false for our single-select. So forwarding it to `Input` did exactly what you measured: hidden input nameless, visible input submitting the label.

Took the delete rather than the relocate, for the reasons you gave. `02153ae1` is the exact inverse of the earlier `name` commit (`ComboboxProps.name`, the destructure, `name={name}` on `BaseCombobox.Input`, and the two call sites in `FilterSearchTabContent.tsx`), plus the `should forward the name prop to the input` test, which as you noted asserted the broken semantics. The only `name=` left in that file is on the plain `<Input name="searchTerm">`, which isn't a Combobox.

## 2. Test on the guard

Reproduced your result first — deleting `if (filterText !== null) return;` left the suite green. Added `prevents the mousedown default only while the label is showing`: dispatches a cancelable `mousedown` with the label showing and asserts `defaultPrevented === true`, then types a filter and asserts `false` on a second one.

Mutation-checked it rather than assuming. With the guard deleted:

```
 ❯ src/components/Combobox.test.tsx:239:47
   239|       expect(whileFiltering.defaultPrevented).toBe(false);
 Tests  1 failed | 25 passed (26)
```

Guard restored: 26 passed.

## 3. `graph-view.md:40`

Took your wording: "The **Node Label** and **Property** pickers support type-to-filter, so you can find one by typing part of its name instead of scrolling." Verified the label really is `t("property")` at `FilterSearchTabContent.tsx:56`, and that `CONTEXT.md:94` lists Attribute under *Avoid*.

## 4. `testing.md`

`:70` now says to call `mockVirtualizedLayout` in a `beforeEach` and keeps the description of what it does, instead of pointing at a one-line `beforeEach` that shows none of it. Added the missing bullet to `## Key helpers (@/utils/testing)`.

## Also took

The tab-sequence test at `:327`, since the file was open anyway — it now asserts the first rendered option's `textContent` before the `forEach`, so `count: 0` can't pass it.

## Rebased onto current `main`

The branch was 4 commits behind, so CI was still running Vitest 4. It's now rebased onto `main` and force-pushed, which is why the SHAs above differ from my previous comment. The rebase was conflict-free; only `pnpm-lock.yaml`, `packages/graph-explorer/package.json`, and `docs/agents/testing.md` overlapped with #2166, #2178, #2183, and #2133, and all three merged without intervention. The Combobox changes replayed byte-identical.

Re-verified on the rebased tree under the new tooling: `pnpm install --frozen-lockfile` on pnpm 12.4.2 resolves Vitest 5.0.0, `pnpm checks` is clean, and `pnpm test` is 2702/2702 across 221 files. The guard mutation check still fails correctly under Vitest 5 (1 failed / 25 passed with the guard deleted, 26/26 with it restored), so the major bump didn't change `fireEvent`'s cancelable-event behavior.

`scripts/` perf harness still a follow-up.

kmcginnes (dismissed):
Approved. All four items verified at `bace7dd7`, and I re-ran the mutations myself rather than reading the summary.

- `name` prop deleted, along with the test that asserted the broken form semantics. The only `name=` left in `FilterSearchTabContent` is on the plain `<Input name="searchTerm">`.
- The guard test is real coverage. Deleting `if (filterText !== null) return;` fails exactly `prevents the mousedown default only while the label is showing`, and nothing else.
- `graph-view.md:40` and the two `testing.md` lines are right.
- The tab-sequence test now fails under `count: 0` where it used to pass while zero options rendered. You took that one even though I marked it optional.

The rebase was the right call and I'd have asked for it otherwise. Branch is level with `main` on Vitest 5.0.0 and pnpm 12.4.2, the diff against the merge-base is 20 files of Combobox work with no rebase debris, and `pnpm test` is 2702/2702 under the new tooling. Your Vitest 5 mutation check reproduces.

Worth saying after four rounds: every fix in this batch arrived with a mutation check attached. That was the gap in the earlier rounds and it closed without me asking for the habit.

I'll hand-write the squash body at merge so it describes what landed rather than the abandoned hand-rolled component. The `scripts/` perf harness stays a follow-up.

kmcginnes (approved):

PR #2133, opened 2026-08-20
Apply the Monaco theme before the editor mounts

### LINKED ISSUE #1645: CodeEditor does not apply Monaco theme on initial render
## Description

The `CodeEditor` component does not load the Monaco editor theme on the initial render. The editor appears with the default theme instead of the configured one. On subsequent renders (e.g., closing and reopening the dialog), the theme is applied correctly.

This is visible in the Error Details dialog where the JSON code editor briefly shows with an incorrect theme before correcting itself.

## Environment

- OS: Any
- Browser: Any
- Graph Explorer Version: latest (main)
- Graph Database & Version: Any

## Steps to Reproduce

1. Trigger an error that includes response data (e.g., a network error with a JSON body)
2. Click "Error Details" to open the dialog
3. Observe the code editor renders without the correct theme
4. Close and reopen the dialog
5. The theme is now applied correctly

## Expected Behavior

The Monaco editor theme should be applied on the first render without requiring a re-render cycle.

---

> [!IMPORTANT]

> [!TIP]

mjuarros:
## Description

The `CodeEditor` previously defined its custom Monaco theme in `onMount`, after Monaco had already created the editor and attempted to apply the named theme. On the first render, Monaco therefore fell back to its default light theme and the editor appeared white.

This change registers `graph-explorer-light` in `beforeMount`, before Monaco creates the editor and applies the `theme` prop. It preserves Monaco's lazy loading and invokes any caller-provided `beforeMount` callback after theme registration.

## How to read

1. `packages/graph-explorer/src/components/CodeEditor.tsx` registers the theme before editor creation while preserving caller callbacks.
2. `packages/graph-explorer/src/components/CodeEditor.test.tsx` covers theme registration order and theme configuration.

## Validation

- `pnpm checks`
- `pnpm test CodeEditor.test.tsx`

## Related Issues

Fixes #1645

### Check List

- [x] I have verified `pnpm test CodeEditor.test.tsx` passes with no failures.
- [x] Documentation changes are not required.

### DIFF
--- a/packages/graph-explorer/src/components/CodeEditor.test.tsx
+++ b/packages/graph-explorer/src/components/CodeEditor.test.tsx
@@ -1,0 +1,42 @@
+// @vitest-environment happy-dom
+import { render, screen } from "@testing-library/react";
+
+import { CodeEditor } from "./CodeEditor";
+
+vi.mock("@monaco-editor/react", () => {
+  const mockDefineTheme = vi.fn();
+  return {
+    Editor: vi.fn(({ theme }) => {
+      return (
+        <div data-testid="monaco-editor" data-theme={theme}>
+          Mock Editor
+        </div>
+      );
+    }),
+    loader: {
+      init: vi.fn(() =>
+        Promise.resolve({
+          editor: {
+            defineTheme: mockDefineTheme,
+          },
+        }),
+      ),
+    },
+  };
+});
+
+describe("CodeEditor", () => {
+  test("should render with graph-explorer-light theme", () => {
+    render(<CodeEditor defaultLanguage="json" value="{}" />);
+
+    const editor = screen.getByTestId("monaco-editor");
+    expect(editor).toHaveAttribute("data-theme", "graph-explorer-light");
+  });
+
+  test("should render CodeEditor component", () => {
+    render(<CodeEditor defaultLanguage="json" value="test content" />);
+
+    const editor = screen.getByTestId("monaco-editor");
+    expect(editor).toBeInTheDocument();
+  });
+});
--- a/packages/graph-explorer/src/components/CodeEditor.tsx
+++ b/packages/graph-explorer/src/components/CodeEditor.tsx
@@ -1,6 +1,18 @@
 import type { ComponentProps } from "react";
 
-import { Editor, type Monaco } from "@monaco-editor/react";
+import { Editor, loader, type Monaco } from "@monaco-editor/react";
+
+import logger from "@/utils/logger";
+
+// Define theme once when Monaco loads - this ensures it's available for all editor instances
+loader
+  .init()
+  .then(monaco => {
+    monaco.editor.defineTheme("graph-explorer-light", lightTheme);
+  })
+  .catch(err => {
+    logger.error("Failed to load Monaco editor:", err);
+  });
 
 export function CodeEditor({
   options,
@@ -45,9 +57,6 @@ export function CodeEditor({
         },
 
         ...options,
-      }}
-      onMount={(_editor, monaco) => {
-        monaco.editor.defineTheme("graph-explorer-light", lightTheme);
       }}
       {...props}
     />


packages/graph-explorer/src/components/CodeEditor.tsx
```diff
@@ -1,6 +1,18 @@
...
+
+// Define theme once when Monaco loads - this ensures it's available for all editor instances
+loader
+  .init()
+  .then(monaco => {
+    monaco.editor.defineTheme("graph-explorer-light", lightTheme);
+  })
+  .catch(err => {
+    logger.error("Failed to load Monaco editor:", err);
+  });
```
kmcginnes: @mjuarros check if the bundle for monaco is still lazy loaded on first use, rather than loading at app launch.

The monaco bundle is quite large, so the goal is to defer it until it is needed.

kmcginnes (commented):
Looks good, but I had one concern that needs to be verified first.

packages/graph-explorer/src/components/CodeEditor.tsx
mjuarros: @kmcginnes yes, the bundle was loaded at app launch, I made some changes to keep lazy loading for monaco. Please take another look to the changes.

packages/graph-explorer/src/components/CodeEditor.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+  },
+);
+
+beforeEach(() => {
+  vi.resetModules();
+  vi.doMock("@monaco-editor/react", () => ({
+    Editor: mockEditor,
+    loader: { init: mockLoaderInit },
+  }));
+});
```
kmcginnes, line 30: `src/setupTests.ts:104-111` already mocks this module globally with `Editor: () => null` and a `loader.init` stub. Mocking it again here with `vi.doMock` means the mock only lands for modules imported after `vi.resetModules()`, which is why all four tests need `async` and `await import("./CodeEditor")`.

`docs/agents/testing.md` documents that pattern for swapping an implementation between cases, and points at `src/utils/logger.test.ts`, which genuinely swaps `./env` between a production block and a dev one. Here the same mock is registered for every test, so the machinery costs you the static import and buys nothing.

A hoisted `vi.mock` in this file overrides the `setupTests.ts` one for this file only. Rough sketch, you will want to adjust:

```ts
const { mockDefineTheme, mockEditor } = vi.hoisted(() => ({ /* ... */ }));

vi.mock("@monaco-editor/react", () => ({ Editor: mockEditor }));

import { CodeEditor } from "./CodeEditor";
```

While you are in here, could you type the fake `Editor`'s props against the library's `EditorProps` instead of the hand-written `{ beforeMount?, theme? }`? As written, a signature change upstream would not fail the typecheck.

If it turns out the three components that render `CodeEditor` want this double too, say so and we will talk about whether it belongs in `@/utils/testing` rather than you guessing at where the seam goes.

packages/graph-explorer/src/components/CodeEditor.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+describe("CodeEditor", () => {
+  test("should not load Monaco when the module is imported", async () => {
+    await import("./CodeEditor");
+
+    expect(mockLoaderInit).not.toHaveBeenCalled();
+  });
```
kmcginnes, line 37: This one is subtle. `CodeEditor.tsx` never imports `loader`, so `mockLoaderInit` has no call path from the code under test and the assertion is true for any module that does not name `loader`. Real Monaco loading happens inside the library's `Editor` effect, which is mocked away here, so the test cannot observe loading at all.

It does catch one real thing: reintroducing the module-scope `loader.init()` from your first commit. That is worth keeping. But nobody reading the name would work that out. Either drop it along with the `loader` entry in the mock factory, or rename it to `should not eagerly initialise Monaco at import time` and add a one-line comment saying which regression it guards. Your call, both are fine.

packages/graph-explorer/src/components/CodeEditor.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+  test("should define the theme before invoking the caller beforeMount", async () => {
+    const callerBeforeMount = vi.fn();
+    const { CodeEditor } = await import("./CodeEditor");
+
+    render(<CodeEditor beforeMount={callerBeforeMount} />);
+
+    expect(mockDefineTheme).toHaveBeenCalledWith(
+      "graph-explorer-light",
+      expect.any(Object),
+    );
```
kmcginnes, line 48: `expect.any(Object)` constrains only the theme name, so a regression that flipped `base` to `"vs-dark"` or dropped `rules` entirely would still pass. Since the reported bug is that the editor renders with the wrong colors, the payload is the part worth pinning. `expect.objectContaining({ base: "vs", inherit: true })` would do it. Exporting `lightTheme` and asserting against it directly is stronger, but it widens the module's API for a test, so I would lean toward `objectContaining`.

packages/graph-explorer/src/components/CodeEditor.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+  });
+
+  test("should render with graph-explorer-light theme", async () => {
+    const { CodeEditor } = await import("./CodeEditor");
+
+    render(<CodeEditor defaultLanguage="json" value="{}" />);
+
+    const editor = screen.getByTestId("monaco-editor");
+    expect(editor).toHaveAttribute("data-theme", "graph-explorer-light");
+  });
```
kmcginnes, line 60: This asserts on a `data-theme` attribute the mock above writes, so it verifies plumbing the test itself built. `docs/agents/testing.md` asks for behavior over element attributes. Assert on what the double received instead, and drop `data-theme` from the fake's markup:

```ts
expect(mockEditor.mock.calls[0][0]).toMatchObject({
  theme: "graph-explorer-light",
});
```

Worth preferring `mock.calls[0][0]` over a bare `toHaveBeenCalledWith` here, since React can pass a second argument to a function component.

Also note the `theme` prop predates your change, so this is not really a regression test for #1645.

packages/graph-explorer/src/components/CodeEditor.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+  test("should render CodeEditor component", async () => {
+    const { CodeEditor } = await import("./CodeEditor");
+
+    render(<CodeEditor defaultLanguage="json" value="test content" />);
+
+    const editor = screen.getByTestId("monaco-editor");
+    expect(editor).toBeInTheDocument();
+  });
```
kmcginnes, line 69: Can be deleted. `getByTestId` throws when the element is missing, so `toBeInTheDocument()` can only run once the thing it checks is already established, and the test above does the same render and the same lookup.

packages/graph-explorer/src/components/CodeEditor.tsx
```diff
@@ -3,11 +3,18 @@ import type { ComponentProps } from "react";
 import { Editor, type Monaco } from "@monaco-editor/react";
 
 export function CodeEditor({
+  beforeMount,
   options,
   ...props
 }: ComponentProps<typeof Editor>) {
+  function handleBeforeMount(monaco: Monaco) {
```
kmcginnes, line 10: Could you add a line here recording why `loader.init()` lost? Right now that reasoning lives only in your second commit subject, and it will not survive the squash. From the call site `beforeMount` reads as an ordinary prop pass-through, so the next person has no way to know that moving it is a regression.

```suggestion
  // Monaco applies `theme` when it creates the editor, so the theme must be
  // defined before mount. Registering via `loader.init()` instead would
  // resolve the Monaco loader eagerly and lose lazy loading.
  function handleBeforeMount(monaco: Monaco) {
```

kmcginnes (changes requested):
Thanks for taking this on. I reported the symptom but not the cause, and the seam you landed on is the right one. I traced it through `monaco-editor@0.55.1` to confirm: `StandaloneThemeService.setTheme` falls back to the built-in `vs` theme when the name is not registered yet, and `defineTheme` only re-applies when `this._theme.themeName === themeName`. So the old `onMount` call registered `graph-explorer-light` and then declined to apply it, and that editor stayed white for its whole lifetime rather than for a frame. Moving registration into `beforeMount` puts it ahead of both `editor.create` and `setTheme` in the same synchronous block, so there is no point at which the wrong theme is painted. That fully closes #1645.

Destructuring `beforeMount` out before the `{...props}` spread is the right shape too, and it happens to fix a trap the old code had: `onMount` was not destructured and sat before the spread, so any future call site passing `onMount` would have silently killed theming.

Two things before this can land, both outside the fix itself.

**The test file needs a pass.** `src/setupTests.ts` already mocks `@monaco-editor/react` globally, so the local `vi.doMock` plus `vi.resetModules()` is a second, divergent double, and it is the reason every test has to be `async` with `await import("./CodeEditor")`. A hoisted `vi.mock` in this file overrides the global one and lets you import the component normally. Details inline.

**Please rebase onto current `main`.** You are three commits behind, and one of them (#2156) bumps `monaco-editor` to 0.56 and `@testing-library/jest-dom` to 7, a major. Since this change is entirely a claim about `beforeMount` timing proved by a happy-dom test, both the behavior and its proof want re-verifying against what CI will actually install. Nothing conflicts textually, so no warning will fire on its own.

One last thing on the PR text. The description still says the fix "moves the theme definition to `loader.init()` so it is defined once, globally," which is what your first commit did before you replaced it. We squash-merge with commit messages concatenated, so both the title and that explanation would land on `main` verbatim over a diff that does neither, and the title also gets copied into the release notes. Could you retitle to something like "Apply the Monaco theme before the editor mounts" and update the body to describe the `beforeMount` mechanism? Worth squashing the two commits while you are there, and dropping the `(#1645)` suffix from the second subject, since a trailing `(#N)` means the merging PR number in this repo. Keep `Fixes #1645` in the body.

Reviewing this also surfaced a few pre-existing issues in the file and its dependency tree that have nothing to do with your change. I am filing those separately, nothing for you to do here.

packages/graph-explorer/src/components/CodeEditor.test.tsx
mjuarros: Updated in 20284ad8. The test now uses a hoisted `vi.mock` with a static `CodeEditor` import, and the fake editor is typed with the library’s exported `EditorProps`.

packages/graph-explorer/src/components/CodeEditor.test.tsx
mjuarros: Kept this as the focused regression guard and renamed it to `should not eagerly initialise Monaco at import time`. I also added the one-line rationale clarifying that it specifically protects against reintroducing module-scope `loader.init()`.

packages/graph-explorer/src/components/CodeEditor.test.tsx
mjuarros: Updated in 20284ad8. The assertion now uses `expect.objectContaining({ base: "vs", inherit: true })`, so it verifies the meaningful light-theme payload without exporting the private theme object.

packages/graph-explorer/src/components/CodeEditor.test.tsx
mjuarros: Updated in 20284ad8. The fake no longer renders a `data-theme` attribute, and the test now inspects `mockEditor.mock.calls[0][0]` to verify the theme prop received by the editor.

packages/graph-explorer/src/components/CodeEditor.test.tsx
mjuarros: Removed in 20284ad8. The remaining tests cover the relevant theme configuration, callback ordering, and eager-loading regression without the duplicate render assertion.

packages/graph-explorer/src/components/CodeEditor.tsx
mjuarros: Added in 20284ad8. The comment now records that Monaco applies the theme during editor creation and that using `loader.init()` would sacrifice lazy loading.

packages/graph-explorer/src/components/CodeEditor.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+vi.mock("@monaco-editor/react", () => ({
+  Editor: mockEditor,
+  loader: { init: mockLoaderInit },
+}));
+
+describe("CodeEditor", () => {
+  test("should not eagerly initialise Monaco at import time", () => {
+    // Guards against calling loader.init() at module scope and losing lazy loading.
+    expect(mockLoaderInit).not.toHaveBeenCalled();
+  });
```
kmcginnes: This one is now dead, and it is my fault rather than yours. Asking you to move to a hoisted `vi.mock` with a static import is what killed it, and I did not think it through when I suggested it.

`vite.config.ts` sets `resetMocks: true`, so every spy's call history is wiped before each test runs. The static `import { CodeEditor } from "./CodeEditor"` at line 7 now happens at module load, well before that reset. So whatever `loader.init` did at import time is erased by the time the assertion reads it.

I proved it rather than guessing. With this on top of `CodeEditor.tsx`:

```ts
import { Editor, loader, type Monaco } from "@monaco-editor/react";

loader.init();
```

which is precisely the regression the comment says it guards, all three tests still pass. The old version caught it only because `await import("./CodeEditor")` ran inside the test body, after the reset.

Two ways out, and I verified both:

**Delete it.** My preference. Lazy loading is a behavior of `@monaco-editor/react`, not of this component, and the `Editor` that triggers it is mocked away here, so this file is a slightly awkward home for the guarantee. Drop the test and the `loader` entry from the mock factory at line 28.

**Or capture the count before the reset.** If you would rather keep the tripwire, this works. It passes on the current code and fails on the `loader.init()` mutation above, which I checked both ways:

```ts
// Captured before per-test mock reset wipes the import-time call record.
let initCallsAtImport = -1;
beforeAll(() => {
  initCallsAtImport = mockLoaderInit.mock.calls.length;
});

test("should not eagerly initialise Monaco at import time", () => {
  expect(initCallsAtImport).toBe(0);
});
```

I lean toward deleting, but you have looked at this file more than I have this week. Either is fine by me and I will approve on whichever you pick.

kmcginnes (commented):
Everything from the last round landed, and I checked each one rather than taking the replies on trust. Rebased onto current `main` with nothing behind, squashed to a single commit, and the title and description now describe the `beforeMount` mechanism. The hoisted `vi.mock` with a static import is exactly right, `EditorProps` on the fake means a signature change upstream will now break the typecheck, and dropping `data-theme` in favour of `mock.calls[0][0]` removed the last assertion that was reading the mock's own plumbing.

I ran it locally: `pnpm checks` clean, and the full suite passes at 220 files and 2673 tests, so the file-level `vi.mock` is not leaking into anything that relies on the global mock in `setupTests.ts`.

I also mutation-tested all three tests, since a test that cannot fail is worse than no test. Two of them bite:

- Moving `defineTheme` back into `onMount`, the original bug, fails `should configure the theme before invoking the caller beforeMount`.
- Flipping `base` to `"vs-dark"` also fails it, so tightening to `objectContaining({ base: "vs", inherit: true })` did the job.
- Deleting the `theme` prop fails `should render the editor with the graph-explorer-light theme`.

The third one does not, and that is the one thing left. Details inline.

One knock-on: if the guard goes away, the "How to read" line that says the test file covers "the lazy-loading regression" needs a trim too.

packages/graph-explorer/src/components/CodeEditor.test.tsx
mjuarros: Removed in 4444eaca. I chose the deletion path: the ineffective eager-initialization test and its loader mock are gone, and the PR description no longer claims lazy-loading test coverage. `pnpm test CodeEditor.test.tsx` and `pnpm checks` both pass.

kmcginnes (approved):
Verified on `4444eac`. `pnpm checks` clean, full suite green at 2672 tests, and I mutation-tested both remaining tests: reverting `defineTheme` to `onMount`, flipping `base` to `"vs-dark"`, deleting the `theme` prop, and dropping the caller's `beforeMount` each fail a test. No dead assertions left.

Deleting the eager-init guard was the right call. Thanks for the patience across three rounds on a nine-line fix.

PR #2142, opened 2026-08-31
Fix non-square icons squashed instead of scaled to fit

### LINKED ISSUE #2108: Non-square icons are squashed instead of scaled to fit
A node icon that is not square is stretched into a square rather than scaled down to fit, so it renders distorted. Affects both the graph canvas and the DOM surfaces, consistently.

## Reproduce

1. Style a node type with a custom uploaded raster icon whose width and height differ — a wide logo shows it clearly.
2. Look at the node on the graph canvas, and at the same node type in a search result row.

The icon fills a square box, distorted. Expected: scaled down to fit the icon box, keeping its aspect ratio.

## Why it happens

**Canvas.** `components/Graph/styles/defaultNodeStyle.ts` sets:

```ts
backgroundFit: "none",
backgroundWidth: "60%",
backgroundHeight: "60%",
```

Per cytoscape 3.34.0 `drawInscribedImage`, `background-width`/`-height` **override** the image’s natural dimensions, and `background-fit` then scales whatever those left:

```js
var w = imgW, h = imgH;
if (background-width  !== auto) w = pct * nodeTW;   // 0.6 * 24 = 14.4
if (background-height !== auto) h = pct * nodeTH;   // 0.6 * 24 = 14.4
if (fit === "contain") { var scale = Math.min(nodeTW / w, nodeTH / h); w *= scale; h *= scale; }
```

So both axes are forced to 14.4 and the image is stretched into a square.

**`background-fit: contain` does not fix this** — it runs *after* the override, when `w` and `h` are already equal, so it just scales the squashed square up to fill the node. Verified in headless Chrome against real cytoscape 3.34.0.

**DOM.** `components/VertexIcon.tsx` renders ``. Forcing both dimensions with the default `object-fit: fill` stretches the image the same way.

`VertexSymbol` is already correct — its `<image>` carries `preserveAspectRatio="xMidYMid meet"`.

## Fix sketch

There is no configuration-only fix: cytoscape cannot fit an image to a box and preserve its ratio at the same time, so the intrinsic ratio has to be known before the style is generated.

1. Measure the raster’s natural size when it resolves (`core/icons/iconRegistry.ts`) and carry the aspect ratio on the resolved icon.
2. Compute `background-width`/`background-height` from it, shrinking the shorter axis so the icon fits the 60% box instead of stretching to it — e.g. a 4:1 image becomes `60%` / `15%`.
3. `VertexIcon`: add `object-contain` to the ``.
4. Generated SVG icons need nothing — they carry a viewBox and letterbox themselves.

Known costs, which is why this was split out rather than bundled into #2102:

- Raster resolution becomes asynchronous, where today the url needs no work.
- Adds an image-load failure mode, needing a fallback to square.
- No test environment loads images, so measuring never settles without an `Image` test double in `setupTests.ts` — which changes the environment for every test file.
- `useBackgroundImageMap` has to return dimensions alongside the image, so its return type and name both change, rippling into `useGraphStyles`.

## Notes

Pre-existing behavior, unchanged by the icon pipeline rework in #2102 — `defaultNodeStyle.ts` has carried these values since before that branch. Distinct from #2105, which is about *recoloring* custom SVG icons rather than sizing them.

---

> [!IMPORTANT]

mjuarros:
## Description

Non-square custom vertex icons (raster images and SVG) were stretched into a square instead of scaled down to fit, distorting them. This happened consistently on both the graph canvas and DOM surfaces (search results, legend).

The root cause: Cytoscape's `drawInscribedImage` does not preserve aspect ratio when `background-width` and `background-height` are both set to percentages. They override the image's natural dimensions, then `background-fit: contain` scales the already-squashed result — the image is already distorted by then. The DOM rendered images with `object-fit: fill` (default), which has identical behavior.

## How it's fixed

**Icon measurement and aspect-ratio carry**: The icon registry now measures the natural width/height of raster images when they resolve. That ratio is carried on the icon object through the system.

**Aspect-fit computation**: The background width/height for the Cytoscape canvas are computed from the target box (60% of node size) and the image's aspect ratio. The shorter axis is shrunk so the image fits inside the box without stretching — e.g., a 4:1 wide logo becomes `60%` width, `15%` height.

**DOM surface fix**: `VertexIcon` gains `object-contain` so its `` respects the aspect ratio.

**SVG handling**: SVG images already carry a viewBox, so they letterbox themselves naturally with `preserveAspectRatio=xMidYMid meet`. New utility functions extract and validate viewBox data.

## How to read

1. [src/core/icons/aspectFit.ts](https://github.com/aws/graph-explorer/pull/2142/files#diff-3050d860272937958e8bbce3a6f7e9ad072cf0e39ac8c9cfeefa9998e5cf5e90) — core logic to compute fit dimensions from aspect ratio
2. [src/core/icons/iconRegistry.ts](https://github.com/aws/graph-explorer/pull/2142/files#diff-86bbb97245d44765403affaf7e75d27bcef37c8a9d1a68cf1c4f9957e3cf302e) — measures image dimensions and carries aspect ratio
3. [src/components/VertexIcon.tsx](https://github.com/aws/graph-explorer/pull/2142/files#diff-01799c2424365caf20f30abbf6bcf99abf263cce55403e80f6c1e5b206b114d0) — adds object-contain to preserve aspect on DOM
4. [modules/GraphViewer/useBackgroundImageMap.ts](https://github.com/aws/graph-explorer/pull/2142/files#diff-b0f70630fad24aed9b413555182e0a63252b95ee6d20d805a46d39e716deccca) — applies aspect-fit logic to canvas styles
5. [src/core/icons/svgViewBox.ts](https://github.com/aws/graph-explorer/pull/2142/files#diff-2b83650f76b6edd3ad76567762128cfdd07cb1f9717b084cb01be228defbda83) — SVG viewBox extraction and validation

Supporting files (tests, minor updates): CONTEXT.md, setupTests.ts, and test files for all new/modified functions.

## Validation

All unit tests pass. Manual testing confirmed non-square icons now render at their natural aspect ratio on the graph canvas and in DOM surfaces (search results, legend).

Before Fix
wide-logo-fail-handling

After Fix
wide-icon-success-handling

- Closes #2108

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -54,7 +54,7 @@ The classification of a Vertex Type's configured icon into what it takes to rend
 _Avoid_: Icon type (ambiguous with `iconImageType`, the stored MIME string)
 
 **Icon Registry**:
-The single store of resolved icons, keyed by Icon Source Id and shared by every Icon Surface. Holds a color-free artifact — a sanitized SVG string or a raster url — so applying a Vertex Type's color stays a pure transform at the point of use. A plain external store outside React/Jotai, bridged by `useSyncExternalStore`; explicitly **not** TanStack Query, because a per-hook subscription scaled with Vertex Type count and locked up the Schema View at 10k. Resolves a raster url synchronously, allows a failed icon three attempts in total, and never stores a failure as a result. See `docs/adr/20260813-icon-registry-not-react-query.md`.
+The single store of resolved icons, keyed by Icon Source Id and shared by every Icon Surface. Holds a color-free artifact — a sanitized SVG string or a raster url, plus its natural width/height so a non-square icon scales instead of stretching — so applying a Vertex Type's color stays a pure transform at the point of use. A plain external store outside React/Jotai, bridged by `useSyncExternalStore`; explicitly **not** TanStack Query, because a per-hook subscription scaled with Vertex Type count and locked up the Schema View at 10k. Every icon kind resolves asynchronously (a raster's dimensions are measured by loading it), allows a failed icon three attempts in total, and never stores a failure as a result. See `docs/adr/20260813-icon-registry-not-react-query.md`.
 _Avoid_: Icon cache (it is the source of truth for resolution, not a layer in front of one)
 
 **Icon Surface**:
--- a/packages/graph-explorer/src/components/VertexIcon.tsx
+++ b/packages/graph-explorer/src/components/VertexIcon.tsx
@@ -3,13 +3,16 @@ import { DynamicIcon } from "lucide-react/dynamic";
 import SVG from "react-inlinesvg";
 
 import { useVertexStyle, type VertexStyle, type VertexType } from "@/core";
+import { ensureSvgViewBox, SVG_ALLOWED_ATTR } from "@/core/icons";
 import { cn } from "@/utils";
 import { getLucideName, isValidLucideIconName } from "@/utils/lucideIcons";
 
 function sanitizeSvg(svg: string): string {
-  return DOMPurify.sanitize(svg, {
+  const sanitized = DOMPurify.sanitize(svg, {
     USE_PROFILES: { svg: true, svgFilters: true },
+    ALLOWED_ATTR: SVG_ALLOWED_ATTR,
   });
+  return ensureSvgViewBox(sanitized);
 }
 
 interface Props {
@@ -54,7 +57,7 @@ function VertexIcon({ vertexStyle, className, alt }: Props) {
     <img
       src={vertexStyle.iconUrl}
       alt={altText}
-      className={cn("size-6 shrink-0", className)}
+      className={cn("size-6 shrink-0 object-contain", className)}
       style={{ color: vertexStyle.color }}
     />
   );
--- a/packages/graph-explorer/src/core/icons/aspectFit.test.ts
+++ b/packages/graph-explorer/src/core/icons/aspectFit.test.ts
@@ -1,0 +1,17 @@
+import { describe, expect, it } from "vitest";
+
+import { fitAspectRatio } from "./aspectFit";
+
+describe("fitAspectRatio", () => {
+  it("caps the wider axis at base and shrinks the shorter one proportionally", () => {
+    expect(fitAspectRatio(400, 100, 24)).toEqual([24, 6]);
+  });
+
+  it("caps the taller axis at base and shrinks the shorter one proportionally", () => {
+    expect(fitAspectRatio(100, 400, 24)).toEqual([6, 24]);
+  });
+
+  it("returns the base for both axes when already square", () => {
+    expect(fitAspectRatio(100, 100, 24)).toEqual([24, 24]);
+  });
+});
--- a/packages/graph-explorer/src/core/icons/aspectFit.ts
+++ b/packages/graph-explorer/src/core/icons/aspectFit.ts
@@ -1,0 +1,16 @@
+/**
+ * Scales a width/height pair to fit a `base`-sized box, preserving aspect
+ * ratio: the longer axis becomes exactly `base`, the shorter one shrinks
+ * proportionally. Returns raw numbers — each caller formats its own unit
+ * (an absolute pixel size, a cytoscape percentage, ...).
+ */
+export function fitAspectRatio(
+  width: number,
+  height: number,
+  base: number,
+): [width: number, height: number] {
+  const aspectRatio = width / height;
+  return aspectRatio >= 1
+    ? [base, base / aspectRatio]
+    : [base * aspectRatio, base];
+}
--- a/packages/graph-explorer/src/core/icons/iconImageUrl.test.ts
+++ b/packages/graph-explorer/src/core/icons/iconImageUrl.test.ts
@@ -103,4 +103,40 @@ describe("toIconImageUrl", () => {
 
     expect(red).not.toBe(blue);
   });
+
+  // Issue #2108: forcing every icon's intrinsic size to a fixed 24x24 square
+  // bakes a mismatched-aspect letterbox into the rasterized image, which the
+  // consumer's own aspect-aware background-width/height then stretches a
+  // second time — distorting a non-square icon worse than doing nothing.
+  describe("non-square icons (issue #2108)", () => {
+    const WIDE_SVG = `<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 400 100"><rect width="400" height="100"/></svg>`;
+    const TALL_SVG = `<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 100 400"><rect width="100" height="400"/></svg>`;
+
+    it("scales a wide icon's intrinsic width/height to its real aspect ratio", () => {
+      const result = toIconImageUrl(
+        { kind: "svg", svg: WIDE_SVG, width: 400, height: 100 },
+        "#FF0000",
+      );
+
+      expect(decode(result)).toContain('width="24"');
+      expect(decode(result)).toContain('height="6"');
+    });
+
+    it("scales a tall icon's intrinsic width/height to its real aspect ratio", () => {
+      const result = toIconImageUrl(
+        { kind: "svg", svg: TALL_SVG, width: 100, height: 400 },
+        "#FF0000",
+      );
+
+      expect(decode(result)).toContain('width="6"');
+      expect(decode(result)).toContain('height="24"');
+    });
+
+    it("falls back to a 24x24 square when dimensions are unknown", () => {
+      const result = toIconImageUrl({ kind: "svg", svg: WIDE_SVG }, "#FF0000");
+
+      expect(decode(result)).toContain('width="24"');
+      expect(decode(result)).toContain('height="24"');
+    });
+  });
 });
--- a/packages/graph-explorer/src/core/icons/iconImageUrl.ts
+++ b/packages/graph-explorer/src/core/icons/iconImageUrl.ts
@@ -1,7 +1,9 @@
 import type { ResolvedIcon } from "./iconRegistry";
 
-/** Intrinsic size; both consumers scale from it. Matches the cytoscape node size. */
-const ICON_SIZE = "24";
+import { fitAspectRatio } from "./aspectFit";
+
+/** Intrinsic size baseline; matches the cytoscape node size. */
+const ICON_SIZE = 24;
 
 /**
  * Pure transform to an image url.
@@ -16,17 +18,42 @@ export function toIconImageUrl(icon: ResolvedIcon, color: string): string {
     case "raster":
       return icon.url;
     case "svg":
-      return encodeSvg(applySizeAndColor(icon.svg, color));
+      return encodeSvg(
+        applySizeAndColor(icon.svg, color, icon.width, icon.height),
+      );
   }
 }
 
-function applySizeAndColor(svgContent: string, color: string): string {
+/**
+ * Sets the SVG's own intrinsic width/height. This must preserve the icon's
+ * real aspect ratio (scaled to fit a 24px box), not force a fixed square:
+ * forcing a square here bakes a mismatched-aspect letterbox into the
+ * rasterized image, which the consumer's own aspect-aware background-width/
+ * height then stretches a second time, distorting worse than doing nothing.
+ */
+function applySizeAndColor(
+  svgContent: string,
+  color: string,
+  naturalWidth?: number,
+  naturalHeight?: number,
+): string {
   const doc = new DOMParser().parseFromString(svgContent, "application/xml");
   const root = doc.documentElement;
-  root.setAttribute("width", ICON_SIZE);
-  root.setAttribute("height", ICON_SIZE);
+  const [width, height] = fitToIconSize(naturalWidth, naturalHeight);
+  root.setAttribute("width", String(width));
+  root.setAttribute("height", String(height));
   applyColor(root, color);
   return new XMLSerializer().serializeToString(root);
+}
+
+function fitToIconSize(
+  naturalWidth?: number,
+  naturalHeight?: number,
+): [width: number, height: number] {
+  if (!naturalWidth || !naturalHeight) {
+    return [ICON_SIZE, ICON_SIZE];
+  }
+  return fitAspectRatio(naturalWidth, naturalHeight, ICON_SIZE);
 }
 
 /**
--- a/packages/graph-explorer/src/core/icons/iconRegistry.test.ts
+++ b/packages/graph-explorer/src/core/icons/iconRegistry.test.ts
@@ -49,7 +49,7 @@ describe("iconRegistry", () => {
     iconRegistry.request([source]);
     await settle();
 
-    expect(iconRegistry.getSnapshot().get(iconSourceId(source)!)).toStrictEqual(
+    expect(iconRegistry.getSnapshot().get(iconSourceId(source)!)).toMatchObject(
       {
         kind: "raster",
         url: "https://example.test/a.png",
@@ -58,18 +58,23 @@ describe("iconRegistry", () => {
     expect(fetch).not.toBeCalled();
   });
 
-  // A url needs no resolution, so making the consumer wait a render for it
-  // would be a pointless async round trip.
-  it("resolves a raster icon synchronously", () => {
+  // Measuring a raster's natural size requires loading it, so — unlike a url,
+  // which needs no resolution — this can no longer settle in the same tick.
+  it("measures a raster icon's natural dimensions", async () => {
     const source = classifyIconSource({
       iconUrl: "https://example.test/a.png",
       iconImageType: "image/png",
     });
 
     iconRegistry.request([source]);
+    await settle();
 
-    expect(iconRegistry.getSnapshot().has(iconSourceId(source)!)).toBe(true);
-    expect(iconRegistry.pendingCount).toBe(0);
+    expect(iconRegistry.getSnapshot().get(iconSourceId(source)!)).toMatchObject(
+      {
+        width: expect.any(Number),
+        height: expect.any(Number),
+      },
+    );
   });
 
   it("fetches and sanitizes a remote svg", async () => {
--- a/packages/graph-explorer/src/core/icons/iconRegistry.ts
+++ b/packages/graph-explorer/src/core/icons/iconRegistry.ts
@@ -4,11 +4,13 @@ import { logger } from "@/utils";
 import { getLucideSvgString } from "@/utils/lucideIcons";
 
 import { type IconSource, type IconSourceId, iconSourceId } from "./iconSource";
+import { SVG_ALLOWED_ATTR } from "./svgSanitize";
+import { ensureSvgViewBox } from "./svgViewBox";
 
 /** An icon resolved to a renderable form, with no color applied yet. */
 export type ResolvedIcon =
-  | { kind: "raster"; url: string }
-  | { kind: "svg"; svg: string };
+  | { kind: "raster"; url: string; width?: number; height?: number }
+  | { kind: "svg"; svg: string; width?: number; height?: number };
 
 /**
  * Bounded so a permanently broken icon stops re-fetching, but not one-shot: a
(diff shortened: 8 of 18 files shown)


kmcginnes:
Your root-cause analysis is correct. There's a shorter route to the same pixels, though.

Wrap every icon, raster included, in a padded square SVG and let `preserveAspectRatio` do the fitting:

```
background-fit: contain
background-width: auto
background-height: auto

<svg viewBox="0 0 100 100">
  <image href="<icon>" x="20" y="20" width="60" height="60"
         preserveAspectRatio="xMidYMid meet"/>
</svg>
```

The prototype I sent produces pixel-identical output to your branch across 4:1, 1:4, 1:1, 10:1 and a raster, in Chrome, Firefox and Safari 27. No measurement anywhere.

Note that `contain` with `auto` alone is not enough: it fits the 24x24 bounding box, and our nodes are ellipses, so a square-ish icon's corners spill outside the circle. The padded wrapper is what keeps the 60% inset.

This also lands the canvas on the same mechanism `VertexSymbol` already uses on the DOM side.

## What it removes

`measureImageDimensions` and `new Image()`, async raster resolution, `width?`/`height?` on `ResolvedIcon`, `fitAspectRatio`, `computeAspectRatioAwareDimensions`, `extractSvgDimensions`, `ensureSvgViewBox`, and the `Image` double in `setupTests.ts`. Around 40 lines instead of 448, none async. Most of my review notes disappear with that code rather than needing fixes.

## Still needed

Delete `svgSanitize.ts` and both `ALLOWED_ATTR` arguments. The option never takes effect: DOMPurify reads it at `purify.cjs.js:742`, then line 806 runs `ALLOWED_ATTR = create(null)` inside `if (USE_PROFILES)` and rebuilds from the profile sets. Only `ADD_ATTR` composes. Output is byte-identical with and without it across an 8-SVG diff, and the svg profile at line 316 already allows `width`, `height`, `viewbox` and `preserveaspectratio`, so the doc comment's premise is wrong too. Worse than dead: the list omits `d`, `points`, `transform` and `stroke-width`, so removing the apparently-redundant `USE_PROFILES` later would blank every path-based icon.

Restore the `\u0000` cache key separator in `useBackgroundImageMap.ts`. `|` occurs in both the id and the color, so two entries can collide and swap icons between types.

Comment the `background-fit` line with why `contain` alone fails, plus a line in `CONTEXT.md`.

Fix the PR description's SVG bullet. It says SVGs "already carry a viewBox, so they letterbox themselves naturally," but the branch rewrites every SVG's intrinsic size. Your commit message has it right.

Rebase too. The `safeSessionStorage.test.ts` failure is a stale base, fixed by 5a50963d.

## Pre-existing

The review surfaced several older issues in files you touched. [#2182](https://github.com/aws/graph-explorer/issues/2182) is filed; I'm tracking the rest. None block this.

mjuarros:
Thanks — the prototype made this much easier to reason about, and four of the five action items were straightforward. All are applied and the branch is rebased onto `main` (now at `66603850`).

## Applied

**Deleted `svgSanitize.ts` and both `ALLOWED_ATTR` arguments.** You're right, and it's worse than dead. I verified against the installed DOMPurify 3.4.15: line 804 resolves `ALLOWED_ATTR` from config, then line 870 — inside `if (USE_PROFILES)` — runs `ALLOWED_ATTR = create(null)` and rebuilds purely from the profile sets, so the option never applied. The `svg` profile already allows `width`, `height`, `viewbox`, `preserveaspectratio`, so the doc comment's premise was false too. And it also allows `d`, `points`, `transform`, `stroke-width`, which my list omitted — exactly the trap you described. Rather than just delete it I added a test that pins those eight attributes surviving sanitization, so nobody reintroduces an allowlist on the same false premise. Since the allowlist was introduced by this branch, I folded the removal into the original commit; it never appears in the history now.

**Restored the `\u0000` cache-key separator.** Confirmed reachable, and I have it on record as a red test first: with `|`, the second vertex type resolved to `style="color:x|#FF0000"` — the *other* type's color, i.e. the icon swap. One nuance for the record: `IconSourceId` embeds the user-supplied url verbatim, but it's `color` that has to contain the `|` for keys to collide, and the color picker can't produce that. So it's malformed-config hardening rather than something users hit today. `\u0000` is still strictly correct and free.

**Added the `background-fit` rationale comment and a `CONTEXT.md` line.** Both now record the ellipse/inset reason, plus the constraint in the next section.

**Rebased.** `safeSessionStorage.test.ts` passes. One conflict: `main` has since dropped `expect.extend(matchers)` (now covered by the `@testing-library/jest-dom/vitest` import), so I took `main`'s side and kept only the `Image` double.

`pnpm check:lint`, `pnpm check:types` and `pnpm test` (2719 tests) are green.

## On variant E — it breaks remote raster icons

This is the one place I'd push back, and I think the harness hides it.

I ran the wrapper through Chrome the way cytoscape consumes it (`Image` → canvas) and measured the pixels. The test image is a uniform opaque orange 400×100 PNG:

| case | rendered box | ratio | is it our image? |
|---|---|---|---|
| E: wrapper + **external http** raster | 54×60 | 0.9 | **no** |
| E: wrapper + **inline data-url** raster | 60×16 | 3.75 ✔ | yes |
| today: bare external raster url | 100×100 | 1.0 (the bug) | yes |

In the external case the rendered colors are `88,174,57` / `163,163,163` / `197,214,243` with **zero orange pixels** — that's Chrome's broken-image placeholder, not the icon. Same for `href` and `xlink:href`.

The cause is the sandbox your own `VertexSymbolIcon` comment already names: an SVG consumed as an image is a separate, script-disabled image document, and it won't fetch external resources either. The wrapper is exactly that once it's a `data:` uri. The last row shows the URL itself is perfectly loadable — it's the nesting that kills it.

The prototype can't surface this because its raster is `canvas.toDataURL()`, i.e. inline, which is the one raster case that works. In production `toIconImageUrl` returns `icon.url` verbatim for rasters, and the fixtures are `https://example.test/a.png`.

So for a remote raster, E trades "distorted but recognizable" for "a broken-image glyph on every node of that type."

On "the same mechanism `VertexSymbol` already uses" — structurally true, and I checked: `ICON_RATIO = 0.6`, `viewBox 0 0 96 96`, `<image x=19.2 y=19.2 width=57.6 height=57.6 preserveAspectRatio="xMidYMid meet">`. But `VertexSymbol` is **live DOM**, where an external `href` fetches normally; that's why it renders remote rasters today. Same markup, opposite behaviour, one level further in.

That's why I kept the measurement. Working the `drawInscribedImage` arithmetic, the alternatives for a remote raster are: `contain` + `auto`/`auto` preserves ratio but fills the full 24px and loses the inset (variant B), and mixing one percentage with one `auto` is nonsense (`60%` width + auto height on 4:1 gives 3.46×24). Aspect *and* inset requires knowing the ratio — so the measurement is load-bearing rather than incidental, unless we accept raster icons rendering visibly larger than SVG ones and diverging from `VertexSymbol`'s 0.6.

Happy to be wrong about this if it reproduces differently for you. The probe is self-contained (spins its own server, generates its own PNG, no deps) and prints that table — I can drop it in a gist or add it to the prototype if useful. Worth re-running in Firefox/Safari; I only had Chrome headless, though the restriction is standard SVG-as-image sandboxing rather than a Chrome quirk.

Two smaller notes on the framing: the 448 lines include ~196 lines of tests plus the 32-line file that's now gone (production delta is ~220), and the raster row in "pixel-identical across 4:1, 1:4, 1:1, 10:1 and a raster" is the inline-data-url case.

## PR description

You're right that the SVG bullet was wrong — the branch does rewrite every SVG's intrinsic size. Corrected bullet:

> **SVG handling**: An SVG's intrinsic `width`/`height` are rewritten to the aspect-fitted size rather than forced to a square. Forcing a square bakes a mismatched-aspect letterbox into the rasterized image, which the consumer's own aspect-aware `background-width`/`height` then stretches a second time. A `viewBox` is synthesized from `width`/`height` when absent, because without one there is no coordinate system to scale from and resizing the root just clips the content.

I have applied that to the description, and corrected the "How to read" entry for `svgViewBox.ts` — it synthesizes a `viewBox` rather than extracting and validating one.

mjuarros:
Correcting my previous comment: **my objection to variant E was wrong**, and I'd rather retract it explicitly than leave it standing.

## Where I was wrong

I claimed the padded-SVG wrapper would break remote raster icons. The browser behaviour I measured is real — inside a `data:` uri wrapper an external `<image href>` is not fetched, and Chrome paints a broken-image glyph — but I never checked whether a remote raster icon is reachable in the first place. It isn't. `ICON_VALUE_PATTERN` in `core/styling/stylingParser.ts` gates every icon value to `lucide:<name>` or `data:image/*;base64,`, enforced at both write paths (`isAllowedIconValue` at the upload seam, `safeIconValue` in the import parser, which strips an injected `iconUrl` specifically so it cannot bypass the gate). Every icon is inline. The `https://example.test/a.png` fixtures are test-only, and I over-read them as representative.

So the failure mode I described cannot occur, and my conclusion that the measurement is load-bearing does not follow from it.

## What I verified instead

I spiked variant E against the app's real cytoscape 3.34.3 with production node geometry from `defaultNodeStyle.ts`, icons supplied as base64 data uris exactly as the upload seam stores them, measuring the drawn icon as a pixel diff against an icon-free baseline. Two runs, identical output.

| case | current (measured %) | variant E | E + `ensureSvgViewBox` |
|---|---|---|---|
| png 4:1 | 116×30 r3.9 | 116×30 r3.9 | 116×30 r3.9 |
| png 1:4 | 28×116 r0.24 | 30×116 r0.26 | 30×116 r0.26 |
| png 1:1 | 116×116 r1 | 116×116 r1 | 116×116 r1 |
| svg 4:1 | 116×30 r3.9 | 116×30 r3.9 | 116×30 r3.9 |
| svg 4:1 `currentColor` | 116×30 r3.9 | 116×30 r3.9 | 116×30 r3.9 |
| lucide-like 1:1 | 96×96 r1 | 96×96 r1 | 96×96 r1 |
| **svg 4:1, no viewBox** | **116×30 r3.9** | **116×116 r1.0** | **116×30 r3.9** |
| svg 4:1, viewBox only | 116×30 r3.9 | 116×30 r3.9 | 116×30 r3.9 |

You're right on the main point: E is pixel-equivalent everywhere it matters (the ±2px is antialiasing at 8× zoom), and it keeps the 60% inset. I also checked the thing I thought might bite — `currentColor` survives the extra nesting level, so baking the colour into the inner SVG still works when that SVG is one level deeper as an `<image href>`.

## One entry to keep on the removal list

`ensureSvgViewBox` is load-bearing. An SVG with `width`/`height` and no `viewBox` renders **square** under E — 116×116, ratio 1.0 — which is #2108 reintroduced. `preserveAspectRatio` has no intrinsic ratio to fit, so the image fills the padded box. Synthesizing the viewBox first restores the correct 116×30.

That case is reachable: it is what a plain `<svg width height>` export produces, and users upload arbitrary SVGs.

## Plan

Implementing E now, keeping `ensureSvgViewBox`, with a regression test for the no-viewBox case so it can't quietly come back. That removes `measureImageDimensions` and `new Image()`, the async raster path, `width?`/`height?` on `ResolvedIcon`, `fitAspectRatio`, `computeAspectRatioAwareDimensions`, `extractSvgDimensions`, and the `Image` double in `setupTests.ts`; `useBackgroundImageMap` goes back to `Map<VertexType, string>` and the sizing moves to `background-fit: contain` with `auto` axes in `defaultNodeStyle`.

One implementation note: the wrapper has to be applied on the canvas path only, not inside `toIconImageUrl`. `VertexSymbolIcon` already insets to 60% in its own SVG coordinates, so wrapping there too would apply the inset twice and render at 36%.

Also worth noting the double percent-encoding is a non-issue: it inflates the style string 1.2× for rasters, 1.6× for SVG, 2.2× for lucide, with a largest absolute value of ~2.5KB. Unique icons stay in the dozens, so it doesn't approach the Schema View concern.

Caveats on my numbers: Chrome headless only, the lucide row is a stand-in rather than real lucide output, and the bbox carries ~2px of antialiasing noise.

mjuarros:
Pushed as `4ca8fdd9` — variant E is in, with `ensureSvgViewBox` kept for the reason above.

What went away: `measureImageDimensions` and `new Image()`, the async raster path (a raster url resolves synchronously again), `width?`/`height?` on `ResolvedIcon`, `fitAspectRatio` and `aspectFit.ts`, `computeAspectRatioAwareDimensions`, `extractSvgDimensions`, and the `Image` double in `setupTests.ts`. `useBackgroundImageMap` is back to `Map<VertexType, string>`, and `useGraphStyles.ts` and `setupTests.ts` are now identical to `main` again. Net diff is 14 files, +328/-47, of which ~240 lines are tests.

What stayed: `ensureSvgViewBox`, and the parse/serialize in `toIconImageUrl` that bakes the colour.

I verified the branch's real output rather than a port of it — emitting the actual urls the hook produces, then rendering them on cytoscape 3.34.3 with the updated `defaultNodeStyle`:

| icon | drawn box | ratio | width vs node |
|---|---|---|---|
| 4:1, with viewBox | 116×30 | 3.87 | 60.4% |
| 4:1, **no viewBox** | 116×30 | 3.87 | 60.4% |
| 1:4, no viewBox | 30×116 | 0.26 | 15.6% |
| 1:1, no viewBox | 116×116 | 1.00 | 60.4% |

The description is updated to describe the wrapper. The two commits from the earlier round (the NUL cache key, and never introducing the `ALLOWED_ATTR` list) are unchanged; I squashed the rationale commit into this one since it documented the approach that just got replaced.

kmcginnes:
This is the approach I wanted, and it works. I verified it in the real app rather than a harness: dev server off your branch against a live `air_routes` TinkerPop container, geometry measured off the cytoscape canvas.

| Icon | Extent (device px) | Circle | Aspect |
|---|---|---|---|
| 4:1 SVG | 144x36 | 28x28 | 1.000 |
| 1:4 SVG | 36x142 | 28x28 | 1.000 |
| 1:1 SVG | 144x142 | 114x114 | 1.000 |
| 4:1 PNG | 144x36 | 28x28 | 1.000 |
| viewBox-less 4:1 SVG | 144x36 | 28x28 | 1.000 |

The inset is exact, 0.600 of the node box on the major axis in every case, centred to within half a device pixel. The 1:1 icon clears the ellipse outline by 18.9px, which is the part the first attempt got wrong. I reproduced the bug on the merge base with the same script for comparison: the 4:1 PNG came out 144x142 with the circle crushed to 28x114, aspect 0.246.

One thing worth knowing that neither of us had established: a viewBox-carrying SVG was already rendering correctly before this branch. #2108 only ever bit raster icons and viewBox-less SVGs. My earlier comment overstated the SVG half.

The nesting question is settled too. In Chrome, a wrapper with a nested `data:` URL paints fine and issues zero external requests, so the SVG-as-image sandbox is not in the way.

## One blocker

`insetIconImage` introduces the first `encodeURIComponent` on the raster path, and `encodeURIComponent` throws `URIError` on a string that isn't well-formed UTF-16. A stored `iconUrl` containing a lone surrogate therefore throws during render of `useGraphStyles`, which lands in the `ErrorBoundary` around the whole `<Outlet />`. The app is replaced by `AppErrorPage`, whose only action reloads and re-throws, and the styling panel that would clear the bad value is inside the blanked route. At the merge base the raster arm returned the url by reference, so nothing on that path could throw.

Fix: test `isWellFormed()` on the url and skip the icon when it fails, logging a warning. The vertex then renders without a background image instead of taking the app down.

There are also two validation gaps further upstream that let a bad value get stored in the first place. Those predate this branch and I'm tracking them separately, so don't worry about them here.

## Before merge

`ensureSvgViewBox` silently does nothing for several real inputs, including an SVG using `xlink:href` without declaring `xmlns:xlink`, and one with no `xmlns` at all. The intermediate `application/xml` parse fails and the function returns its input, so no `viewBox` is synthesized and the icon still renders square. That is the class of icon this branch exists to fix. Parsing as `image/svg+xml` should cover it.

Share the inset geometry instead of redeclaring it. `ICON_RATIO = 0.6` now exists in both `useBackgroundImageMap.ts` and `VertexSymbol.tsx`, and `CONTEXT.md` states it as one rule, so changing either silently desyncs the canvas from the style preview with no test to catch it. `encodeSvg` is duplicated the same way. Also `BOX = 100` is commented "arbitrary", but it sets the resolution the wrapper rasterizes at, and `VertexSymbol` already has the reasoned value for that quantity, `VIEWBOX = 96`, which is exactly 4x a node.

Two tests don't test what they claim. In `iconRegistry.test.ts`, `toContain("d")` passes no matter what, because the fixture's own `preserveAspectRatio` contains a `d`, and `toContain("width")` is satisfied by `stroke-width` alone. Assert values rather than bare names. And in `useGraphStyles.test.tsx` the raster assertion only checks the `data:` prefix, so it passes even if the wrapper lost the url.

There's no `VertexIcon.test.tsx` anywhere, so the inline-DOM surface is the only one of the three whose fix ships unverified.

`CONTEXT.md` still says the canvas passes a raster "as its plain url" two sentences before the new clause about wrapping, so it now describes rasters both ways. The icon-registry ADR has the same staleness in its raster row, and its note about image contexts blocking external references reads as forbidding what the canvas now does. Worth a qualifier distinguishing `data:` URLs.

## Small

- `backgroundWidth`/`backgroundHeight: "auto"` restate cytoscape's own defaults, so the comment crediting them with the fix is off. `contain` plus dropping the percentages is what changed the rendering.
- That same comment should say why `contain` alone fails, naming the ellipse. That was the specific thing I asked for.
- `ensureSvgViewBox` guards on `isNaN` and `> 0`, so `width="1e400"` writes `viewBox="0 0 Infinity Infinity"` and the icon vanishes. `width="100%"` reads as 100 user units and crops the artwork; I measured a 200-wide rect cut off at x=100. Reject anything that isn't a plain number, and use `Number.isFinite`.
- The `try`/`catch` in `ensureSvgViewBox` is unreachable. `DOMParser` with `application/xml` returns a `parsererror` document rather than throwing.
- Copy the "wrapper belongs on the canvas path only, or it insets twice" rule from `CONTEXT.md` into the `toIconImageUrl` doc comment, which is where someone would make that mistake.
- No tall-icon fixture in the new tests. One line.

Nothing else: `pnpm checks` and all 2714 tests pass, the escaping held under every payload I threw at it, and the sanitizer config is unchanged from the merge base.

mjuarros:
Thank you for the geometry verification against a live container — that's a stronger check than anything I ran. All fixed, pushed as four commits on top of the rebase (now at `b6137046`).

## Fixed

**The blocker.** Confirmed: `encodeURIComponent` throws `URIError` on a lone surrogate, this runs during style computation, and `DefaultLayout.tsx` wraps the whole `<Outlet />` with no in-app recovery. `insetIconImage` now checks `iconUrl.isWellFormed()` first and skips the icon with a `logger.warn` instead of calling `encodeURIComponent` on it. Test asserts `renderMap` doesn't throw on a lone-surrogate url and the vertex type is absent from the result.

**`ensureSvgViewBox`'s two numeric bugs.** Switched to `Number.isFinite` and reject a trailing `%`; `1e400` and `100%` are now left untouched instead of synthesizing `Infinity`/a cropping viewBox. `400px` still works. Removed the `try`/`catch` — traced through it and it's genuinely dead: `DOMParser` with `application/xml` never throws, it returns a document whose root tag doesn't match what was requested, so `root.localName !== "svg"` already rejects it before reaching the numeric guard.

**Shared geometry.** `ICON_RATIO`, the box size, and `encodeSvg` now live in one module (`iconGeometry.ts`) that both `useBackgroundImageMap` and `VertexSymbol` import, with a test pinning the values. Took the box size to `96` — your reasoned value — over my arbitrary `100`.

**The two weak tests.** `iconRegistry.test.ts`'s geometry check now asserts full attribute values with distinct numbers per attribute rather than bare names (verified it now fails when `stroke-width`/`d` are actually wrong). `useGraphStyles.test.tsx`'s raster check now decodes the wrapper and confirms the real url is nested inside.

**`VertexIcon.test.tsx`** now exists — object-contain on the raster path, lucide color inheritance, unknown-lucide-renders-nothing.

**Docs.** `CONTEXT.md`'s raster sentence is scoped so it doesn't contradict itself. The ADR's table and external-references line are updated, with a new paragraph on why the wrapper doesn't cost the sandbox. `defaultNodeStyle`'s comment credits `contain` correctly and names the ellipse.

Tall-icon fixture added alongside the wide one.

## Two things I want to correct, since I'd rather flag it than let it stand

**`image/svg+xml` doesn't fix the `xlink` case** — I ran it. Both `application/xml` and `image/svg+xml` are XML-strict in `DOMParser`, so `xlink:href` without a declared `xmlns:xlink` fails identically either way; only lenient `text/html` parsing recovers it. Separately, that input never reaches `ensureSvgViewBox` at all on the untrusted-SVG path: DOMPurify's sanitized output doesn't round-trip through `isParseableSvg`'s `application/xml` check either, so the icon is rejected upstream, in code that predates this branch. The symptom for that input today is a missing icon, not a squashed one.

The "no `xmlns` at all" case, though, I couldn't reproduce — `<svg width="400" height="100">` with no `xmlns` parses fine as `application/xml` and gets its viewBox synthesized correctly in my testing.

Given both of those, I've left `ensureSvgViewBox` on `application/xml` for now rather than switching parsers, since the fix you proposed doesn't close the gap and I don't have a case in hand where it currently fails incorrectly. If you have a concrete input that reaches `ensureSvgViewBox` and still gets skipped, I'd like to see it — that'd change my answer.

**The `px`-suffix concern:** rejecting "anything that isn't a plain number" would have taken `width="400px"` down with it, which is legitimate SVG that was working before. I kept `parseFloat` for that reason and only added the `%` and finiteness checks.

`pnpm checks` and all 2726 tests pass.

PR #2148, opened 2026-09-09
Make safeSessionStorage fallback test independent of Node version

### LINKED ISSUE #2146: safeSessionStorage fallback test depends on the ambient Node version
## Description

The test `resolveSessionStorage > warns and falls back to in-memory storage when sessionStorage is unavailable` asserts the in-memory fallback path but never establishes the condition that triggers it. It depends on the ambient test environment happening to have no `globalThis.sessionStorage`, an assumption its own comment states:

```ts
// The default test environment is non-DOM, so globalThis.sessionStorage is
// undefined and resolveSessionStorage exercises the in-memory fallback.
```

That assumption no longer holds. Node exposes a native `sessionStorage` from v26, so `resolveSessionStorage()` takes its `logger.debug` branch, returns the real store, and the assertion that `logger.warn` was called once fails.

The defect is in the test, not in `resolveSessionStorage`. The production code behaves correctly in both cases. The test covers the fallback branch only by luck of the environment, and it will stop covering that branch entirely once `.nvmrc` moves to Node 26.

Measured:

| Node | `typeof globalThis.sessionStorage` | Result |
| --- | --- | --- |
| v24.16.0, pinned in `.nvmrc` | `undefined` | passes, 3 of 3 |
| v26.7.0 | `object` | fails, 1 of 3 |

Scope is exactly one test. The full suite on v24.16.0 passes 218 of 218 files and 2667 of 2667 tests; on v26.7.0 the same suite reports 1 failed and 2666 passed. No other test depends on the ambient Node version.

CI is unaffected, because `.github/workflows/unit.yml` installs Node from `.nvmrc`. This only bites contributors whose local Node is newer than the pinned version.

## Environment

- OS: macOS, Darwin 25.6.0
- Node: fails on v26.7.0, passes on v24.16.0 (`.nvmrc` pins v24.16.0)
- Graph Explorer Version: 3.2.2 (`main` at `0f17c393`)
- Browser and graph database: not applicable, this affects the test suite only

## Steps to Reproduce

1. Use a Node version of 26 or newer, e.g. `fnm use v26.7.0`.
2. Run `pnpm test safeSessionStorage`.
3. See the failure:

   ```
   FAIL  src/core/StateProvider/safeSessionStorage.test.ts
     > resolveSessionStorage > warns and falls back to in-memory storage when sessionStorage is unavailable
   AssertionError: expected "vi.fn()" to be called once, but got 0 times
     at safeSessionStorage.test.ts:28
   ```

4. Run `fnm exec --using=v24.16.0 pnpm test safeSessionStorage` and it passes.

Note the run also prints `ExperimentalWarning: localStorage is not available because --localstorage-file was not provided`. That is unrelated. `localStorage` still needs the flag; `sessionStorage` does not.

## Expected Behavior

The test forces `globalThis.sessionStorage` to be absent, so it exercises the fallback branch and passes on any Node version.

## Suggested Fix

Make the test hermetic rather than environment-dependent. The second test in the same file already establishes its own condition, forcing the getter to throw and restoring it in a `finally`. Copy that pattern to force the property absent.

Files:

- `packages/graph-explorer/src/core/StateProvider/safeSessionStorage.test.ts`
- `packages/graph-explorer/src/core/StateProvider/safeSessionStorage.ts`

Done when the test passes under both v24.16.0 and v26.7.0, and still fails if the fallback branch is deliberately broken, confirming it covers the branch rather than passing vacuously.

Whether `.nvmrc` should move to Node 26 is a separate question and should not be bundled into this fix.

<audience-footer>

> [!IMPORTANT]

</audience-footer>

mjuarros:
## Description

- Make the unavailable-sessionStorage test establish its own precondition instead of relying on the ambient Node.js runtime.
- Restore the original global property descriptor after the assertion so the test remains isolated on runtimes both with and without native sessionStorage.

## How to read

1. [packages/graph-explorer/src/core/StateProvider/safeSessionStorage.test.ts](https://github.com/aws/graph-explorer/pull/2148/files#diff-642d18c0f25b50d4f9f46275c9db4d212df27b079477e29e08b5d95e96c22f18) — the complete test-only change.

## Validation

- `pnpm test safeSessionStorage` — 3 tests passed
- `pnpm test` — 218 files and 2667 tests passed
- `pnpm check:types` — passed
- Lint — passed
- `git diff --check` — passed
- `pnpm checks` — passed

## Related Issues

- Closes #2146

### Check List

### DIFF
--- a/packages/graph-explorer/src/core/StateProvider/safeSessionStorage.test.ts
+++ b/packages/graph-explorer/src/core/StateProvider/safeSessionStorage.test.ts
@@ -20,14 +20,29 @@ describe("createInMemorySessionStorage", () => {
 });
 
 describe("resolveSessionStorage", () => {
-  // The default test environment is non-DOM, so globalThis.sessionStorage is
-  // undefined and resolveSessionStorage exercises the in-memory fallback.
   test("warns and falls back to in-memory storage when sessionStorage is unavailable", () => {
-    const storage = resolveSessionStorage();
+    const original = Object.getOwnPropertyDescriptor(
+      globalThis,
+      "sessionStorage",
+    );
+    Object.defineProperty(globalThis, "sessionStorage", {
+      configurable: true,
+      value: undefined,
+    });
 
-    expect(vi.mocked(logger.warn)).toHaveBeenCalledOnce();
-    storage.setItem("key", "value");
-    expect(storage.getItem("key")).toBe("value");
+    try {
+      const storage = resolveSessionStorage();
+
+      expect(vi.mocked(logger.warn)).toHaveBeenCalledOnce();
+      storage.setItem("key", "value");
+      expect(storage.getItem("key")).toBe("value");
+    } finally {
+      if (original) {
+        Object.defineProperty(globalThis, "sessionStorage", original);
+      } else {
+        delete (globalThis as { sessionStorage?: Storage }).sessionStorage;
+      }
+    }
   });
 
   test("includes the thrown error in the warning when sessionStorage access throws", () => {
--- a/packages/graph-explorer/src/core/StateProvider/safeSessionStorage.test.ts
+++ b/packages/graph-explorer/src/core/StateProvider/safeSessionStorage.test.ts
@@ -21,28 +21,13 @@ describe("createInMemorySessionStorage", () => {
 
 describe("resolveSessionStorage", () => {
   test("warns and falls back to in-memory storage when sessionStorage is unavailable", () => {
-    const original = Object.getOwnPropertyDescriptor(
-      globalThis,
-      "sessionStorage",
-    );
-    Object.defineProperty(globalThis, "sessionStorage", {
-      configurable: true,
-      value: undefined,
-    });
+    vi.stubGlobal("sessionStorage", undefined);
 
-    try {
-      const storage = resolveSessionStorage();
+    const storage = resolveSessionStorage();
 
-      expect(vi.mocked(logger.warn)).toHaveBeenCalledOnce();
-      storage.setItem("key", "value");
-      expect(storage.getItem("key")).toBe("value");
-    } finally {
-      if (original) {
-        Object.defineProperty(globalThis, "sessionStorage", original);
-      } else {
-        delete (globalThis as { sessionStorage?: Storage }).sessionStorage;
-      }
-    }
+    expect(vi.mocked(logger.warn)).toHaveBeenCalledOnce();
+    storage.setItem("key", "value");
+    expect(storage.getItem("key")).toBe("value");
   });
 
   test("includes the thrown error in the warning when sessionStorage access throws", () => {


kmcginnes (commented):
The fix is correct, and the issue's diagnosis was right that the defect was in the test rather than in `resolveSessionStorage`. Verified on both Node versions locally: 3/3 on v24.16.0 and on v26.7.0, and three mutations of `resolveSessionStorage` (`logger.warn` to `logger.debug`, and returning the native store from the fallback branch) all fail on both. It guards the branch instead of passing vacuously. `.nvmrc` is untouched, so the separate Node 26 question stays separate.

One request before merge.

## Use vi.stubGlobal instead of hand-rolling the descriptor save and restore

`packages/graph-explorer/vite.config.ts:79` already sets `unstubGlobals: true`, so vitest snapshots and restores the property descriptor for anything stubbed through `vi.stubGlobal`. Thirteen other test files in this package use it. The new block reimplements that mechanism by hand.

```ts
test("warns and falls back to in-memory storage when sessionStorage is unavailable", () => {
  vi.stubGlobal("sessionStorage", undefined);

  const storage = resolveSessionStorage();

  expect(vi.mocked(logger.warn)).toHaveBeenCalledOnce();
  storage.setItem("key", "value");
  expect(storage.getItem("key")).toBe("value");
});
```

24 lines becomes 9. The `try`/`finally`, the `if (original)` branch, and the `delete (globalThis as { sessionStorage?: Storage })` cast all go away, and that cast is the kind AGENTS.md asks us to avoid. Against `main` this is +1 line rather than +18.

I ran exactly this: 3/3 on both versions, the mutation check still fails on both, and a temporary probe test confirmed the descriptor matches its import-time value afterwards, on both the absent-property path (Node 24) and the native-accessor path (Node 26).

Two dead ends I checked so nobody else has to:

- `vi.spyOn(globalThis, "sessionStorage", "get")` is not usable for the second test. It fails on v24.16.0 with `The property "sessionStorage" is not defined on the object.`, because no accessor exists to spy on before Node 26. It passes on Node 26 and fails CI.
- Hoisting the stub into a `beforeEach` is not worth it. The second test needs a throwing getter, not an absent property, so the stub is dead there and its only real effect is cleanup as a side effect of the descriptor snapshot. I simulated the obvious later refactor of moving the stub back into the first test and deleting the single-use `beforeEach`: the second test then silently leaks a throwing getter into every test after it.

Leave the second test's explicit `try`/`finally` alone. It is pre-existing, and the verbosity is the right trade for a global that throws on read.

kmcginnes (approved):
Verified at `6f553957`. Diff against `main` is now +1/-2, and the stale comment about the non-DOM test environment went with it.

| Check | v24.16.0 (`.nvmrc`, CI) | v26.7.0 |
| --- | --- | --- |
| `pnpm test safeSessionStorage` | 3/3 | 3/3 |
| Mutant: `logger.warn` to `logger.debug` | fails | fails |
| Mutant: fallback returns the native store | fails | fails |
| Restoration probe | pass | pass |

Each mutant kills only the intended test, so the test still covers both halves of the fallback branch, the warning and the in-memory store. The probe was a temporary fourth test asserting the `sessionStorage` descriptor matches its import-time value after both tests; it confirms `unstubGlobals` restores correctly on the absent-property path (Node 24) and the native-accessor path (Node 26), so nothing leaks into later tests.

`pnpm checks` and `pnpm test` are both clean, 218 files and 2667 tests.

One thing to fix in the squash box rather than in a new commit: `test: simplify session storage fallback setup (#2146)` carries a `test:` prefix, and AGENTS.md asks for no commit message prefixes. The trailing number is also the issue rather than the PR, which is the slot GitHub fills at squash time. Editing the title on merge handles both.

Issue #2149, opened 2026-09-10
Deferred dependency updates that need source changes


kmcginnes:
## Description

A broad dependency sweep brought most of the tree to the latest versions, but seven package updates
were held back because each one needs source changes rather than just a version bump. This epic
collects those follow-ups so they can be picked off one at a time instead of blocking a routine
update round.

Everything that updated cleanly is already done — this is only the remainder.

## High-Level Plan

Each child is independent and can land on its own:

- oxc toolchain (`oxlint`, `oxfmt`, `oxlint-tsgolint`) — newer versions report violations to fix
- `vitest` 5 — matcher types no longer resolve, plus an order-dependent test failure it exposes
- `jotai` 3 — a removed import subpath and the type errors that follow from it
- `crypto-js` — deprecated upstream with no newer version, so it has to be replaced

Ordering suggestion: the oxc toolchain and `crypto-js` work are self-contained. `vitest` 5 and
`jotai` 3 both need investigation before the size of the change is clear.

## Related Issues

- Children are linked as sub-issues below

<audience-footer>

> [!IMPORTANT]
> Internal only — this issue is maintained by the core team and is not accepting external contributions.

</audience-footer>

mjuarros:
I would like to pick this up

Issue #1850, opened 2026-06-18
Add assertion messages to .toThrow() calls to unblock oxlint 1.70 upgrade


kmcginnes:
oxlint 1.70 enables `vitest/require-to-throw-message` by default, which flags 24 pre-existing `.toThrow()` / `.rejects.toThrow()` calls that assert no message. The oxlint bump was held at 1.62.0 in the latest dependency update so this surfaced separately rather than blocking routine bumps.

Add a matcher argument (string or regex) to each call so the assertions verify *which* error is thrown, then bump oxlint to 1.70+ (currently latest) and drop the hold.

## Affected files (24 sites)

| File | Count |
| --- | --- |
| `packages/graph-explorer/src/connector/sparql/parseAndMapQuads.test.ts` | 11 |
| `packages/graph-explorer/src/connector/sparql/fetchNeighbors/index.test.ts` | 4 |
| `packages/graph-explorer/src/connector/queries/edgeConnectionsQuery.test.ts` | 3 |
| `packages/graph-explorer/src/connector/queries/schemaSyncQuery.test.ts` | 1 |
| `packages/graph-explorer/src/connector/sparql/neighborCounts.test.ts` | 1 |
| `packages/graph-explorer/src/connector/sparql/rawQuery.test.ts` | 1 |
| `packages/graph-explorer/src/connector/fetchDatabaseRequest.test.ts` | 1 |
| `packages/graph-explorer/src/core/StateProvider/localDb.test.ts` | 1 |
| `packages/graph-explorer-proxy-server/src/server.test.ts` | 1 |

## Done when

- Every flagged `.toThrow()` asserts a specific message/pattern
- oxlint bumped to 1.70+ and the 1.62.0 hold removed from root `package.json`
- `pnpm run checks` passes

Related: #1753 (enabling other disabled oxlint rules)

<audience-footer>

> [!IMPORTANT]
> Internal only — this issue is maintained by the core team and is not accepting external contributions.

</audience-footer>

kmcginnes:
Overlaps [#2150](https://github.com/aws/graph-explorer/issues/2150), filed in the latest dependency round. Same rule, same call sites — #2150 counts 25 rather than 24, the extra site being `connector/sparql/rawQuery.test.ts:156`, and it also covers oxfmt and oxlint-tsgolint, which are held back for the same reason.

The oxlint 1.70 target here is now stale; latest is 1.82. Flagging so these two don't get worked twice — worth consolidating into one.

[...]

Issue #1753, opened 2026-05-05
Enable remaining disabled oxlint rules


kmcginnes:
Several oxlint rules are currently disabled because they have violations that need code fixes. This issue tracks enabling them.

### `typescript/unbound-method` (65 violations)
Flags passing class methods as callbacks without binding. Many are likely false positives (Jotai setters, React Query methods that don't use `this`), but some could be real bugs. Needs triage.

### `typescript/no-floating-promises` (28 violations)
Requires all Promises to be awaited, returned, or explicitly voided. High-value rule — unhandled rejections are a common source of silent failures. Audit each violation: add `void` for intentional fire-and-forget, `await` for the rest.

### `jsx-a11y/click-events-have-key-events` + `jsx-a11y/no-static-element-interactions` (1 violation)
`ConnectionRow.tsx` has a `<div onClick>` without keyboard support. Fix: use a `<button>` or add `role="button"` + `onKeyDown` handler.

### `jsx-a11y/mouse-events-have-key-events` (2 violations)
`TabularRow.tsx` has `onMouseOver`/`onMouseOut` without `onFocus`/`onBlur`. Fix: add focus/blur handlers for keyboard users, or accept hover highlighting as visual-only and add inline disables.

---

> [!IMPORTANT]
> If you are interested in working on this issue, please leave a comment.

> [!TIP]
> Please use a 👍 reaction to provide a +1/vote. This helps the community and maintainers prioritize this request.

PR #2162, opened 2026-09-10
Upgrade the Oxc toolchain and resolve #2150 lint findings

### LINKED ISSUE #2150: Update the oxc toolchain (oxlint, oxfmt, oxlint-tsgolint)
Three oxc tools were held back from the latest dependency sweep because each newer version reports
violations that need source changes.

## Bumps

- `oxlint` 1.62.0 → 1.82.0 (root `devDependencies`)
- `oxfmt` 0.55.0 → 0.67.0 (root `devDependencies`)
- `oxlint-tsgolint` 0.23.0 → 7.0.2001 (root `devDependencies`)

## Work

**oxlint 1.82** turns on rules the codebase currently violates — 30 errors:

- 25 × `vitest(require-to-throw-message)` — `toThrow()`/`toThrowError()` calls with no expected
  message, in `connector/sparql/parseAndMapQuads.test.ts` (11),
  `connector/sparql/fetchNeighbors/index.test.ts` (4), `connector/queries/edgeConnectionsQuery.test.ts` (3),
  and one each in `connector/queries/schemaSyncQuery.test.ts`, `connector/sparql/neighborCounts.test.ts`,
  `connector/sparql/rawQuery.test.ts`, `connector/fetchDatabaseRequest.test.ts`,
  `core/StateProvider/localDb.test.ts`, `modules/GraphViewer/exportedGraph.test.ts`, and
  `graph-explorer-proxy-server/src/server.test.ts`

  Note the `.not.toThrow()` calls in `connector/LoggerConnector.test.ts` (5) and
  `connector/queryValueError.test.ts` (1) are correctly argument-less and the rule does not flag them.
  A bare `grep` for `.toThrow()` finds 31 sites; only these 25 need a message.
- 4 × `vitest(expect-expect)` — tests with no assertions in `connector/gremlin/fragments.test.ts` (3)
  and `connector/sparql/fragments.test.ts` (1)
- 1 × `react(set-state-in-effect)`

Adding real messages to `toThrow` is a genuine test-quality improvement, not busywork — worth doing
rather than suppressing the rule.

**oxfmt 0.67** — note the specifier was changed from `^0.55.0` to an exact `0.55.0` pin, since oxfmt
is pre-1.0 and a caret let a lockfile refresh shift formatting output with no commit behind it. Bumping
it means editing that exact pin. It reformats one file: `packages/graph-explorer/src/core/styling/stylingParser.ts`,
collapsing the arrow chain in a `zod` `.transform()` call. Run `pnpm format` and commit the result.

**oxlint-tsgolint 7.0.2001** reports 3 × `typescript-eslint(no-unnecessary-type-assertion)`, all in
`packages/graph-explorer/src/core/StateProvider/vertexStylesTransform.test.ts` (lines 137, 145, 181) —
`"round-tag" as ShapeStyle` style assertions where the receiver already accepts the literal type.

## Acceptance

`pnpm checks` and `pnpm test` pass with all three tools at the new versions.

<audience-footer>

> [!IMPORTANT]

</audience-footer>

mjuarros:
This PR upgrades the Oxc toolchain and resolves the 36 new lint findings reported in #2150.

## Core changes

- `package.json` / `pnpm-lock.yaml`: bump `oxlint` to 1.82.0, `oxfmt` to 0.67.0 (kept as `^0.67.0`), and `oxlint-tsgolint` to 7.0.2001.
- `.oxlintrc.json`: configure `expectRoundTrip` as an assertion helper for `vitest/expect-expect`.
- `packages/graph-explorer/src/core/styling/stylingParser.ts`: apply the `oxfmt` 0.67.0 reformat.

## Lint and test fixes

- Add explicit messages to `toThrow()` / `toThrowError()` calls in connector, schema, and proxy tests.
- Remove 3 unnecessary `ShapeStyle` assertions in `vertexStylesTransform.test.ts`.
- Add disable comments for intentional ARIA role usage in `Field`, `EdgePreview`, and `VertexSymbol`.
- Add a disable comment for the Cytoscape `setState` in `useInitCytoscape`.

## How to read

1. [package.json](https://github.com/aws/graph-explorer/pull/2162/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519) and [pnpm-lock.yaml](https://github.com/aws/graph-explorer/pull/2162/files#diff-32824c984905bb02bc7ffcef96a77addd1f1602cff71a11fbbfdd7f53ee026bb) — dependency bump.
2. [.oxlintrc.json](https://github.com/aws/graph-explorer/pull/2162/files#diff-2113aef2ff3d50c021fca1fb6ecf4d4553a0e3f5f51ded7302ad499e43bc2b1e) — `expectRoundTrip` helper config.
3. [packages/graph-explorer/src/core/styling/stylingParser.ts](https://github.com/aws/graph-explorer/pull/2162/files#diff-1c2cdafafef1cae145218e5399fe8e8c881a6474c11ce28ee2fa6c859f350d00) — only oxfmt output change.
4. [packages/graph-explorer/src/components/Field.tsx](https://github.com/aws/graph-explorer/pull/2162/files#diff-11ed8297ed53b1619445a2cd38c4fd45a8992fb5a2e9d9f8b48d91806a122991), [packages/graph-explorer/src/components/EdgePreview/EdgePreview.tsx](https://github.com/aws/graph-explorer/pull/2162/files#diff-824b802627e2e57149321c9df2de09e59d7da15590069bb8c1a55e88beb6a3b3), and [packages/graph-explorer/src/components/VertexSymbol/VertexSymbol.tsx](https://github.com/aws/graph-explorer/pull/2162/files#diff-fb263a7035fc74475de947ca00c6dc11da4eb0e78db89f38c9996966fe602e4b) — a11y disable comments.
5. [packages/graph-explorer/src/components/Graph/hooks/useInitCytoscape.tsx](https://github.com/aws/graph-explorer/pull/2162/files#diff-439adadb0dbc9b1e9f6fed42e35250c15ecc892d3ab02b1ffe46bf52b5de4039) — set-state-in-effect disable.
6. [packages/graph-explorer/src/connector/sparql/parseAndMapQuads.test.ts](https://github.com/aws/graph-explorer/pull/2162/files#diff-edb927e5651b51b47fefd8a787519eb777143cb3395a5d706fa6396e38e3077f) and [packages/graph-explorer/src/core/StateProvider/vertexStylesTransform.test.ts](https://github.com/aws/graph-explorer/pull/2162/files#diff-399fa236ed4f6001be5f68e65473224faab44b7277da1f64115cbe2f9854fc50) — throw-message and type-assertion fixes.

## Validation

- `pnpm checks` passes.
- `pnpm test` passes (2,667 tests).
- `pnpm check:types` passes.

## Related Issues

- Resolves #2150
- Related to #1850
- Related to #2149
- Related to #1753

## Check List


packages/graph-explorer/src/connector/sparql/parseAndMapQuads.test.ts
```diff
@@ -376,20 +413,31 @@ describe("parseAndMapQuads", () => {
...
+      expect(() => parseAndMapQuads(null)).toThrow(
+        new TypeError("Cannot read properties of null (reading 'code')"),
+      );
     });
 
     it("should handle undefined data", () => {
-      expect(() => parseAndMapQuads(undefined)).toThrow();
+      expect(() => parseAndMapQuads(undefined)).toThrow(
+        new TypeError("Cannot read properties of undefined (reading 'code')"),
+      );
```
kmcginnes, line 428: **R1 (Important, please fix in this PR)** These two assertions pin a bug as the contract.

`isErrorResponse` is:

```ts
// packages/graph-explorer/src/connector/utils/isErrorResponse.ts:3-5
const isErrorResponse = (value: any): value is ErrorResponse => {
  return !!(value.code && value.detailedMessage);
};
```

`value: any` means TypeScript never checks the `value.code` access, so `null` and `undefined` crash before any schema runs. That is what these two tests now assert, under names that say "should handle". Three consequences:

- The crash skips `parseAndMapQuads`'s own `logger.error("Failed to parse SPARQL JSON response", ...)`, so this is the one malformed-input path that produces no diagnostic. The sibling tests just below assert `logger.error` was called and these two cannot.
- `Cannot read properties of null (reading 'code')` is V8's wording, specified nowhere, and it has changed before (prior to Node 16 it was `Cannot read property 'code' of null`). This is the same brittleness 121777fc removed when it stopped hard-coding an fs errno.
- 14 production call sites across all three connectors feed this guard unvalidated response bodies, and `null` is valid JSON.

Since this is in scope for the PR:

1. Change the parameter to `value: unknown`. TypeScript then rejects `value.code` outright, so the guard has to do a real check and the compiler enforces it from then on rather than a reviewer remembering to.
2. Point these two tests at `validationErrorFor(sparqlResponseSchema(sparqlQuadBindingSchema), null)` like their siblings, and assert the `logger.error` call they can now make.
3. Add `null` and `undefined` cases to `isErrorResponse.test.ts`. It has four cases and neither of these, which is how this survived.

One thing you will hit at step 1 that is worth raising rather than papering over: `isErrorResponse.test.ts:6` passes `code: 400` as a number while `ErrorResponse` declares `code: string`. The `any` is currently hiding that mismatch. Say which one is right rather than making the guard accept both.

Collapsing the identical three-line ritual across all 14 call sites into one helper is the better long-term fix, and that is being tracked separately, so please do not take it on here.

packages/graph-explorer/src/utils/testing/errorInstances.ts
```diff
@@ -0,0 +1,44 @@
...
+  try {
+    await fn();
+  } catch (error) {
+    expect(error).toBeInstanceOf(z.ZodError);
+    expect((error as z.ZodError).issues).toEqual(expectedIssues);
+    return;
+  }
+
+  throw new Error("Expected function to throw a ZodError");
+}
```
kmcginnes: **R2 (Important)** `expectZodErrorFor` costs three production exports, and there is a built-in that costs none.

Its two callers required exporting `SerializedBackupSchema`, `graphExportPayloadSchema`, and `fileEnvelopeSchema`, the last into the `core/fileEnvelope` public barrel, so a schema whose only production use is a `safeParse` inside `parseFileEnvelope` is now part of that module's interface. AGENTS.md asks for deep modules with a simple interface.

The JSDoc's reasoning is also wrong in a way worth knowing. It is not that "`new ZodError` carries internal state". Zod 4's `parse`/`safeParse` construct `ZodRealError` (a `$constructor` built with `Parent: Error`) while `z.ZodError` is built with `Parent = Object`, and each instance stores its own constructor at `_zod.constr`. That one reference is what differs; `message`, `name`, `cause`, and even a copied `stack` all compare equal. The useful consequence is that two independently produced parse errors *do* deep-equal each other, so the whole helper collapses to one line:

```ts
await expect(fn()).rejects.toThrowError(schema.safeParse(data).error);
```

That deletes the helper but keeps the schema import. To drop the exports too, assert on the thrown error alone:

```ts
// localDb.test.ts, reverts the SerializedBackupSchema export
const error = await readBackupDataFromFile(blob).catch(e => e);
expect(error).toBeInstanceOf(ZodError);
expect(z.prettifyError(error)).toMatchInlineSnapshot();
```

```ts
// exportedGraph.test.ts:350, reverts the fileEnvelopeSchema barrel export
await expect(parseExportedGraph(toGraphFileBlob(malformed))).rejects.toMatchObject({
  message: "File does not have the expected envelope structure",
  issues: [{ code: "invalid_type", path: ["meta", "version"] }],
});
```

`prettifyError` for the two `ZodError` sites, `toMatchObject` for the `FileEnvelopeError` site. The split matters: `ZodError.issues` is non-enumerable and Vitest's pretty-printer only walks enumerable keys, so a failed `toMatchObject` against a `ZodError` renders as `ZodError { "message": "[ ...escaped JSON... ]" }` with no `issues`, giving a diff that does not line up with your expectation. `FileEnvelopeError.issues` is a plain class field and renders fine. All of this was verified against this branch with negative controls; tests pass and `pnpm checks` is clean.

`validationErrorFor` is fine and should stay. Its 16 call sites use `sparqlResponseSchema` and `sparqlQuadBindingSchema`, both already exported on main, so it widens nothing.

packages/graph-explorer/src/core/fileEnvelope/index.ts
```diff
@@ -2,6 +2,7 @@ export {
   createFileEnvelope,
   type EnvelopeExpectation,
   type EnvelopeVersion,
+  fileEnvelopeSchema,
```
kmcginnes: **R2 (Important)** This is the export I would most like to see reverted, since `index.ts` is what the rest of the app imports as `@/core/fileEnvelope`, so a test-only need has widened a public module interface. See the thread on `utils/testing/errorInstances.ts` for a `toMatchObject` form that gets the same assertion strength with no schema import.

packages/graph-explorer/src/core/StateProvider/localDb.ts
```diff
@@ -116,7 +116,7 @@ export interface LocalDb {
 }
 
 /** Serialized backup data structure schema. This allows validation and parsing of `unknown` data. */
-const SerializedBackupSchema = z.object({
+export const SerializedBackupSchema = z.object({
```
kmcginnes: **R2 (Important)** Exported only so a test can rebuild the expected error. See the thread on `utils/testing/errorInstances.ts`: `z.prettifyError` plus `toMatchInlineSnapshot` asserts the same failure without needing the schema.

packages/graph-explorer/src/modules/GraphViewer/exportedGraph.ts
```diff
@@ -45,7 +45,7 @@ export const GRAPH_EXPORT_VERSION = 1;
  */
 export const GRAPH_EXPORT_WIRE_VERSION: EnvelopeVersion = "1.0";
 
-const graphExportPayloadSchema = z.object({
+export const graphExportPayloadSchema = z.object({
```
kmcginnes: **R2 (Important)** Same as `localDb.ts`, exported only for a test. See the thread on `utils/testing/errorInstances.ts`.

.oxlintrc.json
```diff
@@ -25,6 +25,16 @@
...
+    "vitest/expect-expect": [
+      "error",
+      {
+        "assertFunctionNames": [
+          "expect",
+          "expectRoundTrip",
+          "expectZodErrorFor"
+        ]
+      }
+    ],
```
kmcginnes, line 33: **R3 (Nit)** `expectZodErrorFor` is `async` and does all of its asserting inside the returned promise, while `typescript/no-floating-promises` is `"off"` at line 41 of this file. A call written without `await` lints clean, satisfies `expect-expect` because the name is allowlisted, and lets the test body finish before any `expect()` runs. Both current call sites do await, so nothing is broken today. `expectRoundTrip` is synchronous and has no such hole. If R2 removes `expectZodErrorFor` this resolves itself; otherwise drop it from the list or enable `no-floating-promises` for `**/*.test.ts`.

For the record the config itself is sound: I checked and the four `vitest/expect-expect` reports really were false positives, since `expectRoundTrip` runs two real `expect` calls. Adding the rule at `"error"` where it was previously unconfigured is a net improvement over padding four tests with a redundant assertion.

packages/graph-explorer/src/connector/sparql/neighborCounts.test.ts
```diff
@@ -251,7 +259,17 @@ describe("neighborCounts", () => {
...
-    ).rejects.toThrow();
+    ).rejects.toThrow(
+      validationErrorFor(
+        sparqlResponseSchema(
+          z.object({
+            resource: sparqlResourceValueSchema,
+            totalCount: sparqlNumberValueSchema,
+          }),
+        ),
+        invalidTotalCountResponse,
```
kmcginnes: **R4 (Nit)** This rebuilds a module-private production schema in the test, so the assertion holds a second copy of the contract it is checking. `neighborCounts.ts:127-132` composes this inline and never exports it. Make `totalCount` optional in production and this copy keeps producing the old issues, failing with a diff that points at the wrong place. The other 14 `validationErrorFor` call sites import the real schema from `./types`, so this is the odd one out.

Either export the composed schema (lift it to a named module const) and import it, or drop this to a class-level check so no copy exists. That is the same widen-vs-loosen trade-off as R2, so the two are worth deciding together. Same thing at `rawQuery.test.ts:159`.

packages/graph-explorer/src/connector/sparql/rawQuery.test.ts
```diff
@@ -153,7 +156,12 @@ describe("rawQuery", () => {
 
     await expect(
       rawQuery(mockFetch, { query: "SELECT ?name WHERE { ?s ?p ?name }" }),
-    ).rejects.toThrow();
+    ).rejects.toThrow(
+      validationErrorFor(
+        sparqlResponseSchema(z.record(z.string(), sparqlValueSchema)),
+        mockResponse,
+      ),
+    );
```
kmcginnes: **R4 (Nit)** Same as the thread on `neighborCounts.test.ts:263`: this retypes `rawQueryResponseSchema` from `rawquery.ts:27-28`, which is module-private, rather than importing it.

packages/graph-explorer/src/components/Field.tsx
```diff
@@ -81,13 +81,16 @@ function Field({
...
+    // role="group" so screen readers expose the related label/control cluster.
+    /* eslint-disable jsx-a11y/prefer-tag-over-role */
     <div
       role="group"
       data-slot="field"
       data-orientation={orientation}
       className={cn(fieldVariants({ orientation }), className)}
       {...props}
-    />
+    /> /* eslint-enable jsx-a11y/prefer-tag-over-role */
```
kmcginnes: Two separate things on this one.

**R5 (Nit)** The block form works. I verified the span closes correctly, a probe `<div role="group" />` appended past the enable still errors and deleting the enable makes it stop, so this is not a correctness problem. It is consistency: the repo uses `eslint-disable-next-line` everywhere else, including on `useInitCytoscape.tsx` in this same PR. I tested the scoped form here and oxlint exits 0 with `oxfmt --check` clean, so nothing about the block form is required by the tooling. Moving the comment plus `// eslint-disable-next-line jsx-a11y/prefer-tag-over-role` inside the opening tag above `role="group"` also bounds the suppression to the one attribute instead of the element's whole subtree. Same in `EdgePreview.tsx` and `VertexSymbol.tsx`.

**R7 (Nit)** The rationale says `role="group"` makes screen readers "expose the related label/control cluster", but no call site names the group: `EdgeStyleDialog.tsx` (12), `NodeStyleDialog.tsx` (10), and `NodeExpandFilters.tsx` (2) all render a bare `<Field>` or `<Field className=...>`. An ARIA `group` with no accessible name conveys only a boundary, and the label-to-control association users rely on already comes from `FieldLabel`. Suppressing is still the right call, since `<fieldset>` without a `legend` is equally nameless and would break the `has-[>[data-slot=field]]` composition and pull in default styling. The problem is only that the comment claims a benefit that does not occur, and it is the only rationale a future reader will find. This is inherited shadcn behaviour, not something this PR introduced. Worth rewording to the real reason and dropping the screen-reader claim.

packages/graph-explorer/src/components/EdgePreview/EdgePreview.tsx
```diff
@@ -75,6 +75,9 @@ export function EdgePreview({
   );
 
   return (
+    // The preview is a composite of SVG arrows and a label; it cannot be an
+    // <img> because it has children, so the div carries role="img" instead.
+    /* eslint-disable jsx-a11y/prefer-tag-over-role */
```
kmcginnes: **R5 (Nit)** Same as the thread on `Field.tsx:84`: prefer `eslint-disable-next-line` on the `role` attribute over a block wrapping the element. The rationale comment itself is accurate here. `role="img"` with `aria-label` is the right pattern for a composite that cannot be an ``, and it correctly makes the drawn label text presentational rather than announcing it twice.

packages/graph-explorer/src/components/VertexSymbol/VertexSymbol.tsx
```diff
@@ -40,6 +40,9 @@ export function VertexSymbol({ vertexStyle, className }: Props) {
   const shapeEl = renderShape(geometry, insetSize, transform);
 
   return (
+    // <svg> is the correct vector-graphics element; role="img" is the
+    // standard ARIA pattern for exposing it as an image to assistive tech.
+    /* eslint-disable jsx-a11y/prefer-tag-over-role */
```
kmcginnes: **R5 (Nit)** Same as the thread on `Field.tsx:84`. The rationale is correct here too: `<svg role="img" aria-label>` is the canonical accessible-SVG pattern, and the rule's `` suggestion is wrong for an inline SVG that needs a `clipPath` and a nested `VertexSymbolIcon`.

packages/graph-explorer/src/connector/fetchDatabaseRequest.test.ts
```diff
@@ -460,7 +460,7 @@ describe("fetchDatabaseRequest", () => {
           method: "POST",
           signal: controller.signal,
         }),
-      ).rejects.toThrow();
+      ).rejects.toThrow(controller.signal.reason);
```
kmcginnes: **R6 (Nit)** `mockFetch.mockRejectedValue(controller.signal.reason)` is seven lines up, so this deep-equals the value against itself. The test is named for propagation and identity is what propagation means: `toBe` fails if the connector re-wraps the abort in a new object carrying the same message, which deep-equal lets through. Line 451 already uses `expect(caught.cause).toBe(error)`, so identity is this file's idiom. It also drops out of `require-to-throw-message` scope entirely.

```suggestion
      ).rejects.toBe(controller.signal.reason);
```

packages/graph-explorer/src/connector/queries/edgeConnectionsQuery.test.ts
```diff
@@ -322,7 +322,7 @@ describe("edgeConnectionsQuery", () => {
         signal: abortController.signal,
         meta: { store, explorer },
       } as any),
-    ).rejects.toThrow();
+    ).rejects.toThrow(abortController.signal.reason);
```
kmcginnes: **R6 (Nit)** Same as the thread on `fetchDatabaseRequest.test.ts:463`. `options?.signal?.throwIfAborted()` throws that same object, so identity is assertable and strictly stronger than deep equality here.

```suggestion
    ).rejects.toBe(abortController.signal.reason);
```

packages/graph-explorer/src/utils/testing/index.ts
```diff
@@ -1,5 +1,7 @@
 export * from "./createMockExplorer";
+export * from "./createCancelledError";
```
kmcginnes: **R11 (Nit)** The move sorted the one previously-unsorted entry but landed one position off. `createCancelledError` sorts before `createMockExplorer`.

```suggestion
export * from "./createCancelledError";
export * from "./createMockExplorer";
```

kmcginnes (changes requested):
Reviewed the toolchain bump end to end. It does what #2150 asked and the mechanics check out. One thing needs fixing before this lands; the rest is optional.

**Please fix in this PR:** R1, the two `TypeError` assertions in `parseAndMapQuads.test.ts`. They document a null-dereference bug in `isErrorResponse` as the intended contract, and that guard is shared by 14 call sites across all three connectors. Details in the inline thread, including why the smaller `value: unknown` change is the ask rather than the 14-site refactor.

**Worth your judgement:** R2, the three production `export`s that `expectZodErrorFor` needed. There is a built-in that gets the same assertion strength without them, with verified code in that thread.

Everything else, R3 through R11, is a nit. R6 and R11 have one-click suggestions.

### What I verified

- All three bumps land at the versions #2150 named, and the installed binaries report `1.82.0`, `0.67.0`, `7.0.2001`.
- All 25 `require-to-throw-message` sites match the issue's per-file counts, and the sites it said were correctly argument-less are untouched. There are 6 more `.not.toThrow()` calls in `allowed-db-origins.test.ts` that the issue's grep of 31 seems to have missed, also correctly untouched.
- The three `jsx-a11y/prefer-tag-over-role` suppressions are not scope creep. Running oxlint 1.82 against merge-base copies of those three components reproduces all three errors, so the issue's inventory of 30 undercounted the real 33.
- Type-aware linting still fires after the `oxlint-tsgolint` 0.23.0 to 7.0.2001 jump. A planted `x as string` still reports `no-unnecessary-type-assertion` and a planted async-without-await still reports `require-await`. The version jump is explained by `oxlint@1.82.0` declaring `peerDependencies: { "oxlint-tsgolint": ">=7.0.2001" }`.
- Lockfile churn is fully accounted for: identical added and removed name sets, 48 each, all three tools plus their platform bindings plus `tinypool` 2.1.0 to 2.1.2, which is oxfmt's only runtime dependency. No new or removed package names, no registry or resolution changes, and all 48 integrity hashes match registry.npmjs.org.
- `pnpm checks` and `pnpm test` pass (218 files, 2667 tests).
- The `useInitCytoscape` suppression names the right rule, `react/set-state-in-effect`, despite the config key reading `react-compiler/set-state-in-effect`. It also cannot loop, since the effect's deps are `[memoizedConfig, wrapper]` and `setCy` touches neither.

### One correction to an earlier read

The roughly 14 assertions that already had an argument and were rewritten to error-instance form are strictly *stronger*, not weaker. Vitest 4 deep-compares the thrown value including class, `name`, and own enumerable properties rather than only `message`, which I confirmed with probes that fail on extra-field, class, and `name` mismatch. So `toThrow(new FileEnvelopeError(msg, issues))` really does assert the class and the `issues` payload. Nothing to change there.

### Not blocking, and outside the code

- The description says the oxfmt specifier was "kept as `^0.67.0`". `package.json:39` actually pins it exactly as `0.67.0`, which is what #2150 asked for, so the code is right and the description is wrong on the one point the issue called out as deliberate. Worth correcting so a later reader does not reintroduce the caret.
- `docs/agents/testing.md` does not mention `errorInstances.ts` in its "Key helpers" list, and its line-70 rule ("assert the full error... deep-compares every property") is exactly the pattern that breaks for `ZodError`. Whatever survives R2 should get a line there, plus a note that a helper wrapping `expect()` needs registering in `.oxlintrc.json`'s `assertFunctionNames`.
- The branch carries a merge commit from main plus two "review feedback" commits, one of which partially reverts the other. Since this repo squash-merges, none of it survives, so this is only about the PR's own commit list being readable. Rebasing and squashing the fixups would tidy it.

Thanks for taking this on. The `toThrow` message work is a real improvement over suppressing the rule, and the `assertFunctionNames` call was the right one.

packages/graph-explorer/src/connector/sparql/parseAndMapQuads.test.ts
mjuarros: Fixed in 8e722d04. `isErrorResponse` now accepts `unknown` and checks for a non-null object before reading its fields. The SPARQL tests now expect schema validation errors and verify logging, and `isErrorResponse.test.ts` covers both `null` and `undefined`. I also corrected the fixture to use the declared string type for `code`.

packages/graph-explorer/src/utils/testing/errorInstances.ts
mjuarros: Addressed in 91393765. I removed `expectZodErrorFor`, switched the two Zod assertions to `z.prettifyError` inline snapshots, and used `toMatchObject` for `FileEnvelopeError`. This also let me remove all three test-only schema exports.

packages/graph-explorer/src/core/fileEnvelope/index.ts
mjuarros: Reverted in 91393765. `fileEnvelopeSchema` is module-private again and has been removed from the public barrel. The test now asserts the `FileEnvelopeError` shape directly with `toMatchObject`.

packages/graph-explorer/src/core/StateProvider/localDb.ts
mjuarros: Reverted in 91393765. `SerializedBackupSchema` is private again, and the test now checks the thrown `ZodError` using `z.prettifyError` and an inline snapshot.

packages/graph-explorer/src/modules/GraphViewer/exportedGraph.ts
mjuarros: Reverted in 91393765. `graphExportPayloadSchema` is module-private again, with the test asserting the thrown `ZodError` via `z.prettifyError` and an inline snapshot.

.oxlintrc.json
mjuarros: Resolved as part of 91393765. `expectZodErrorFor` was removed, along with its `assertFunctionNames` entry, so the unawaited-helper risk no longer applies.

packages/graph-explorer/src/connector/sparql/neighborCounts.test.ts
mjuarros: Addressed in 2b7d2399. I chose the class-level assertion so the test no longer duplicates or exports the module-private schema. It now verifies that malformed total-count responses reject with `ValidationError`.

packages/graph-explorer/src/connector/sparql/rawQuery.test.ts
mjuarros: Addressed in 2b7d2399. This now uses a class-level `ValidationError` assertion instead of duplicating the private raw-query response schema.

packages/graph-explorer/src/components/Field.tsx
mjuarros: Addressed both points in 2b7d2399. The suppression is now scoped with `eslint-disable-next-line` directly above the `role` attribute, and the rationale now explains the layout/default-styling constraint without claiming the unnamed group is announced as a labeled cluster.

packages/graph-explorer/src/components/EdgePreview/EdgePreview.tsx
mjuarros: Updated in 2b7d2399. The suppression is now an `eslint-disable-next-line` immediately above `role="img"`, limiting it to that attribute.

packages/graph-explorer/src/components/VertexSymbol/VertexSymbol.tsx
mjuarros: Updated in 2b7d2399. The suppression is now scoped directly to the `role="img"` attribute, with the rationale focused on the inline SVG requirements.

packages/graph-explorer/src/connector/fetchDatabaseRequest.test.ts
mjuarros: Applied in 2b7d2399. The test now uses `rejects.toBe(controller.signal.reason)` to verify that the exact abort reason is propagated.

packages/graph-explorer/src/connector/queries/edgeConnectionsQuery.test.ts
mjuarros: Applied in 2b7d2399. This now uses `rejects.toBe(abortController.signal.reason)` to assert identity rather than deep equality.

packages/graph-explorer/src/utils/testing/index.ts
mjuarros: Fixed in 2b7d2399. `createCancelledError` now sorts before `createMockExplorer`.

kmcginnes (approved):

Issue #1853, opened 2026-06-18
Spike: intermittent flake in proxy-server POST /gremlin "sends query as JSON with gremlin key" test


kmcginnes:
## Summary

`packages/graph-explorer-proxy-server/src/app.test.ts` > `createApp` > `POST /gremlin` > `sends query as JSON with gremlin key` failed once during a full `pnpm test` run (`1 failed | 1807 passed`), then could not be reproduced across 35+ subsequent runs (12× the file in isolation, 3× full suite, 20× under deliberate CPU contention from a parallel graph-explorer run). This spike is to reproduce it reliably, confirm the root cause, and decide on a fix — not to patch blind.

## Why it is a flake, not a regression

The package depends only on `@graph-explorer/shared` and runs as its own threaded vitest project; nothing in its `src/` imports the `graph-explorer` app package. It was first observed while reviewing an unrelated graph-explorer persistence change (#1831), which has no causal path to this test. The failure pre-dates and is independent of that work.

## Leading hypothesis (unconfirmed)

The `/gremlin` handler (`app.ts:374-433`) registers `req.on("close")` / `res.on("close")` listeners that call `cancelQuery()` — which issues an additional `mockFetch` call — when the connection closes before `req.complete` / `res.writableFinished`. The response body is streamed via `pipeline()` (`app.ts:264`), which settles asynchronously *after* the handler returns and possibly after supertest has resolved. Combined with a module-shared `mockFetch`, `mockReset()` in `beforeEach`, and `restoreMocks: true`, there is a window where a late close-event `mockFetch` from one test can bleed into another test's expectations.

Caveat: this best explains a stray `calls[1]`, whereas the observed failure asserted on `calls[0][1].body`. The mechanism is plausible but not proven — hence a spike.

## Suggested approach

1. Reproduce reliably — try `vitest --no-isolate`, `--sequence.seed`, tighter stress loops, and CI hardware; instrument the close/cancel path and the mock call log.
2. Confirm the exact mechanism (close-event race vs. shared-mock teardown vs. something else).
3. Fix with certainty. Likely candidates regardless of exact trigger:
   - Per-test mock instance instead of a module-shared `mockFetch`.
   - Assert on the *specific* fetch call (URL/method) rather than positional `calls[0]`.
   - Ensure the handler's streaming/close path has settled before assertions run.

## Acceptance criteria

[...]

PR #2166, opened 2026-09-11
Update Vitest to version 5

### LINKED ISSUE #2151: Update vitest to 5
`vitest` and `@vitest/coverage-v8` stayed on 4.1.11 in the latest dependency sweep. Both must move
together (5.0.0).

## Blockers found

**jest-dom matcher types no longer resolve** — 172 `error TS2339` across the test suite:
`toBeInTheDocument`, `toBeDisabled`, `toBeChecked`, `toHaveAttribute`, `toHaveTextContent`, and
`toHaveValue` all report "does not exist on type `Assertion<void, HTMLElement>`". vitest 5 changed the
`Assertion` type, so `@testing-library/jest-dom`'s module augmentation no longer merges in. Needs the
setup file and/or `tsconfig` `types` wiring updated.

**Order-dependent test failure surfaced** — `packages/graph-explorer-proxy-server/src/app.test.ts:311`
intermittently expects status 400 and receives 200 during a full-suite run, while passing when the
file runs alone. vitest 5 reuses workers across files by default, so state that used to be isolated
per file now leaks. It reproduced once and passed on two subsequent full runs, so it's intermittent —
worth tracking down rather than waiting for it to bite CI.

## Acceptance

`pnpm checks` and `pnpm test` pass with vitest and `@vitest/coverage-v8` at 5.x, and the proxy server
app tests pass repeatably in a full-suite run.

<audience-footer>

> [!IMPORTANT]

</audience-footer>

mjuarros:
## Description

- Upgrade Vitest and its V8 coverage provider together to version 5 across every workspace.
- Use the official jest-dom Vitest integration so custom matcher types remain available with Vitest 5.
- Absorb the two v5 behavior changes that affect this repo: the `-t` test-name separator and the `clearMocks` default.

## How to read

1. [packages/graph-explorer/src/setupTests.ts](https://github.com/aws/graph-explorer/pull/2166/files#diff-da067bb0b6f92b46e5cb3ebab37cd9809fabd01fbeb5321b95d44c951f34194b) — Vitest-specific matcher registration.
2. [packages/graph-explorer/tsconfig.json](https://github.com/aws/graph-explorer/pull/2166/files#diff-9a123d9c986e6d49b25941b73487100452414c4e1b2f4b80a8b7ef062f3c2f5c) — Vitest-specific matcher type declarations.
3. [pnpm-lock.yaml](https://github.com/aws/graph-explorer/pull/2166/files#diff-32824c984905bb02bc7ffcef96a77addd1f1602cff71a11fbbfdd7f53ee026bb) — regenerated Vitest 5 dependency graph.

Both halves of the jest-dom change are load-bearing. The import registers the matchers at runtime; the `tsconfig` `types` entry is what puts the `vitest.Matchers` augmentation into the program. Dropping either brings back the 172 `TS2339` errors from #2151.

## Behavior changes absorbed

**`-t` now matches the reporter-style name.** Vitest 5 joins suite and test name with a spaced `>` rather than a space, so `-t "NetworkError is instanceof NetworkError"` silently matches nothing and still exits 0. `AGENTS.md` documented the old form, so it is updated here. A single segment such as `-t "is instanceof NetworkError"` is separator-agnostic and still works.

**`clearMocks` defaults to `true`.** `packages/graph-explorer/vite.config.ts` already pins it, but `packages/graph-explorer-proxy-server/vitest.config.ts` was inheriting the flip silently, so it is now pinned to the v5 default. `restoreMocks: true` does not cover this: `restoreAllMocks()` only touches `vi.spyOn` registrations, while `clearAllMocks()` touches every mock including standalone `vi.fn()`. No test depended on cumulative mock history, and the package has no `beforeAll`.

Isolation did not change. Both 4.1.11 and 5.0.0 default `isolate: true`; #2151's claim that v5 reuses workers across files is incorrect and is corrected on the issue.

## Incidental fixes

- `pnpm-lock.yaml` is regenerated against pnpm 12.4.2 after rebasing onto #2178.
- The pre-commit hook failed on any commit whose only JS/TS file is covered by an oxlint `ignorePatterns` entry, which is the case for a `*.config.ts`-only commit. `oxlint --fix` now gets the `--no-error-on-unmatched-pattern` guard the sibling `oxfmt` entry already had.
- Removed two dead references to the root `vitest.workspace.ts`, deleted back in #1027 when the repo moved to `projects`: the oxlint ignore pattern and the VS Code `vitest.workspaceConfig` setting.

## Validation

- [x] `pnpm install --frozen-lockfile` — clean on pnpm 12.4.2
- [x] `pnpm checks`
- [x] `pnpm test` — 220 files and 2,672 tests
- [x] `pnpm coverage`
- [x] Verified with Node 24.21.0 and Vitest 5.0.0

## Related Issues

- Fixes #2151

### Check List


kmcginnes (commented):
The upgrade is clean and I verified it independently: `pnpm checks`, `pnpm test` at 220 files / 2,672 tests, and `pnpm coverage` all pass.

The jest-dom rewiring is the documented fix, not a workaround. The [Vitest 5 migration guide](https://vitest.dev/guide/migration/), under "Assertion Types Expose Return and Received Types", says "Vitest no longer reads custom matcher declarations from the global `jest.Matchers` interface" and tells libraries supporting both Jest and Vitest to augment `jest.Matchers` and `vitest.Matchers` separately. jest-dom's root entry only references `jest.d.ts`, which is exactly why it broke, and the `./vitest` subpath augments `vitest.Matchers`. Both halves of your change are load-bearing: the import registers the matchers at runtime, and the `tsconfig` `types` entry is what puts the augmentation into the program. Dropping either brings the 172 errors back.

Three things before this goes in.

**Rebase onto main.** [Update pnpm to 12.4.2 (#2178)](https://github.com/aws/graph-explorer/pull/2178) landed after you branched. It clears the three high advisories `pnpm audit` currently reports against the `pnpm@11.9.0` pin, and nothing else is needed for them.

**Drop the isolation claim from the description.** "Preserve per-file isolation" describes work that was never needed. Both 4.1.11 and 5.0.0 default `isolate: true`; I unpacked both packages and diffed `configDefaults` to confirm. Neither the migration guide nor the [v5.0.0 release notes](https://github.com/vitest-dev/vitest/releases/tag/v5.0.0) mention an isolation change. What v5 added is a hint printed only when `isolate` is not set explicitly:

```
at least ~9.85s faster with isolate: false — reuses workers across files instead of one per file
```

That is almost certainly where [Update vitest to 5 (#2151)](https://github.com/aws/graph-explorer/issues/2151)'s "vitest 5 reuses workers across files by default" came from. Worth correcting on the issue too, so nobody chases a change that never happened.

**Optional nit.** The migration guide's "`clearMocks` is Enabled by Default" section flips that option from `false` to `true`. `packages/graph-explorer/vite.config.ts:75` already pins it, but `packages/graph-explorer-proxy-server/vitest.config.ts` now inherits the new value silently. Nothing is affected: the guide says the exposure is tests recording mock calls outside the test body, and this repo has no `beforeAll` calls anywhere and no module-scope `vi.fn()` declarations. Pinning it just keeps the config honest.

Two things I checked that are not your problem:

The flake #2151 described is pre-existing. It reproduces deterministically at your merge-base and fails identically on 4.1.11 and 5.0.0. Across 150 runs of the proxy package on each version I saw 2 failures on v4 and 0 on v5, so if anything the upgrade makes it rarer. Root cause is an unawaited `pipeline` in `app.ts` dating to April, written up in [Unawaited response pipeline leaks a cancelQuery fetch into the next test's mock history (#2190)](https://github.com/aws/graph-explorer/issues/2190).

I also checked this branch against the rest of the v5 breaking-change list and found nothing else live. The one worth naming is `toHaveTextContent`, which v5 makes strict and stops accepting RegExp: harmless here, because Vitest 5 does not ship that matcher at all and jest-dom still owns all 16 call sites. `PersistenceStatusIndicator.test.tsx:69` asserts `/failed writes/i` and passes, which is the proof. Separately, two coverage-config problems predating this branch are in [Project-level coverage.exclude is inert, so 56 icon components are measured (#2191)](https://github.com/aws/graph-explorer/issues/2191); one of them is why `pnpm coverage` dirties your tree by rewriting the `thresholds` block.

kmcginnes (commented):
Two bits of Vitest housekeeping I think belong in this PR rather than a follow-up: it is the change that makes the first one wrong, and the natural place to sweep the second.

**`AGENTS.md:76` needs the `>` separator.** Vitest 5 matches `-t` against the full test name joined with ` > `, the same string the reporter prints, rather than a space-joined concatenation. So `-t "NetworkError is instanceof NetworkError"` now matches nothing where it used to work, and it fails silently:

```
vitest run <file> -t "NetworkError is instanceof NetworkError"
→ Test Files 1 skipped (1) / Tests 18 skipped (18), exit code 0
```

Green summary, zero tests run, no indication the filter matched nothing. Since [Update pnpm to 12.4.2 (#2178)](https://github.com/aws/graph-explorer/pull/2178) already fixed the `--` half of that line, this upgrade is the only thing still making it inaccurate. Suggested replacement:

```
- `pnpm test -t "suite > name"` — test by name, joined with ` > ` (no `--` separator; pnpm forwards args verbatim, unlike npm)
```

A single segment such as `-t "renders empty state"` still works and is separator-agnostic, so it may be worth saying that too.

**`.oxlintrc.json:23` can go.** `"**/vitest.workspace.ts"` excludes a file that has never existed in this repo; `git log --all` for that path is empty and `find` turns up nothing. `vitest.workspace.ts` was Vitest's pre-`projects` way of declaring a multi-project workspace, and we use `projects: ["packages/*"]` at `vitest.config.ts:5` instead. It is not redundant with the `**/*.config.{js,ts,mjs}` entry above it either, since `.workspace.ts` does not match `.config.ts`, so it is an independent line pointing at nothing.

Zero lint impact, but it implies we have or had a Vitest workspace file, which sends the next reader looking for one.

Both optional. Say if you would rather keep this PR to the version bump and I will take them separately.

mjuarros:
Thanks — worked through all five points. All valid, and I've taken all of them. Two have supporting evidence worth correcting, and one is slightly bigger than described.

Branch is rebased onto `main` and now carries three additional commits. `pnpm install --frozen-lockfile` is clean on pnpm 12.4.2, `pnpm checks` passes, `pnpm test` is 220 files / 2,672 tests, `pnpm coverage` passes.

## Taken as described

**Rebase.** Done. The lockfile auto-merged with no conflict, which I didn't want to take on faith — `pnpm install --frozen-lockfile` reports "Lockfile is up to date" under pnpm 12.4.2, and the supply-chain policy check passes across all 891 entries.

**Isolation claim.** Dropped from the description, and you're right about the cause. I confirmed the defaults by unpacking both packages and reading `configDefaults` directly:

| | `isolate` | `clearMocks` |
|---|---|---|
| 4.1.11 | `true` | `false` |
| 5.0.0 | `true` | `true` |

The reporter hint is also exactly where #2151's claim came from — it prints on every run of this suite:

```
Isolate  220 workers spawned · ~297ms startup each (spawn + environment, per file)
         at least ~6.23s faster with isolate: false — reuses workers across files instead of one per file
```

Correcting #2151 separately.

**`-t` separator.** Reproduced in both directions, and confirmed it is genuinely new in v5 rather than pre-existing:

```
v5.0.0  -t "NetworkError is instanceof NetworkError"    → 18 skipped, exit 0
v5.0.0  -t "NetworkError > is instanceof NetworkError"  → 1 passed
v5.0.0  -t "is instanceof NetworkError"                 → 1 passed
v4.1.11 -t "NetworkError is instanceof NetworkError"    → 1 passed
```

Took your wording with one adjustment: `oxfmt` strips the spaces inside an inline code span, so `` ` > ` `` normalizes to `` `>` `` and loses the point. Reworded to "joined by a spaced `>` exactly as the reporter prints them" so it survives formatting, and added the single-segment note. `AGENTS.md:76` was the only occurrence — `docs/agents/testing.md` doesn't document `-t`.

## Two corrections

**`clearMocks` is a real behavior change for the proxy package, not a no-op.** `restoreMocks: true` does not already cover it. In `@vitest/spy`, `restoreAllMocks()` iterates `MOCK_RESTORE`, which only holds `vi.spyOn` registrations, whereas `clearAllMocks()` iterates `DIRTY_MOCK_STATES`, which holds every mock including standalone `vi.fn()`. So the package genuinely gains per-test clearing it didn't have before.

Your conclusion still holds — nothing breaks, and I confirmed `beforeAll` has zero usages anywhere in `packages/`. One detail though: `packages/graph-explorer-proxy-server/src/test-setup.ts:25` *is* a module-scope `vi.fn()` (the `node-fetch` default inside the `vi.mock` factory). Harmless, because its history is never asserted cumulatively, but it does exist. Pinned to `true` to adopt the v5 default and match `vite.config.ts`, rather than freezing v4 behavior.

**`vitest.workspace.ts` did exist.** `git log --all` for that path isn't empty — the file lived at the repo root for about ten months:

- `4176073c` — Add Vitest to proxy server (#555), 2024-08-26 — added
- `90f7c264` — Update dependencies (#614), 2024-09-13 — modified
- `76103d4e` — Migrate vitest workspaces (#1027), 2025-06-19 — deleted

So the line isn't pointing at a file we never had; it's leftover from the #1027 migration to `projects`. That inverts the rationale but not the conclusion — and it makes the sweep more worthwhile than it looked, because there was a second dangling reference: `.vscode/settings.json:8` still set `"vitest.workspaceConfig": "vitest.workspace.ts"`. Removed both.

## One thing I hit that wasn't on your list

Committing the `clearMocks` change on its own broke the pre-commit hook:

```
✖ oxlint --fix:
No files found to lint. Please check your paths and ignore patterns.
husky - pre-commit script failed (code 1)
```

`vitest.config.ts` is the only JS/TS file in that commit and it's covered by the `**/*.config.{js,ts,mjs}` ignore pattern, so `oxlint` selects nothing and exits non-zero. Any `*.config.ts`-only commit hits this. The `oxfmt` entry in the same `lint-staged` block already had `--no-error-on-unmatched-pattern`; the `oxlint --fix` entry didn't. Added it as its own commit. Happy to pull that out if you'd rather keep this PR narrower.

Also confirmed #2191's threshold churn first-hand — `pnpm coverage` rewrote the `thresholds` block in `vitest.config.ts` (statements 65→70, branches 44→53, functions 58→63, lines 73→78). Reverted; left out of this PR.

kmcginnes (approved):
All five items addressed. Verified on `27c9ec70`: `pnpm install --frozen-lockfile`, `pnpm checks`, and `pnpm test` (220 files / 2,672 tests) all pass, tree clean, rebased onto `959b62e7` with 0 commits behind main and the pin at `pnpm@12.4.2`.

Two places you corrected me, both worth recording:

`restoreMocks: true` does not subsume `clearMocks`. I had it the other way round. `restoreAllMocks()` only touches `vi.spyOn` registrations while `clearAllMocks()` touches every mock including standalone `vi.fn()`, so pinning it in the proxy config was the right call rather than a redundant one.

The second `vitest.workspace.ts` reference in `.vscode/settings.json` I missed entirely. I only found the oxlint ignore pattern.

The fourth commit is not scope creep, for anyone reading later. Pinning `clearMocks` meant editing `vitest.config.ts`, which matches the `**/*.config.{js,ts,mjs}` entry in oxlint's `ignorePatterns`, so a config-only commit handed `oxlint --fix` nothing lintable and it exited non-zero. The `--no-error-on-unmatched-pattern` guard matches what the sibling `oxfmt` entry already had.

Pre-existing work split out and not blocking this: [#2190](https://github.com/aws/graph-explorer/issues/2190) for the proxy test race, which also supersedes the older spike in #1853, and [#2191](https://github.com/aws/graph-explorer/issues/2191) for the inert project-level `coverage.exclude` and the threshold drift.

PR #2175, opened 2026-09-14
Update jotai to 3

### LINKED ISSUE #2152: Update jotai to 3
`jotai` stayed on ^2.20.1 in the latest dependency sweep. 3.0.0 breaks the build in two ways.
(`jotai-family` is already on ^1.1.0 and needs no further change.)

## Blockers found

**Removed subpath** — `packages/graph-explorer/src/core/queryClient.ts:1` imports from
`jotai/vanilla/store`, which no longer exists in v3: `error TS2307: Cannot find module
'jotai/vanilla/store'`. Needs the v3 equivalent for whatever store type it pulls in.

**Inference regressions** — 6 × `error TS7006` implicit `any` on callback parameters, all in the
bulk/search query modules:

- `connector/queries/bulkEdgeDetailsQuery.ts:59, :64`
- `connector/queries/bulkVertexDetailsQuery.ts:59, :64`
- `connector/queries/bulkNeighborCountsQuery.ts:53`
- `connector/queries/searchQuery.ts:34`

These are likely downstream of the store type import failing rather than six independent problems —
fix the import first and re-measure.

## Acceptance

`pnpm checks` and `pnpm test` pass with jotai at 3.x.

<audience-footer>

> [!IMPORTANT]

</audience-footer>

mjuarros:
## Description

Bumps `jotai` from `^2.20.1` to `^3.0.0`, adapts the places the v3 release breaks, and fixes one production regression the bump exposed.

The behavior change that drives everything here: v2 always re-rendered once right after mount, even without a value change. v3 drops that extra render, so `useAtomValue` only re-renders on an actual change — and a store write that lands between a component's first render and its subscribing effect is now missed until the next change.

Core changes:

- Swap the removed `jotai/vanilla/store` import for the local `AppStore` type alias in `queryClient.ts`.
- Move the `DataExplorer` redirect into `useEffect`. react-router sets its `activeRef` in a layout effect, so the render-body `navigate` was already a no-op on the first render; v2's forced mount re-render supplied the second render that made it fire. Without this the redirect would have silently stopped working under v3.
- Subscribe to the schema before schema sync mounts in `SchemaDiscoveryBoundary`. See below.
- Add a `flushPendingAtomUpdates()` test helper in `renderHookWithJotai.tsx` and apply it across `useSchemaSync.test.ts` to keep the now-separate post-mount render inside `act()`.

Also swaps `.values().toArray()` for `[...vtConfigs.values()]`, which needs no iterator-helpers polyfill.

## The regression

`SchemaDiscoveryBoundary` called `useSchemaSync()` and then read the schema on the next two lines with `useHasActiveSchema()` and `useMaybeActiveSchema()`. TanStack Query subscribes through `useSyncExternalStore`, which React interleaves with `useEffect` by hook declaration order, so the mount fetch fired at the first hook's position — and `edgeConnectionsQuery.ts` writes `schemaAtom` synchronously in its zero-edge-types early return, before any `await`. The write landed before the two readers subscribed, and v2's forced render is what used to cover it.

Symptom: on a vertices-only graph, Schema Explorer showed "No Edge Connections Available" with a Synchronize button, on a graph with nothing to synchronize. It self-healed on that button or on navigating away and back.

Moving the two schema reads above `useSchemaSync()` fixes it. The ordering is load-bearing, so it carries a comment.

`SchemaDiscoveryBoundary.test.tsx` mocks all four of its hooks, so it cannot observe this by construction. `SchemaDiscoveryBoundary.integration.test.tsx` renders against the real Jotai store and query client instead; it fails 3/3 runs without the reorder and passes 3/3 with it.

## How to read

1. [packages/graph-explorer/package.json](https://github.com/aws/graph-explorer/pull/2175/files#diff-b366270b56b4904f97446575d6ff470b09574663747edd1b0dcd1d7582d89c4e) — dependency bump
2. [packages/graph-explorer/src/core/queryClient.ts](https://github.com/aws/graph-explorer/pull/2175/files#diff-ff19930233c6604faef6c676f62c6283c49e5edc8af3fce7e7afb28ded3aa694) — store type substitution
3. `packages/graph-explorer/src/components/SchemaDiscoveryBoundary.tsx` — hook reorder, the regression fix
4. `packages/graph-explorer/src/components/SchemaDiscoveryBoundary.integration.test.tsx` — regression coverage against the real store
5. [packages/graph-explorer/src/utils/testing/renderHookWithJotai.tsx](https://github.com/aws/graph-explorer/pull/2175/files#diff-ebad835e55ab79f2440ad5a7ae3dafe7dcf9a43cbd397766e26308c3a5066d1f) — new test helper for jotai v3 timing
6. [packages/graph-explorer/src/hooks/useSchemaSync.test.ts](https://github.com/aws/graph-explorer/pull/2175/files#diff-232eb33c373eafbf69ff2e371ed03068df5fe56cdd6b97caa0f121723ccad4d9) — test stabilizations
7. [packages/graph-explorer/src/routes/DataExplorer/DataExplorer.tsx](https://github.com/aws/graph-explorer/pull/2175/files#diff-c840af1b5b4ca5a9c9bc84a2ae23d0df39fa1f7d2c2e62b1bdbf5d36ee89a6c6) — redirect-in-effect and Map cleanup

## Notes

**The six `TS7006` errors from #2152 are gone, with no suppressions.** They were downstream of the `jotai/vanilla/store` import failing, exactly as the issue predicted, which is why none of the four files appear in this diff: `bulkEdgeDetailsQuery.ts`, `bulkVertexDetailsQuery.ts`, `bulkNeighborCountsQuery.ts`, and `searchQuery.ts` are untouched and contain no `ts-expect-error`, `ts-ignore`, or `any`. Their callback parameters infer cleanly once the store type resolves.

**The `useIsSyncing` assertion became a `waitFor`.** `expect(result.current.syncing).toBe(true)` was a synchronous read. Under v3 the syncing flag only goes truthy after an effect flush, so the assertion now waits rather than reading the first render.

## Validation

- `pnpm checks` passes with no errors.
- `pnpm test` passes with no failures (2673/2673 across 221 files), with no `act()` warnings.
- Rebased onto main, which picked up pnpm 12.4.2 (#2178) and Vitest 5 (#2166). The lockfile is regenerated from main's with `corepack pnpm install`, so the stray peer suffixes are gone and `pnpm install --frozen-lockfile` is clean.

## Related Issues

- Closes #2152
- Related to #2149

### Check List


kmcginnes (changes requested):
Verified this against the [v3 migration guide](https://github.com/pmndrs/jotai/blob/main/docs/guides/migrating-to-v3.mdx) and the migration is complete. `atomFamily` already comes from `jotai-family` at all 9 sites, and `loadable`, `setSelf`, the `delay` option, and `jotai/babel` are unused. The four `jotai/utils` APIs we still use are untouched by v3, and no deep imports remain outside the 7 supported entry points. So `jotai/vanilla/store` really was the only breakage, as [Update jotai to 3 (#2152)](https://github.com/aws/graph-explorer/issues/2152) predicted. `pnpm checks` and 2670/2670 pass for me too, and `useSchemaSync.test.ts` was clean across 10 consecutive runs.

Moving the redirect into an effect is right, and worth saying why since it reads like a style change: react-router sets its `activeRef` in a layout effect, so the render-body `navigate` was a no-op on the first render, and v2's forced mount re-render supplied the second render that made it fire. v3 removes that, so the old code would have silently stopped redirecting. Swapping `.values().toArray()` for `[...vtConfigs.values()]` is also the right direction, since the spread needs no iterator-helpers polyfill and we're planning to drop core-js.

Two things to fix, and the first caused the second.

## The helper's comment describes a mechanism that does not exist

`flushPendingAtomUpdates` says v3 "notifies `useAtomValue` subscribers on mount via a microtask." There is no microtask in jotai 3.0.0. `storeSub` calls `flushCallbacks` inline and that's a synchronous `do` loop, and grepping `dist/vanilla` and `dist/react` for `queueMicrotask`, `Promise.resolve().then`, and `setTimeout` finds only `atomWithObservable`'s unsubscribe timer.

The real mechanism is in `useAtomValueRaw.js`, which v3's `useAtomValue` delegates to. It seeds from `store.get(atom)` during render, subscribes with `useEffect(() => store.sub(atom, rerender))`, and never re-reads after subscribing. jotai's `docs/core/use-atom.mdx` describes the change in words worth lifting: v2 "always re-rendered once right after mounting, even without a value change," and in v3 "that extra render is gone: `useAtomValue` only re-renders on an actual change." Please reword to that and drop "microtask" from the summary and body.

## That misreading hid a production regression

The v2 forced render existed to catch writes landing between first render and subscription. jotai says v3 "can miss that specific kind of update until the next one." We have one:

`SchemaDiscoveryBoundary` calls `useSchemaSync()` and then reads the schema on the next two lines with `useHasActiveSchema()` and `useMaybeActiveSchema()`. TanStack Query subscribes through `useSyncExternalStore`, which React interleaves with `useEffect` by hook declaration order, so the mount fetch fires at the first hook's position. And `edgeConnectionsQuery.ts:49` writes `schemaAtom` synchronously in the zero-edge-types early return, before any `await`. The write lands before the two readers subscribe, and v2's forced render is what used to cover it.

The tell is that `useSchemaSync` reads the same atom internally and does get the update, because it's declared earlier. Two reads of one atom disagreeing inside one component.

Symptom: on a vertices-only graph, Schema Explorer shows "No Edge Connections Available" with a Synchronize button, on a graph with nothing to synchronize. It self-heals on that button or on navigating away and back, which is why it slipped through.

Moving the two schema reads above `useSchemaSync()` fixes it and kills the class for that component. `pnpm test SchemaDiscoveryBoundary` and `pnpm test useSchemaSync` both stay green with the reorder.

Here's a regression test. It fails 3/3 runs on this branch and passes 3/3 with the reorder. `SchemaDiscoveryBoundary.test.tsx` mocks all four of its hooks, so the existing suite can't see this by construction, which is the part I find most worth fixing.

```tsx
// @vitest-environment happy-dom
import { render, screen, waitFor } from "@testing-library/react";
import { MemoryRouter } from "react-router";
import { beforeEach, describe, expect, test } from "vitest";

import { getAppStore, schemaAtom } from "@/core";
import { createQueryClient } from "@/core/queryClient";
import {
  DbState,
  FakeExplorer,
  flushPendingAtomUpdates,
  TestProvider,
} from "@/utils/testing";

import { SchemaDiscoveryBoundary } from "./SchemaDiscoveryBoundary";

/**
 * These tests render against the real Jotai store and query client rather than
 * mocking the schema hooks, so they observe how the boundary reacts to the
 * store write that edge connection discovery performs.
 */
describe("SchemaDiscoveryBoundary against the real store", () => {
  let explorer: FakeExplorer;

  beforeEach(() => {
    explorer = new FakeExplorer();
  });

  function renderBoundary(state: DbState) {
    const store = getAppStore();
    state.applyTo(store);

    return render(
      <TestProvider client={createQueryClient()} store={store}>
        <MemoryRouter>
          <SchemaDiscoveryBoundary requireEdgeConnections>
            <div>Children</div>
          </SchemaDiscoveryBoundary>
        </MemoryRouter>
      </TestProvider>,
    );
  }

  test("renders children once discovery resolves zero edge types to an empty edge connection list", async () => {
    const state = new DbState(explorer);
    state.activeSchema.edges = [];
    state.activeSchema.edgeConnections = undefined;

    renderBoundary(state);

    const store = getAppStore();
    await waitFor(() => {
      expect(
        store.get(schemaAtom).get(state.activeConfig.id)?.edgeConnections,
      ).toStrictEqual([]);
    });
    await waitFor(() => {
      expect(screen.queryByText("Synchronizing...")).not.toBeInTheDocument();
    });
    await flushPendingAtomUpdates();

    expect(screen.getByText("Children")).toBeInTheDocument();
    expect(screen.queryByText(/Available$/)).not.toBeInTheDocument();
  });
});
```

The first `waitFor` is what makes the failure diagnostic rather than ambiguous: it proves the store already holds `[]`, so a subsequent empty state can only be a stale subscriber. The second one is needed or the assertion fires during the loading branch and fails for the wrong reason.

## Smaller

- Line 486's flush sits after the final `waitFor`, so it does nothing.
- The lockfile picked up a `(supports-color@7.2.0)` peer suffix on five babel packages at unchanged versions, which reads as a different pnpm than the pinned 11.9.0. You're also one commit behind main, and that commit ([Move core-js to production dependencies (#2167)](https://github.com/aws/graph-explorer/pull/2167)) touches the same two files. A rebase plus `corepack pnpm install` settles both.

Two notes for the description. [#2152](https://github.com/aws/graph-explorer/issues/2152) asked you to re-measure the six `TS7006` errors; I checked and they're genuinely gone with no suppressions, but say so, since those four files aren't in the diff. And converting the `useIsSyncing` assertion to `waitFor` deserves a line, since the syncing indicator now goes truthy only after an effect flush.

Filing the rest separately so this stays a dependency bump: reworking how Data Explorer resolves its node type, where the test settle seam belongs plus documenting it, a `react.md` rule about side effects in a render body, and a sweep for other components with this same hook-order hazard.

mjuarros:
Thanks — the regression catch was the valuable part, and you were right that the wrong comment is what hid it. All four items are addressed. Branch is rebased, so the SHAs changed.

## The comment

Reworded to describe the real mechanism, no microtask. I confirmed your reading independently: the only `setTimeout`/`queueMicrotask`/`Promise.resolve().then` anywhere in jotai 3.0.0's `dist` is `atomWithObservable`'s unsubscribe timer, and `useAtomValueRaw` seeds from `store.get(atom)` in the reducer initializer, subscribes in `useEffect(() => store.sub(atom, rerender))`, and never re-reads after subscribing. The helper's doc now says v2 always re-rendered once after mount, that v3 drops that render, and that the write therefore arrives as its own render. "Microtask" is gone from the description too.

## The regression

Confirmed and fixed. Your test fails 3/3 on the old order — and the failure output is unambiguous, it renders `No Object Properties Available`, the RDF wording of that empty state. Passes 3/3 with the reorder. Added as `SchemaDiscoveryBoundary.integration.test.tsx`; it had to be a separate file because the existing suite's module-level `vi.mock("@/core")` would defeat the point.

One deliberate addition. The reorder makes correctness depend on hook declaration order, which is invisible at the call site — the next person to tidy that component reintroduces the bug and the only signal is a stale empty state on a vertices-only graph. So it carries a comment:

```tsx
const config = useConfiguration();
// Must precede useSchemaSync(): its mount fetch can write the schema
// synchronously, and these readers only observe writes that land after they
// subscribe, which React does in hook declaration order.
const hasSchema = useHasActiveSchema();
const schema = useMaybeActiveSchema();
```

Happy to drop that if the hazard sweep you're filing is going to remove the ordering dependency structurally — a comment defending an invariant is worse than not needing the invariant.

## Smaller

Dead flush removed. The rebase turned out bigger than one commit: main had since landed pnpm 12.4.2 (#2178) and Vitest 5 (#2166). I took main's lockfile and re-resolved with `corepack pnpm install` rather than fixing up the old one, so the peer suffixes are gone and the lockfile diff is now +11/-74. `pnpm install --frozen-lockfile` is clean, and the v3 test changes needed nothing further under Vitest 5.

## Description

Both notes added. On the `TS7006` re-measurement I checked rather than assumed: the four files are absent from the diff and contain no `ts-expect-error`, `ts-ignore`, or `any`, so the six errors really were downstream of the store import. The `useIsSyncing` note explains that the flag only goes truthy after an effect flush now.

Full run: `pnpm checks` clean, `pnpm test` 2673/2673 across 221 files, no `act()` warnings.

Agreed on keeping the follow-ups out of this one.

kmcginnes (approved):
All four addressed. I verified the reorder rather than taking the test's word for it: reverting the hook order makes your `SchemaDiscoveryBoundary.integration.test.tsx` fail with `Unable to find an element with the text: Children`, and restoring it passes 8/8. That test is load-bearing.

Lockfile is clean, +11/-74 with no peer-suffix churn, `--frozen-lockfile` passes, and the base is level with main. `pnpm checks` clean and 2673/2673 with no act warnings across three full runs.

On the comment defending hook order: keep it. The sweep may not land soon and may not restructure that component at all, so until the invariant actually goes away the comment is the only signal at the call site. It also explains why rather than restating what the code does, which is the kind we want. Delete it when the ordering dependency is gone, not in anticipation. The test is the enforcement and the comment is the explanation, and I'd rather have both.

Two notes, neither blocking.

You removed three flush calls rather than the one dead one I flagged. Tests pass and there are no act warnings, so the removals look right, but it wasn't in your summary and I only noticed by counting.

My first full-suite run reported 1 failed of 2673 and I didn't capture which test. Three runs since are clean and the new integration test is 8/8, so I can't reproduce or attribute it. Probably a worker blip given vitest spawns one per file. Mentioning it in case you see something similar.

Approving. Worth letting the workflows finish before merging, since this hasn't been through CI yet.

Issue #2177, opened 2026-09-14
Run pnpm dedupe to settle lockfile drift


kmcginnes:
`pnpm dedupe --check` exits 1 with `ERR_PNPM_DEDUPE_CHECK_ISSUES` and 247 lines of proposed changes, so the committed `pnpm-lock.yaml` is valid but is not what pnpm would write today.

Surfaced while reviewing the pnpm 12 bump (#2155), but it predates it. pnpm 11.9.0 reports the same drift, and that bump left the lockfile's project-dependency document byte-identical.

## What it wants to change

Two kinds of drift, neither a declared version change:

- **Peer-suffix churn**, the bulk of it. Entries gain `supports-color@7.2.0` inside their peer-dependency suffix, e.g. `@babel/core 7.29.7 → 7.29.7(supports-color@7.2.0)`.
- **A transitive bump**, `@aws-sdk/types 3.973.13 → 3.974.5` under several `@aws-crypto/*` and `@aws-sdk/*` packages.

## Work

- Run `pnpm dedupe` and commit the result on its own, so a few hundred lines of lockfile do not ride along with an unrelated change.
- Confirm `pnpm install --frozen-lockfile`, `pnpm checks`, and `pnpm test` pass afterwards.
- Decide whether `pnpm dedupe --check` belongs in `.github/workflows/unit.yml`. Nothing runs it today, which is why this went unnoticed.

## Related Issues

- Part of #2149
- Related to #2155

<audience-footer>

> [!IMPORTANT]
> Internal only — this issue is maintained by the core team and is not accepting external contributions.

</audience-footer>

mjuarros:
Picked this up in #2205. Two things in the description above no longer match what `pnpm dedupe` actually proposes, noting them so the diff does not read as a surprise.

**The peer-suffix churn is gone.** No `supports-color@7.2.0` peer suffixes remain anywhere in the lockfile. That was described as the bulk of the 247 lines, and it cleared on its own — most likely with the pnpm 12.4.2 bump (#2178) or Vitest 5 (#2166), both of which merged after this was filed.

[...]

Issue #2158, opened 2026-09-10
Spike: how the container image should acquire Node and pnpm


kmcginnes:
## Goal

How should the container image acquire Node and pnpm? The current approach in the `Dockerfile` fetches both from the public internet at build time, outside every guardrail the repo applies to its other dependencies.

```dockerfile
curl -fsSL https://nodejs.org/dist/v${NODE_VERSION}/node-v${NODE_VERSION}-linux-${NODE_ARCH}.tar.xz | tar -xJ -C /usr/local --strip-components=1 && \
npm install --global corepack@latest && \
```

Two concrete gaps. The Node tarball is piped straight into `tar` with no checksum check against the published `SHASUMS256.txt`, so whatever the CDN returns is what gets extracted and run. And `corepack@latest` is an unpinned floating version resolved at build time, which sits outside the lockfile and therefore outside `minimumReleaseAge: 1440`, `trustPolicy: no-downgrade`, and `blockExoticSubdeps: true` — settings the repo applies to everything else it installs. A build today and a build next week can produce different images from the same commit.

There's a related maintenance cost: the Node version is pinned in five places (`Dockerfile` `NODE_VERSION`, `.nvmrc`, `engines.node` in two manifests, `devEngines.runtime.version`) that have to be moved together, plus the requirements in `docs/development.md`.

## Approaches to investigate

- **Node as a package dependency.** Declaring Node in `package.json` would put the runtime under the lockfile and the release-age gate, and would decouple the image from whatever Node the build machine happens to have. Worth checking what this actually looks like in practice — which package provides it, how it interacts with `devEngines`, whether it collapses several of the five pins into one, and what it does to image size.
- **Install pnpm without a separate Node bootstrap.** pnpm publishes standalone binaries that bundle their own Node, so the image may be able to stay entirely within the pnpm world rather than fetching Node and then layering corepack on top. Investigate whether this can replace both the tarball fetch and the corepack install, and how it squares with the `packageManager` field.
- **Keep the current shape and harden it.** Verify the tarball against `SHASUMS256.txt` before extracting, and pin corepack to an exact version. Lowest-risk option, and the baseline the others should be compared against.
- **A prebuilt Node base image.** Using an official Node image (or a distroless variant) instead of Amazon Linux plus a manual Node install. Note the current base image choice is deliberate, so this needs to weigh what would be lost.

[...]

Issue #2153, opened 2026-09-10
Move to Node 26 when it reaches Active LTS


kmcginnes:
We track the latest Active LTS Node release. Node 24 is Active LTS today, so the pins are correct as
of now — this issue is the scheduled move to Node 26, which enters Active LTS on **2026-10-28**.

`@types/node` 26 is already published and typechecks cleanly against the current source, so the code
isn't the obstacle. We're holding it at ^24 only because shipping types ahead of the runtime would let
code compile against APIs the runtime doesn't have.

## Work

When Node 26 reaches Active LTS, change all of these together to the then-current 26.x LTS patch:

- `NODE_VERSION` in the `Dockerfile`
- `.nvmrc`
- `engines.node` and `devEngines.runtime.version` in the root `package.json`
- `engines.node` in `packages/graph-explorer/package.json`
- `@types/node` to `^26.x` in the root, `packages/graph-explorer`, and
  `packages/graph-explorer-proxy-server` manifests
- The Node version references in `docs/development.md`

`.github/workflows/unit.yml` now reads `node-version-file: ".nvmrc"`, so CI follows the `.nvmrc` pin
directly — update that file and CI moves with it.

**How to check the trigger:** DefinitelyTyped pins the `@types/node` `latest` dist-tag to a much older
major, so `pnpm outdated` reports the highest in-range 24.x and will never surface 26.x. Use
`npm view @types/node versions` instead. For the runtime itself, the LTS date is in
[`nodejs/Release`](https://github.com/nodejs/Release)'s `schedule.json`.

Verify with `pnpm checks`, `pnpm test`, a Docker image build, and the container smoke checks in
`.github/workflows/test_build_docker.yml`.

## Acceptance

All Node version pins and `@types/node` sit on the current Node 26 LTS release, and CI is green.

<audience-footer>

> [!IMPORTANT]
> Internal only — this issue is maintained by the core team and is not accepting external contributions.

</audience-footer>

kmcginnes:
One blocker to add to the work list: **Node 26 does not ship Corepack.**

Confirmed against local installs. Node 24.13.0, 24.16.0, and 24.21.0 each contain a `corepack` binary; Node 26.7.0 does not. The `node-v24.21.0-linux-*` tarball the `Dockerfile` fetches ships `corepack` 0.36.0.

Two things break on the move:

- `Dockerfile` runs `corepack enable` to acquire pnpm. On Node 26 that command does not exist, so the image build fails.
[...]

PR #2178, opened 2026-09-14
Update pnpm to 12.4.2

### LINKED ISSUE #2155: Update pnpm to the latest 12.x release
`pnpm` was pinned at 11.9.0 via `packageManager` in the root `package.json`. It was left alone in the dependency sweep because it does not behave like an ordinary manifest bump. Target is the latest 12.x, currently **12.4.2**.

## Why it was held back

pnpm records its own version inside the lockfile, in a `packageManagerDependencies` block under the root importer:

```yaml
importers:
  .:
    configDependencies: {}
    packageManagerDependencies:
      '@pnpm/exe':
        specifier: 11.9.0
        version: 11.9.0
      pnpm:
        specifier: 11.9.0
        version: 11.9.0
```

Editing `packageManager` by hand does not update that block. `pnpm install`, `pnpm install --lockfile-only`, and `pnpm install --force` all report "Already up to date" and leave the old specifiers in place, along with the `@pnpm/exe` and per-platform `@pnpm/*` entries in `packages:`. Deleting the block by hand does not help either, because nothing regenerates it and the orphaned entries are not pruned.

**Resolved:** `corepack use pnpm@<version>` is the procedure. It rewrites `packageManager` with a correct `+sha512` hash and then runs a real install, which regenerates the block and prunes the orphans in one step.

## Work

- Bump `packageManager` to `pnpm@12.4.2` with a matching `+sha512.<hex>` integrity hash. Generate it with `corepack use`; do not hand-write it. The hash is the hex encoding of npm's base64 `dist.integrity`, not the base64 itself.
- Update the pnpm requirement in `docs/development.md`.
- Get the lockfile's `packageManagerDependencies` block and the `@pnpm/exe.*` entries onto 12.4.2 consistently, with the pnpm 11 entries removed.
- Move `.github/workflows/unit.yml` to a `pnpm/action-setup` release that supports pnpm 12. v5 does not; v6.1.0 added it.
- Pin the `Dockerfile`'s `corepack@latest` to an exact version. The container acquires pnpm only through `corepack enable`, and pnpm 12 ships as a native executable, so the Corepack version is coupled to the pnpm major.
- Remove `engines.pnpm`. pnpm 12 does not enforce a root project's `engines.pnpm` at all, so it is a declarative range duplicating the exact version `packageManager` already pins. npm and Yarn are already blocked without it, npm by `EUNSUPPORTEDPROTOCOL` on the `workspace:*` dependencies and Yarn by `packageManager` naming pnpm.
- Do not add a check comparing the `+sha512` hash against the lockfile. This was originally required here on the premise that nothing verifies the hash. That premise is wrong; see "Verification, corrected" below.

## Verification, corrected

A check was built for this and then removed, because testing it disproved its premise.

pnpm never reads the `+sha512` hash. With `packageManager` set to `pnpm@12.3.4+sha512.` followed by 128 zeros and a correct lockfile, a cold-cache install still downloads and runs the pinned version. What verifies the bytes is the install path: pnpm resolves through `pnpm-lock.yaml` and checks each download against npm's registry signature, so corrupting the lockfile's `integrity` while leaving the manifest hash correct stops the install with "Refusing to run pnpm@12.3.4: its npm registry signature could not be verified". Corepack is the tool that does read the hash, and it fails closed on a mismatch. So both install paths verify what they fetch, and the hash stays in `packageManager` because Corepack depends on it.

The check also could not catch the case that matters. `corepack use` and `pnpm install` write the manifest hash and the lockfile integrity from separate downloads, which was the argument for comparing them. But if both ran against the same compromised registry, both values come from the same bad bytes and the comparison passes.

One related result, which narrows the "Why it was held back" section above rather than contradicting it. That section records `pnpm install` leaving `packageManagerDependencies` stale, observed with pnpm 11.9.0 pinned. Under pnpm 12 it does not: with the pin at 12.4.2 and the lockfile recording 12.3.4, `pnpm install` rewrote the block and every `@pnpm/exe.*` entry to 12.4.2. It did so under `--frozen-lockfile` too, and still exited 0, which is worth knowing separately. pnpm 11's behaviour was not retested, so the original observation stands as recorded. Either way `corepack use` remains the procedure, because it is the only thing that writes a correct hash.

Full reproduction output is in [this comment on #2178](https://github.com/aws/graph-explorer/pull/2178#issuecomment-5702288006).

## Not in scope

Whether the repo should rely on pnpm's own version management or on Corepack was originally listed here. Local development is settled: `docs/development.md` leads with letting pnpm switch itself, because that route resolves through the lockfile-pinned `@pnpm/exe.*` integrity, and presents Corepack as the fallback. How the **container** acquires Node and pnpm is a larger question and belongs to #2158; pinning the existing `corepack@latest` to an exact version is the conservative baseline that issue names, and is in scope here.

## Acceptance

`packageManager`, `pnpm-lock.yaml`, and `docs/development.md` all agree on 12.4.2; `engines.pnpm` is gone; the `Dockerfile` pins an exact Corepack version; `pnpm install --frozen-lockfile` succeeds from a clean checkout; `pnpm checks` and `pnpm test` pass; CI is green.

## Related Issues

- Part of #2149
- Related to #2158
- Related to #2177
- Related to #2179

kmcginnes:
## Description

Moves the pinned toolchain from pnpm 11.9.0 to 12.4.2, the latest 12.x. #2155 held this back on the belief that editing `packageManager` by hand leaves the lockfile's `packageManagerDependencies` block stale. It does not, and I checked: pnpm rewrites that block, and every `@pnpm/exe.*` entry, on the next install. The reason to run `corepack use pnpm@12.4.2` is the `+sha512` hash, which you cannot compute by hand, and the orphaned `@pnpm/exe` and `@pnpm/*` platform entries it prunes in the same step. That procedure is now written down, so the next bump is one command plus a docs line.

Three things beyond the version number:

**`engines.pnpm` is gone.** pnpm 12 does not enforce a root project's `engines.pnpm`; an install succeeds with `">=99.0.0"` set. It was a declarative range duplicating the exact version `packageManager` already pins with an integrity hash. Nothing regresses by removing it: npm is already refused by `EUNSUPPORTEDPROTOCOL` on the `workspace:*` dependencies, and Yarn by `packageManager` naming pnpm. `devEngines.packageManager` would give npm a clearer message, but pnpm 12 refuses to honour both fields at once and ignores `packageManager` when they coexist, which is why it was removed in 2b1a4acb and why it stays out.

**Nothing here checks the `+sha512` hash, and nothing needs to.** Earlier revisions of this PR added a check for it. I removed it, because its premise was wrong. pnpm never reads that hash: with `packageManager` set to 128 zeros, a cold-cache install still downloads and runs the pinned version in 100ms. What verifies the bytes is the install path itself. pnpm resolves the version through `pnpm-lock.yaml` and checks each download against npm's registry signature, so corrupting the lockfile's `integrity` while leaving the manifest hash correct stops the install outright:

```
[ERROR] Refusing to run pnpm@12.3.4: its npm registry signature could not be verified
(pnpm@12.3.4 has an invalid registry signature with keyid
SHA256:DhQ8wR5APBvFHLF/+Tc+AYvPOdTpcIDqOhxsBHRwC7U). The bytes selected by this
project's lockfile/registry do not match a published, signed pnpm release.
```

Corepack is the tool that does read the hash, and it fails closed on a mismatch with `Error: Mismatch hashes.` So both install paths verify what they fetch without help. The check also could not catch the case that actually matters: if `corepack use` and `pnpm install` both ran against the same compromised registry, the two recorded values agree and the comparison passes. The hash stays in `packageManager` because Corepack users depend on it.

**The docs lead with letting pnpm switch itself.** pnpm 12 is a native executable, and Corepack installs no dependencies, so under Corepack the binary arrives in a second download verified against npm's registry signature rather than against the pinned hash. Letting any pnpm 11+ self-manage instead resolves through the lockfile-pinned `@pnpm/exe.*` integrity, so that is now the documented default and Corepack is the fallback for someone with Node but no pnpm.

The `pnpm/action-setup` bump to v6.1.0 is required, not incidental: v5 does not support pnpm 12. It is pinned to the commit rather than the annotated tag object, matching the other eight action pins in the repo, since the commit stays reachable from the action's default branch while a tag object does not.

## How to read

1. [package.json](https://github.com/aws/graph-explorer/pull/2178/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519) — the pin and the `engines.pnpm` removal
2. [docs/development.md](https://github.com/aws/graph-explorer/pull/2178/files#diff-97db29a7915320e63d41d38a0440360a87055ee8ed03757aa263116dbbb4aabe) — the contributor path and the upgrade runbook; the largest prose change
3. [.github/workflows/unit.yml](https://github.com/aws/graph-explorer/pull/2178/files#diff-e17ac7ab478a18e9dac875e191cc3e0d754fab62eddddfb05d4619495be927d9) — the action repin, one line
4. [Dockerfile](https://github.com/aws/graph-explorer/pull/2178/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557) — corepack pinned instead of tracking `latest`
5. [AGENTS.md](https://github.com/aws/graph-explorer/pull/2178/files#diff-a54ff182c7e8acf56acfd6e4b9c3ff41e2c41a31c9b211b2deb9df75d9a478f9) — points agents at the upgrade procedure
6. [pnpm-lock.yaml](https://github.com/aws/graph-explorer/pull/2178/files#diff-32824c984905bb02bc7ffcef96a77addd1f1602cff71a11fbbfdd7f53ee026bb) — mechanical; read last

The lockfile diff is 269 of the ~341 changed lines and touches nothing but pnpm's own entries. The project-dependency document below the `---` separator is byte-identical to `main`. `@reflink/reflink` and `detect-libc` disappear because they were pnpm 11's JavaScript dependencies and pnpm 12 is a Rust binary.

The second commit is unrelated to pnpm 12 and can be read on its own: `pnpm test -- -t "pattern"` never filtered anything, because pnpm forwards `--` to the script literally, so it silently ran the whole suite and reported it green.

The four middle commits add and refine the pin check; the last one removes it. They are kept because the reasoning is worth reading, and this branch squash merges.

## Validation

- `pnpm install --frozen-lockfile` from a clean checkout, with no lockfile movement
- `pnpm checks` clean, `pnpm test` 220 files / 2672 tests passing
- The `+sha512` hash independently confirmed three ways: generated by `corepack use`, matched against the registry's published `dist.integrity` converted base64 to hex, and matched against a sha512 of a separately downloaded tarball
- Corepack confirmed to enforce that hash and fail closed, tested against a wrong hash from a cold cache
- pnpm confirmed to ignore the hash and to refuse a tampered lockfile, each tested from a genuinely cold store isolated by `PNPM_HOME` and `--store-dir`
- `docker build` succeeds and the container assertions from `test_build_docker.yml` still pass, including that pnpm, npm, and corepack are absent from the finished image, with no cached native binary left behind
- All four `pnpm-workspace.yaml` hardening settings confirmed still recognized, which matters because pnpm 12 hard-fails on unrecognized keys instead of ignoring them
- Bootstrapping verified on pnpm 11.9.0, 11.22.0, and 11.26.0, and through Corepack on Node 24

One surprise worth recording, unrelated to anything this PR changes: when `packageManager` names a version the lockfile does not record, `pnpm install --frozen-lockfile` rewrites `packageManagerDependencies` and exits 0 rather than failing. So CI will not catch a hand-edited pin the way you might assume.

## Related Issues

- Fixes #2155
- Part of #2149
- Related to #2158
- Related to #2153
- Related to #2177
- Related to #2179

### Check List


scripts/verify-pnpm-pin.test.ts
```diff
@@ -0,0 +1,68 @@
+import { describe, expect, test } from "vitest";
```
mjuarros: [Important] Only the pure helpers `parsePin`/`integrityToHex` are tested here; `main()`'s actual behavior (registry-hash mismatch, running-version mismatch, and the success log) is never exercised. Per `docs/agents/testing.md` ("mock only external systems"), mock `execFileSync`/`node:child_process` so these branches — including the exact error text built in each — are covered too.

scripts/verify-pnpm-pin.ts
```diff
@@ -0,0 +1,85 @@
+// Asserts the `packageManager` pin is internally honest and is what is running.
+//
+// `pnpm/action-setup` parses the version out of `packageManager` and discards the
+// integrity hash, so without this check the hash is never verified in CI and a
+// wrong or hand-written one would go unnoticed.
+import { execFileSync } from "node:child_process";
+import { readFileSync } from "node:fs";
+
+const PIN_PATTERN = /^pnpm@(\d+\.\d+\.\d+)\+sha512\.([0-9a-f]{128})$/;
```
mjuarros, line 9: [Nit] The error message's expected-shape string (`"pnpm@<x.y.z>+sha512.<128 hex chars>"`, line 24) is a separate hand-written description of this pattern rather than derived from it, so the two can silently drift if the pattern is ever adjusted.

scripts/verify-pnpm-pin.ts
```diff
@@ -0,0 +1,85 @@
...
+  const base64 = integrity.replace(/["\s]/g, "").replace(/^sha512-/, "");
+  if (base64.length === 0) {
+    throw new Error("the registry returned an empty dist.integrity");
+  }
+  return Buffer.from(base64, "base64").toString("hex");
+}
+
+function main() {
+  const { packageManager } = JSON.parse(
+    readFileSync("package.json", "utf8"),
```
mjuarros: [Nit] `JSON.parse(...) as { packageManager?: unknown }` is a bare cast at a file-boundary read; AGENTS.md prefers Zod at boundaries over casts. Low risk since `parsePin` validates the value immediately after, but it still adds an unchecked assumption about `package.json`'s shape.

scripts/verify-pnpm-pin.ts
```diff
@@ -0,0 +1,85 @@
...
+        `Regenerate the pin with "corepack use pnpm@${pin.version}" rather than editing it by hand.`,
+    );
+  }
+
+  const running = execFileSync("pnpm", ["--version"], {
+    encoding: "utf8",
+  }).trim();
+  if (running !== pin.version) {
+    throw new Error(
+      `packageManager pins pnpm@${pin.version} but pnpm ${running} is running`,
```
mjuarros: [Important] This version-mismatch error gives no remedy, unlike the hash-mismatch branch just above it which tells the reader to run `corepack use pnpm@<version>`. Since self-switching to a v12 pin needs pnpm >=11.10.0, add an actionable hint here (e.g. "upgrade your global pnpm — self-switching to v12 needs pnpm >=11.10.0").

docs/development.md
```diff
@@ -4,7 +4,7 @@ Build instructions and development setup for contributing to Graph Explorer. For
 
 ## Requirements
 
-- pnpm >=11.9.0
+- pnpm 12.4.1, pinned by `packageManager`; any pnpm 11 or newer switches to it automatically
```
mjuarros, line 7: [Important] "any pnpm 11 or newer switches to it automatically" overstates pnpm's capability: `packageManager` self-switching to a v12 pin only works from pnpm >=11.10.0, not any 11.x. This repo's own previous minimum was 11.9.0, so a contributor still on exactly that version would silently fail to auto-switch with no explanation from this doc. Same issue repeated at line 22.

docs/development.md
```diff
@@ -17,15 +17,45 @@ nvm use
 
 Otherwise, use whatever method you use to install [Node v24.21.0](https://nodejs.org/en/download).
 
-### Node Corepack
+### pnpm version
 
-[Corepack](https://nodejs.org/api/corepack.html) is used to ensure the package manager used for the project is consistent.
+`packageManager` in the root `package.json` pins the exact pnpm version, and you do not have to match it yourself. Any pnpm 11 or newer reads that field and hands the command to the pinned version, so install pnpm however you like and let it switch:
```
mjuarros, line 22: [Important] Same overstatement as line 7 — "Any pnpm 11 or newer" should read "pnpm >=11.10.0", with a note on what to do if a contributor is on an older 11.x (upgrade global pnpm first).

AGENTS.md
```diff
@@ -72,8 +73,9 @@ Run from project root with `pnpm`. Use only these scripts — never invoke `tsc`
 - `pnpm check:format` / `pnpm format` — check / fix formatting
 - `pnpm test` — run all tests
 - `pnpm test <path>` — test files matching a path substring (file, dir, or partial)
-- `pnpm test -- -t "pattern"` — test by name
+- `pnpm test -t "pattern"` — test by name
```
mjuarros: [Nit] Switching `pnpm test -- -t "pattern"` to `pnpm test -t "pattern"` is correct pnpm behavior (test/start/stop forward args verbatim), but nothing here calls out that this differs from npm's `--` requirement, which could trip up contributors relying on npm muscle memory. Consider a short parenthetical.

scripts/verify-pnpm-pin.test.ts
kmcginnes: Done in 60e64bcb. Exported `main` and mocked `node:child_process` and `node:fs`, which `docs/agents/testing.md:9` sanctions since `pnpm view` is a network call. Five new tests cover the registry-hash mismatch and the running-version mismatch with their exact message text, the success log, that a malformed pin short-circuits before the registry is consulted, and a manifest that is not a JSON object.

I mutation-checked the two important branches rather than trusting green: inverting `registryHash !== pin.hash` fails one test, and deleting the remedy clause from the version-mismatch message fails another.

scripts/verify-pnpm-pin.ts
kmcginnes: Fixed in 60e64bcb, though not with the suggested wording, because the premise is wrong (see the reply on `docs/development.md:7`). pnpm 11.9.0 does self-switch to a v12 pin. The message now reads:

```
packageManager pins pnpm@12.4.1 but pnpm 11.9.0 is running.
pnpm reads packageManager and switches to the pinned version from pnpm 11 onwards, so upgrade your pnpm (or run "corepack enable") and try again.
```

scripts/verify-pnpm-pin.ts
kmcginnes: Fixed in 60e64bcb, without adding Zod. `zod` is not a root dependency — it is declared only in `packages/graph-explorer` and `packages/graph-explorer-proxy-server` — and pulling it into the root manifest for one 80-line script is more than this boundary needs.

The cast is gone instead. A `readPinField` helper narrows the parsed JSON to an object and reads the one field without asserting anything about the rest of the manifest, and `parsePin` still takes `unknown`. A new test covers the non-object case.

AGENTS.md
kmcginnes: Added in 60e64bcb: "no `--` separator; pnpm forwards args verbatim, unlike npm". Worth it given the whole reason the line was wrong is npm muscle memory.

docs/development.md
kmcginnes: Not changing this one, because the 11.10.0 threshold does not hold up. I tested it directly, since it would have been a real problem if true.

pnpm 11.9.0 self-switches to this v12 pin. Against a fixture pinning `pnpm@12.4.1`, with `~/Library/pnpm/.tools/pnpm/12.4.1*` deleted first so nothing was warm:

```
pnpm 11.9.0    -> + dompurify 3.4.15  Done in 198ms using pnpm v12.4.1
pnpm 11.10.0   -> + dompurify 3.4.15  Done in 206ms using pnpm v12.4.1
pnpm 11.26.0   -> + dompurify 3.4.15  Done in 195ms using pnpm v12.4.1
```

I did not trust that output on its own, because "using pnpm v12.4.1" could just be echoing the pin. Two further checks:

1. Re-ran 11.9.0 with `PATH` stripped to `/usr/bin:/bin` plus the Node bin dir, so no pnpm 12 on `PATH` could be picked up. Same result.
2. Pointed `packageManager` at a version that does not exist. Both 11.9.0 and 11.26.0 fail with `[ERROR] No matching version found for pnpm@99.99.99 while fetching it from https://registry.npmjs.org/`, which proves the switch is a real fetch rather than a printed string.

The 197ms is fast because the tarball is still in the content-addressable store under `~/Library/pnpm/store`; deleting `.tools` does not evict it.

Where there *is* a real floor is the pnpm 10 line, and it is not a clean cutoff: 10.28.1 fails with `Failed to switch pnpm to v12.4.1 ... spawnSync .../bin/pnpm ENOEXEC` because pnpm 12 moved the executable into `@pnpm/exe.<platform>` optional dependencies, while 10.34.5 succeeds. That is exactly why the doc says 11 rather than 10 — an earlier draft said "pnpm 10 or newer" and this branch corrected it.

Happy to be shown wrong if you have a case where a specific 11.x fails; I would want the exact version and whether the tools cache was cold.

docs/development.md
kmcginnes: Same as the reply on line 7: 11.9.0 self-switches fine, verified from a cold cache, with a stripped `PATH`, and with a nonexistent-version control proving the fetch is real. Leaving both lines at "pnpm 11 or newer".

scripts/verify-pnpm-pin.ts
kmcginnes: Leaving this one. Deriving the human-readable shape from `PIN_PATTERN` would mean either printing `PIN_PATTERN.source` (`^pnpm@(\d+\.\d+\.\d+)\+sha512\.([0-9a-f]{128})$`), which is worse for the person reading a CI failure, or building the pattern from parts and reassembling prose from them, which is more machinery than an 80-line script warrants.

The drift risk is real but bounded and now caught by tests: `verify-pnpm-pin.test.ts` asserts the message for ten distinct malformed pins, including base64-instead-of-hex, sha1-instead-of-sha512, a version range, and a partial version. Loosening the pattern without updating the message breaks those.

docs/development.md
mjuarros: Appreciate the empirical check, but this contradicts pnpm's own documented cutoff, not just my read of it. From the pnpm 11.10 release notes and the `pnpm self-update` docs, verbatim: "Since v11.10.0, `pnpm self-update` (and `packageManager` version-switching) can install and link pnpm v12, the Rust port." (https://pnpm.io/blog/releases/11.10, https://pnpm.io/cli/self-update). That's the vendor stating switching *to v12 specifically* wasn't possible before 11.10.0 — a dated feature addition, not a vague compatibility note.

That doesn't mean your test result was fake — the likely explanation is Corepack. If Corepack was enabled on the machine you tested on, the `pnpm` resolved on `PATH` may have been Corepack's shim rather than the raw 11.9.0 binary, and Corepack does its own independent `packageManager`-based switching regardless of the underlying pnpm version. Stripping `PATH` to `/usr/bin:/bin` plus the Node bin dir wouldn't rule that out, since Corepack's shims typically live in the Node bin dir.

Can you confirm whether Corepack was active during that test (e.g. check if the resolved `pnpm` binary is a Corepack shim, or retest with Corepack fully disabled/uninstalled)? If it switches without Corepack in the picture at all, I'll retract this — but as it stands it contradicts pnpm's own docs, so I don't think we should keep "any pnpm 11 or newer" without ruling that out.

mjuarros: Retracting my pushback — I reproduced it and you're right.

Method: installed a real, non-Corepack pnpm 11.9.0 via `npm install --prefix . pnpm@11.9.0` (a plain npm package, nothing to do with Corepack), then ran its `bin/pnpm.mjs` directly against a fixture pinning `packageManager: pnpm@12.4.1+sha512...`.

First attempt was contaminated: I put node's bin dir on `PATH` so `node` would resolve, but on this (nvm) machine that's the exact same directory the Corepack `pnpm` shim lives in, so it wasn't a clean test. Fixed by symlinking only `node` into an empty directory and using that as the sole `PATH` entry, confirming `which pnpm` and `which corepack` both fail (exit 1) in that environment. Ran the raw 11.9.0 binary again:

```
env -i PATH="/tmp/clean-path-bin:/usr/bin:/bin" HOME="$HOME" /path/to/pnpm@11.9.0/bin/pnpm.mjs --version
# -> 12.4.1
```

Confirmed it's a real switch, not a fluke: it wrote a `pnpm-lock.yaml` in the fixture recording `packageManagerDependencies: pnpm: 12.4.1` with the platform `@pnpm/exe.*` optional deps and the pin's own registry-verified integrity hash (`sha512-LoHjmdc/6Dk...`, matching this repo's own pin exactly). That's genuine dependency resolution against the registry, not a printed string or a corepack handoff — as a control, the same binary prints `11.9.0` outside any pinned project.

So the pnpm blog's "since v11.10.0" note must be about something narrower than I read it as (likely the pre-GA `next-12` dist-tag handling while v12 was still prerelease), not the general packageManager-switch mechanism, which clearly already existed in 11.9.0 and works fine against today's GA v12.4.1. "Any pnpm 11 or newer" stands — no doc change needed here or on line 22.

kmcginnes: Thanks for reproducing it, and your contamination point was fair about my method. `$NODE_BIN` on this machine is an fnm bin dir that does contain a `pnpm`, so my `PATH="/usr/bin:/bin:$NODE_BIN"` run was open to exactly the objection you raised. Your symlink-only-`node` setup is the sounder test. My nonexistent-version control still argues against a Corepack handoff, since the failure came back in pnpm's own wording rather than Corepack's, but that is weaker evidence than what you ran.

I can confirm your closing hypothesis from the release note itself rather than leaving it as a guess. The 11.10 post is real and says what you quoted, but that section is scoped to the prerelease: it describes v12 as "published under both the `pnpm` and `@pnpm/exe` names on the **`next-12` dist-tag**" and gives `pnpm self-update next-12` as the example. So 11.10 added handling for the pre-GA v12 layout, not the `packageManager` switch mechanism, which 11.9.0 already had and which works against GA 12.4.1.

I considered adding a line to `docs/development.md` warning that pnpm's own 11.10 note reads stricter than reality, and decided against it. Release-note archaeology is noise for a contributor who just wants to install and go, and this thread is the durable record if anyone hits the same doubt.

kmcginnes:
I removed the pin check in 9cc36c8e. The `packageManager` pin and its `+sha512` hash stay, as does the `engines.pnpm` removal, the `pnpm/action-setup` v6 bump, and the `corepack use` runbook. That makes the four threads on `scripts/verify-pnpm-pin.ts` and its tests moot, so rather than just closing them, here is why, with the output.

The check's premise was that nothing verifies the hash. That is wrong for the path almost everyone uses, and I only found out by testing it.

**pnpm never reads the hash.** With `packageManager` set to `pnpm@12.3.4+sha512.` followed by 128 zeros, then again by 128 `a` characters, and a correct lockfile. My first run finished in 5ms and proved nothing, because the version was already in the local store, so I re-ran with the store, `PNPM_HOME`, and the cache all redirected to an empty directory to force a real fetch:

```
$ PNPM_HOME=/tmp/coldpnpm XDG_CACHE_HOME=/tmp/coldpnpm/cache \
    pnpm --store-dir /tmp/coldpnpm/store install
Done in 100ms using pnpm v12.3.4
$ du -sh /tmp/coldpnpm
 74M	/tmp/coldpnpm
```

74MB downloaded, pinned version running, zeroed hash ignored. Same for the 128 `a` variant.

**What does verify the bytes is the install path.** Corrupting the `integrity` that `pnpm-lock.yaml` records for `pnpm`, while leaving the manifest hash correct, from a cold store:

```
[ERROR] Refusing to run pnpm@12.3.4: its npm registry signature could not be verified
(pnpm@12.3.4: pnpm@12.3.4 has an invalid registry signature with keyid
SHA256:DhQ8wR5APBvFHLF/+Tc+AYvPOdTpcIDqOhxsBHRwC7U). The bytes selected by this
project's lockfile/registry do not match a published, signed pnpm release.
This can indicate a tampered lockfile or a malicious/unreachable registry.
```

That is a stronger check than the one I wrote. Mine compared two hex strings; this checks a signature over the actual bytes.

**Corepack does enforce the hash, and fails closed.** This is the case I had never tested and should have, since it is the reason to keep the hash in the manifest:

```
$ corepack pnpm --version     # packageManager hash set to 128 'a' characters
Error: Mismatch hashes. Expected aaaa…aaaa, got 961aa41fb077da3a04a441d9f8e15ebc0c96da8ef7
10b2eb67bf9ee7cb0610eabd48f1fd85f51cffe73846785fa0f87c56a3a872a1d893f8446741b5cce45457
    at installVersion (corepack.cjs:13497)

$ corepack pnpm --version     # correct hash
Downloading the pnpm 12.3.4 binary for darwin-arm64...
12.3.4
```

So both install paths verify what they fetch, one against a signature and one against the pinned hash, and the hash stays in `packageManager` because Corepack depends on it.

**The check could not catch the case that matters.** `corepack use` and `pnpm install` write the manifest hash and the lockfile integrity from separate downloads, which is what I argued made their agreement meaningful. But if both ran against the same compromised registry, both values come from the same bad bytes, they agree, and the comparison passes. It only caught drift between two values that a single command writes together.

One more result, which cuts against a claim I had put in the docs and have now corrected. I expected that hand-editing `packageManager` to a version the lockfile does not record would leave `packageManagerDependencies` stale. It does not:

```
$ pnpm install --frozen-lockfile     # pin at 12.4.2, lockfile recording 12.3.4
Already up to date
Done in 76ms using pnpm v12.4.2
exit: 0

$ git diff pnpm-lock.yaml | head -4
-      specifier: 12.3.4
-      version: 12.3.4
+      specifier: 12.4.2
+      version: 12.4.2
```

pnpm rewrote the block and every `@pnpm/exe.*` entry. So the one drift case the check uniquely caught repairs itself on the next install. Worth noting separately that `--frozen-lockfile` modified the lockfile and still exited 0, which is not what I would expect from that flag, and means CI will not catch a hand-edited pin the way you might assume. Nothing in this PR causes that.

Net effect: 3 files and 32 tests deleted, `checks` and the CI workflow back to one step each, and `scripts/` and its vitest project gone entirely, which leaves #2179 free to decide where root scripts and their tests should live.

scripts/verify-pnpm-pin.test.ts
kmcginnes: Moot now: the check and its 32 tests are deleted in 9cc36c8e. Your original point stood and the tests I added off it caught a real hash-conversion bug, so it was not wasted. The reason the whole thing goes is that its premise was wrong, with the reproduction output here: https://github.com/aws/graph-explorer/pull/2178#issuecomment-5702288006

scripts/verify-pnpm-pin.ts
kmcginnes: Moot now: `scripts/verify-pnpm-pin.ts` is deleted in 9cc36c8e, so there is no hand-written shape string left to drift. Reasoning and output: https://github.com/aws/graph-explorer/pull/2178#issuecomment-5702288006

scripts/verify-pnpm-pin.ts
kmcginnes: Moot now: the file is deleted in 9cc36c8e, so the boundary read is gone rather than fixed. Reasoning and output: https://github.com/aws/graph-explorer/pull/2178#issuecomment-5702288006

scripts/verify-pnpm-pin.ts
kmcginnes: Moot now: the check is deleted in 9cc36c8e, so there is no version-mismatch message left. Worth recording that the remedy you asked for landed on the right idea even though the specific threshold did not, and pnpm's own error for the case is better than mine was: it refuses to run a pnpm whose registry signature does not verify. Output: https://github.com/aws/graph-explorer/pull/2178#issuecomment-5702288006

mjuarros (approved):
Approved. The final changes address the review feedback, and the removal of the pin check is technically justified by the corrected verification model.

One non-blocking follow-up: please update issue #2155 so its acceptance criteria no longer require the removed pin/hash/running-version check and tests. That will keep the originating requirement aligned with the implementation we are approving.

docs/development.md
kmcginnes: Resolving: settled by your own reproduction above, which confirmed pnpm 11.9.0 self-switches to a v12 pin without Corepack in the picture. Docs stay at "pnpm 11 or newer".

docs/development.md
kmcginnes: Resolving for the same reason as the line 7 thread, which you retracted after reproducing it. Line 22 keeps "any pnpm 11 or newer".

AGENTS.md
kmcginnes: Resolving: the parenthetical landed and survives in AGENTS.md at the `pnpm test -t` bullet, "no \`--\` separator; pnpm forwards args verbatim, unlike npm".

PR #2189, opened 2026-09-17
Disable Monaco link detection in code viewers

### LINKED ISSUE #2176: Monaco turns URLs in error and raw response viewers into clickable links
The read-only Monaco viewers turn URLs found in database responses into clickable links, which was never a deliberate choice. Monaco's `links` option defaults to `true` and `CodeEditor` does not set it, so the default applies everywhere.

## Where it shows up

Two of the three `CodeEditor` call sites render content that comes straight from the database or the proxy, both marked `readOnly: true`:

- `src/components/PanelError.tsx:100` renders `errorData`, which is built from `NetworkError.data` (a raw response body), `ZodError.issues` (the values that failed validation, so database content), and `ServerConnectionError.url`.
- `src/modules/SearchSidebar/ShowRawResponseDialog.tsx:42` renders `rawResponseText`, the unmodified query response.

An `http://` or `https://` URL anywhere in that text becomes ctrl-clickable and opens in a new tab. Monaco's link detector accepts only `http:`, `https:`, and `file:` schemes and the standalone opener routes through `windowOpenNoOpener`, so the behavior is limited to navigation. Even so, these two dialogs exist to show the operator exactly what came back, and a live link in that text is an affordance we did not ask for and would not miss.

The third call site, `PersistenceStatusIndicator`, renders local state rather than response data, so it matters less there.

## Work

Add `links: false` to the shared options block in `src/components/CodeEditor.tsx`, next to the other Monaco features already turned off there (`minimap`, `guides`, `stickyScroll`). Setting it once covers every viewer.

The decision worth making first: if any current or future viewer is meant to surface a clickable endpoint URL, it belongs on that call site's `options` instead of the shared default. Nothing does today.

## Acceptance

URLs inside error details and raw query responses render as plain text.

<audience-footer>

> [!IMPORTANT]

</audience-footer>

mjuarros:
## Description

- Set `links: false` in the shared `CodeEditor` options block. Monaco defaults this option to `true` and `CodeEditor` never set it, so `http://`, `https://`, and `file://` strings appearing in database and proxy responses became ctrl-clickable in the read-only viewers. Setting it once covers all three call sites (`PanelError`, `ShowRawResponseDialog`, `PersistenceStatusIndicator`) rather than repeating it per site.
- Placed it above the `...options` spread so a future viewer that genuinely wants a clickable endpoint URL can still opt back in from its own `options`. Nothing does today.
- Added two tests: one for the default, one for the per-call-site override. The override test is what pins the placement — moving `links: false` below the spread fails it, which is the failure mode that would otherwise go unnoticed.

`readOnly: true` does not suppress link detection, which is why this went unnoticed.

## Validation

`pnpm checks` and `pnpm test` (220 files, 2674 tests) both pass.

To check by hand, in either viewer a URL should now render as plain text — no hover underline, no "Follow link (cmd + click)" tooltip, and no ctrl-click navigation — while staying selectable and copyable:

- **Raw query response**: run `g.inject("https://example.com")` (Gremlin) or `RETURN "https://example.com" AS url` (openCypher) in the query tab, then click **Raw Response**. Any SPARQL query works without crafting, since every URI in the response is already `http://`.
- **Error details**: add a direct (non-proxied) connection pointing at a closed port such as `http://localhost:9999/`, let schema discovery fail, then open **Error Details**. The error data shows `{ "url": "http://localhost:9999/" }`.

codeEditorLinkManagement

## Related Issues

- Fixes #2176

### Check List


packages/graph-explorer/src/components/CodeEditor.tsx
```diff
@@ -37,6 +37,9 @@ export function CodeEditor({
         stickyScroll: {
           enabled: false,
         },
+        // These viewers show database and proxy responses verbatim; a URL in
+        // that text is content, not an endpoint we want to offer to open.
+        links: false,
```
kmcginnes: The comment says these are all response viewers, but `PersistenceStatusIndicator` renders `JSON.stringify(failures)`, which is local persistence state. #2176 drew that distinction. `CodeEditor` also can't know its callers, so "These viewers" has no referent here and will read as stale the first time someone builds a real editor on it. Stating the default and the escape hatch keeps the rationale and drops the claim:

```suggestion
        // URLs here are content, not endpoints. A caller that wants a
        // clickable endpoint passes `links: true`.
        links: false,
```

packages/graph-explorer/src/components/CodeEditor.test.tsx
```diff
@@ -43,4 +43,24 @@ describe("CodeEditor", () => {
...
+
+  test("should disable link detection by default", () => {
+    render(<CodeEditor defaultLanguage="json" value="{}" />);
+
+    expect(mockEditor.mock.calls[0][0].options).toMatchObject({
+      links: false,
+    });
+  });
+
+  test("should let a caller opt back in to link detection", () => {
```
kmcginnes: The `...options` spread already guarantees the override for every scalar option, so that isn't the new behavior. What this test pins is that `links: false` sits *above* the spread, which is the regression worth catching:

```suggestion
  test("should let a caller's options override the shared default", () => {
```

Worth knowing the two tests only pin the placement as a pair. Test 1 alone still passes with the line moved below the spread, and test 2 alone still passes if the line is deleted entirely. Fine as is, just don't remove either one in isolation later.

kmcginnes (dismissed):
Approving. Four nits, none blocking. Two are inline suggestions you can apply directly; the other two are below.

**Tests assert forwarding, not rendering.** Both new cases check that the option reaches the mocked `Editor`, not that a URL renders as plain text. That's close to the ceiling given `setupTests.ts:105` mocks `Editor` as `() => null`, so I'm not asking for more. I verified the behavior by hand in both dialogs with `main` as a control: on `main` the URL gets a `detected-link` span, blue underline, and the "Follow link (cmd + click)" tooltip; on this branch there's no link span at all and cmd-hover leaves the cursor as `text`. Your call whether a `PanelError` test with a local mock is worth it.

**The "Open Link" check in the description can't fail.** `OpenLinkAction` is registered with `precondition: undefined` and no `contextMenuOpts` (`contrib/links/browser/links.js:348-374`), and the links contrib adds nothing to `MenuId.EditorContext`, so it's command-only and was never in that menu on either build. The three checks that do differ, and that I confirmed: no hover underline, no "Follow link (cmd + click)" tooltip, no ctrl-click navigation.

Nice call putting this in the shared options block rather than per call site. All three callers are read-only and there's no editable `CodeEditor` anywhere, so the shared default beats three copies, and it means a fourth viewer can't silently get links back.

One unrelated thing I turned up and will file separately: `@monaco-editor/loader` pulls Monaco from `cdn.jsdelivr.net` at runtime with no integrity check, and it defaults to `0.55.1` while `package.json` pins `0.56.0`. So the pinned version never actually runs. Pre-existing, not yours.

mjuarros:
Thanks for the review. Pushed `ce58ee91` addressing the two inline suggestions, and updated the description. Summary of what changed and what I deliberately left alone:

**Applied**

- **Comment reword (`CodeEditor.tsx`).** Dropped the "these viewers…" wording. You're right that `CodeEditor` can't know its callers, so the phrase had no referent and would read as stale the moment a fourth caller appears — and `PersistenceStatusIndicator` renders local persistence state, not a database/proxy response, so the old claim was already slightly off. Now it states the default and the escape hatch:
  ```
  // URLs here are content, not endpoints. A caller that wants a
  // clickable endpoint passes `links: true`.
  links: false,
  ```
- **Test rename (`CodeEditor.test.tsx`).** Renamed to `"should let a caller's options override the shared default"`. Agreed the `...options` spread already overrides any scalar, so the previous title oversold it — what the pair actually pins is that `links: false` sits *above* the spread. Noted your warning that the two tests only guard placement as a pair (test 1 passes with the line moved below the spread; test 2 passes if it's deleted), so neither should be removed in isolation.
- **Description.** Removed the "no 'Open Link' in the context menu" line from Validation, since you confirmed `OpenLinkAction` is command-only and never in `MenuId.EditorContext` on either build. The description now lists only the three differences you verified: no hover underline, no "Follow link (cmd + click)" tooltip, no ctrl-click navigation.

**Deliberately not done**

- **No `PanelError` render test with a local mock.** You said you weren't asking for more given `setupTests.ts` mocks `Editor` as `() => null`, and you verified the rendering by hand. A local mock to assert plain-text rendering isn't worth the complexity here.
- **The `@monaco-editor/loader` CDN / version-pin finding** is pre-existing and out of scope for this PR — leaving it for the separate issue you're filing.

Verification after the change: `pnpm checks` passes (lint + format + types across all workspaces) and all 4 `CodeEditor` tests pass.

kmcginnes (approved):
Re-approving `ce58ee91`. Both suggestions applied verbatim, description updated, and `links: false` itself is untouched, so the behavior I verified by hand still holds and I didn't re-run the browser check.

Confirmed on my side: `pnpm checks` clean (oxlint, oxfmt across 963 files, `tsc --noEmit` across all 4 projects) and the 4 `CodeEditor` tests pass.

Agreed on both of your deliberate omissions. The `PanelError` mock isn't worth it, and the Monaco CDN finding is mine to file separately.

Issue #2180, opened 2026-09-14
Stop using Iterator helpers so core-js can be removed


kmcginnes:
Iterator helpers (`Iterator.prototype.map`/`filter`/`toArray`, `Iterator.from`) are Baseline **newly available** as of March 2025, so they will not reach Baseline Widely Available until roughly September 2027. Vite's default `build.target` is `baseline-widely-available`, which in Vite 8 resolves to `['chrome111', 'edge111', 'firefox114', 'safari16.4', 'ios16.4']`, and [Vite only handles syntax transforms and does not cover polyfills](https://vite.dev/guide/build#browser-compatibility). That is why `packages/graph-explorer/src/index.tsx` imports a core-js polyfill, and why `core-js` is a production dependency (#2130, #2167).

The usage does not earn that cost. Across 30 files (28 source, 2 test) only four helpers appear, and `.toArray()` accounts for the large majority. Every call site is drained eagerly, so nothing relies on laziness.

## Work

Rewrite the call sites to natively supported equivalents:

- `m.values().toArray()` becomes `Array.from(m.values())`
- `m.values().map(f).toArray()` becomes `Array.from(m.values(), f)`, with no intermediate array
- `m.values().filter(p)` becomes `[...m.values()].filter(p)`

The 7 sites that do not end in `.toArray()` are each immediately consumed by `new Map(...)`, `Promise.all(...)`, or a similar iterable consumer, so they rewrite the same way: `connector/sparql/fetchNeighbors/index.ts:39` and `:53`, `connector/sparql/neighborCounts.ts:268`, `core/StateProvider/neighbors.ts:157`, `modules/GraphViewer/GraphViewer.tsx:95` and `:98`.

Then remove the polyfill import from `index.tsx` and remove `core-js` from `packages/graph-explorer/package.json` and `pnpm-lock.yaml`.

## Approach

Narrowing the TypeScript `lib` is the instrument. `packages/graph-explorer/tsconfig.json` currently sets `lib: ["ESNext", "DOM", "DOM.Iterable"]`, which is the same class of mistake as the `devDependencies` entry was: it claims a runtime we do not ship to. Under `ESNext` a `.take()` call typechecks clean, so nothing catches a new helper today.

Measured by editing the `lib` value and running `pnpm check:types`:

| `lib` | Errors | Catches |
| --- | --- | --- |
| `ESNext` (today) | 0 | nothing |
| `ES2025` | 12 | `Error.isError` only; iterator helpers still allowed |
[...]

Issue #2157, opened 2026-09-10
Move babel-plugin-react-compiler to devDependencies


kmcginnes:
`babel-plugin-react-compiler` is declared in `packages/graph-explorer/package.json` under `dependencies`, but it is build-time tooling. It ships into the runtime image as a result.

Nothing in `src/` imports it. It reaches the build through `vite.config.ts`, which calls `reactCompilerPreset()` from `@vitejs/plugin-react` and hands it to `@rolldown/plugin-babel`. The workspace root already declares it correctly in `devDependencies`.

The reason it ends up in the image is the `Dockerfile` sequence: `pnpm install --frozen-lockfile` (full), `pnpm build`, `pnpm clean:dep`, then `pnpm install --prod --frozen-lockfile --ignore-scripts`. The build runs against the full install, so moving the entry to `devDependencies` won't affect it — but the `--prod` install afterwards keeps anything in `dependencies`, so the plugin and its closure (about 3.7 MB) land in the shipped image for nothing.

## Work

Move `babel-plugin-react-compiler` from `dependencies` to `devDependencies` in `packages/graph-explorer/package.json`. Leave the root declaration alone.

## Acceptance

`pnpm build` still emits the compiler's memoization (the output should contain `react/compiler-runtime` imports with cache slots — a quick way to confirm the preset still ran), `pnpm checks` and `pnpm test` pass, the image builds, and `babel-plugin-react-compiler` is absent from the built image's `node_modules`.

<audience-footer>

> [!IMPORTANT]
> Internal only — this issue is maintained by the core team and is not accepting external contributions.

</audience-footer>

kmcginnes:
Done in #2232, merged as 6597fd7e. \`babel-plugin-react-compiler\` is now in \`devDependencies\` for \`packages/graph-explorer\`.

Verified against the acceptance criteria by building the image before and after and diffing the shipped \`node_modules/.pnpm\`. The move drops four packages, the plugin plus its closure:

\`\`\`
babel-plugin-react-compiler@1.0.0
@babel/types@7.29.8
@babel/helper-string-parser@7.29.7
@babel/helper-validator-identifier@7.29.7
\`\`\`

Store goes from 306 entries to 302, 392 MB to 384 MB. \`@babel/runtime\` stays because it is a real runtime dependency. The built image still serves: \`/status\` returns 200 and the shipped \`dist\` keeps the compiler memoization across all 1883 assets.

[...]

Issue #2130, opened 2026-08-15
core-js is a devDependency but is used in production code


kmcginnes:
## Description

`core-js` is declared in `devDependencies` (`packages/graph-explorer/package.json:97`, `^3.49.0`), but it's actually consumed by production application code: `packages/graph-explorer/src/index.tsx:7` has `import "core-js/full/iterator";`, which polyfills the Iterator helpers proposal (`Iterator.prototype.map`/`filter`/`take`/`drop`/`flatMap`/etc.) and ships in the app bundle.

Since this import runs in the browser, `core-js` should be a regular `dependencies` entry, not a `devDependencies` entry — `devDependencies` are conventionally build/test-only and aren't guaranteed to be installed or resolved the same way in all consumption scenarios.

The import was added in commit `954882b3` (November 2024, #658) with no explanation of the dependency placement, so there's no known reason it was placed under `devDependencies`.

## Expected Behavior

`core-js` should be moved from `devDependencies` to `dependencies` in `packages/graph-explorer/package.json`.

---

<audience-footer>

> [!IMPORTANT]
> Internal only — this issue is maintained by the core team and is not accepting external contributions.

</audience-footer>

PR #2192, opened 2026-09-18
Replace Error.isError with instanceof Error in createErrorDetails

### LINKED ISSUE #2181: Error.isError is not available in all supported browsers
## Description

`createErrorDetails` calls `Error.isError`, which is Baseline **limited availability**. [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/isError) states it "does not work in some of the most widely-used browsers." The call is unguarded and nothing polyfills it, so on a browser without the method it throws `TypeError: Error.isError is not a function`.

That throw happens inside the function whose job is to describe an error, so instead of seeing a message the user gets a failure in the error path itself.

- `packages/graph-explorer/src/utils/createErrorDetails.ts:52`
- `packages/graph-explorer/src/utils/createErrorDetails.ts:66`

Five consumers reach it: `components/PanelError.tsx`, `core/StateProvider/persistence/writeQueue.ts`, `core/StateProvider/persistence/persistenceStatusStore.ts`, `core/StateProvider/migrateUserStyling.ts`, and `core/StateProvider/migrateUserLayout.ts`.

Introduced in `204f07cb` (#1644).

This is also the same class of mistake as #2130 and #2157: using a runtime feature newer than the browsers we build for. Vite's default `build.target` is `baseline-widely-available`, which resolves to `['chrome111', 'edge111', 'firefox114', 'safari16.4', 'ios16.4']` in Vite 8, and [Vite only handles syntax transforms and does not cover polyfills](https://vite.dev/guide/build#browser-compatibility).

## Environment

- OS: any
- Browser: any without `Error.isError`, which includes every browser at the low end of our supported range
- Graph Explorer Version: 3.1.0 onward (since #1644)
- Graph Database & Version: not database-specific

## Steps to Reproduce

1. Open Graph Explorer in a browser without `Error.isError`, or simulate it in devtools with `delete Error.isError`
2. Trigger any error that renders through `PanelError`, such as querying a connection whose endpoint is unreachable
3. The error panel fails instead of showing the message. The console reports `TypeError: Error.isError is not a function` from `createErrorDetails`

## Expected Behavior

The error message renders normally on every supported browser.

Replace both calls with `... instanceof Error`. That is sufficient here: `DOMException` subclasses `Error` in browsers, so IndexedDB failures reaching the persistence paths still narrow correctly. The only behavioural difference is cross-realm errors, and the app's one worker boundary is Monaco, whose errors do not route through this function. No polyfill is needed, which matters because #2180 removes `core-js` entirely.

## Related Issues

- Blocks #2180. Narrowing `lib` to `ES2024` in that issue turns this into a compile error: 15 errors, 2 x TS2550 for the method plus 13 x TS18046 where `error` and `cause` stay `unknown` once the type guard stops narrowing. `pnpm check:types` cannot pass until this is fixed.
- Same class as #2130 and #2157

<audience-footer>

> [!IMPORTANT]

</audience-footer>

mjuarros:
## Description

`createErrorDetails` narrowed unknown values with `Error.isError`, which is Baseline limited availability and is not polyfilled here. On a browser without the method, the call threw `TypeError: Error.isError is not a function` from inside the function whose job is to describe an error, so the error panel failed instead of rendering the message.

Both call sites now use `instanceof Error`, which is sufficient here:

- The four earlier branches (`ServerConnectionError`, `NetworkError`, `ZodError`, `QueryValueError`) all match `Error` subclasses and are evaluated first, so their output is unchanged.
- `DOMException` subclasses `Error`, so the IndexedDB failures that reach the persistence and migration paths still narrow correctly.
- The only behavioural difference is cross-realm errors, which do not route through this function.

No polyfill was added, which matters because #2180 removes `core-js` entirely.

## How to read

1. [`createErrorDetails.ts`](https://github.com/aws/graph-explorer/pull/2192/files#diff-3bca17605d563c9474bd7722147d602d8eefbe4018557f3aa54fa2bdea9d4a1b) — the core change: two guards, one in `createErrorDetails` and one in `serializeCause`
2. [`createErrorDetails.test.ts`](https://github.com/aws/graph-explorer/pull/2192/files#diff-da6a4971206e825de665a9611afc20f3cbdd8be9bc77cf9e0589ab70a19531c3) — three tests covering the absent-method path, plus the `withoutErrorIsError` helper. It deletes the property reflectively rather than referencing `Error.isError` as a typed property, so the test still compiles once #2180 narrows `lib` to `ES2024`
3. [`docs/agents/testing.md`](https://github.com/aws/graph-explorer/pull/2192/files#diff-d00af24f807aed3edd173c6e7d4bcf4a694b2ccf26b17a19105e534d94f43164) — supporting: records that absent-global test pattern as a convention

## Validation

- The three new tests were confirmed failing before the fix, with exactly `TypeError: Error.isError is not a function` at `createErrorDetails.ts:52`, and pass after it.
- `pnpm checks` passes. `pnpm test` passes at 2675 tests across 220 files. The branch is rebased onto current `main`, so this includes the Vitest 5 upgrade from #2166.
- Verified the `#2180` unblock rather than assuming it: setting `lib` to `["ES2024", "DOM", "DOM.Iterable"]` and running `pnpm check:types` reports no remaining errors attributable to this call site. That `tsconfig.json` edit was reverted and is not part of this PR.
- No user-visible output changes on browsers that do have the method: only the guard conditions changed, never the branch bodies.

## Related Issues

- Fixes #2181
- Related to #2180, which this unblocks
- Related to #2130 and #2157, the same class of using a runtime feature newer than the browser floor we build for

### Check List


mjuarros:
Nothing here enforces the floor, and nothing in this PR can. I've filed #2237 with the investigation. 
Two things worth knowing up front: oxlint 1.82 can load ESLint plugins via jsPlugins, so this is reachable without adding ESLint, and eslint-plugin-baseline-js pinned to available: 2022 flags Error.isError and Web APIs. Running the candidate stack over src also turned up two violations already shipping on main Array.prototype.toSorted in displayTypeConfigs.ts:109 and displayAttribute.ts:16, which Firefox 114 lacks. So the class is live right now, not just historical. 
I've also commented on #2180, because its planned lib: ES2024 permits Promise.withResolvers and Object.groupBy, both Safari 17.4; ES2023 closes those for 5 more type errors.

On the tests: they are a regression guard for this function, not a guard for the class. Reverting line 52 back to Error.isError fails all three of them, so they hold that one function in place. Each covers something distinct: serializeCause is only reached
  when a cause is present, so the cause-chain case is the only one exercising the second call site at line 66, and the DOMException case is the issue's reproduction.
I did try replacing the reflective deletion with a vi.spyOn mock, which would have been shorter. Two things ruled it out. @vitest/expect calls Error.isError internally, so a test-scoped spy is still installed while assertions run — mocking it to throw fails from inside toStrictEqual, not from the code under test. And vi.spyOn(Error, "isError") stops compiling under the narrowed lib that #2180 introduces, which would mean shipping a cast into the branch whose purpose is to unblock that narrowing. The current helper avoids both by scoping the deletion to the single call and keeping the property name out of the type system.

kmcginnes (approved):
Verified this works. I reproduced the crash on the pre-fix commit in Chrome with `Error.isError` deleted, and it's gone here. With the method present, output is identical on both commits.

Approving. Three non-blocking notes below.

### You convinced me on the test block

I was going to ask you to drop `describe("browsers without Error.isError")` and the helper, on the grounds that nothing in `createErrorDetails` reads `Error.isError` anymore so the deletion is inert. Your reasoning is better than mine. Two points I'd under-weighted: within that block the cause case is the only one reaching the second call site at line 66, since `serializeCause` needs a cause present. And you checked the shorter `vi.spyOn` version and found concrete reasons it fails, including that `@vitest/expect` calls `Error.isError` internally so the mock breaks `toStrictEqual` from the inside. Documenting the alternative you rejected is what settled it. Keeping them.

Two small things in that block, neither blocking:

- `includes the nested cause chain as data` builds a single-level cause, so it never reaches the recursion at `createErrorDetails.ts:75`. It also nearly duplicates the name of `:67`, which does build a chain. Rename it, or add a `mid` level.
- `withoutErrorIsError` ignores the `Reflect.deleteProperty` result. On a runtime without `Error.isError` the delete is a no-op and all three tests pass having exercised nothing. `expect(descriptor).toBeDefined()` before the delete covers that.

### Worth folding into the description

The impact is worse than [Error.isError is not available in all supported browsers #2181](https://github.com/aws/graph-explorer/issues/2181) describes, and its repro steps don't actually work:

- An unreachable endpoint throws `TypeError`, which the `ServerConnectionError` branch catches at line 17, before the guard. Reaching it needs a `DOMException`, for example a listener that accepts and never responds against a fetch timeout.
- On the old code the whole route was lost to the app error boundary, not just the error panel. And on the write queue path a `QuotaExceededError` escaped as `Uncaught (in promise)`, so the "Changes not saved" indicator never appeared at all. A user with full storage got no warning that nothing was being saved.

### Wording

"No user-visible output changes on browsers that do have the method" isn't quite absolute. `Object.create(Error.prototype)` gives `Unknown Error` / `{}` before and `Error` / `""` after, since `Error.isError` tests the internal slot while `instanceof` only walks the prototype chain. Nothing in the app produces such a value, so no code change needed, but worth narrowing since this becomes the squash commit message.

### Before merge

Needs a rebase. The branch is 10 commits behind `main`, and the red `install-and-test` check is the `useSchemaSync` flake fixed by [Stop flaky assertions in proxy server and schema sync tests #2220](https://github.com/aws/graph-explorer/pull/2220), which landed after this branch's base. A re-run at the current head would still use the pre-fix base.

Thanks for [Enforce the baseline-widely-available browser floor #2237](https://github.com/aws/graph-explorer/issues/2237). That's the right home for the enforcement question, and it's out of scope here.

mjuarros:
Fair on both counts, and the second one turned out to be bigger than it looked.

**On enforcement:** you're right that nothing here enforces the floor, and nothing in this PR can. I've filed #2237 with the investigation. Two things worth knowing up front: oxlint 1.82 can load ESLint plugins via , so this is reachable without adding ESLint, and  pinned to  flags  and Web APIs. Running the candidate stack over  also turned up two violations already shipping on  —  in  and , which Firefox 114 lacks. So the class is live right now, not just historical. I've also commented on #2180, because its planned  permits  and , both Safari 17.4;  closes those for 5 more type errors.

**On the tests:** they are a regression guard for this function, not a guard for the class. Reverting line 52 back to  fails all three of them, so they hold that one function in place. Each covers something distinct:  is only reached when a cause is present, so the cause-chain case is the only one exercising the second call site at line 66, and the  case is the issue's reproduction.

I did try replacing the reflective deletion with a  mock, which would have been shorter. Two things ruled it out.  calls  internally, so a test-scoped spy is still installed while assertions run — mocking it to throw fails from inside , not from the code under test. And  stops compiling under the narrowed  that #2180 introduces, which would mean shipping a cast into the branch whose purpose is to unblock that narrowing. The current helper avoids both by scoping the deletion to the single call and keeping the property name out of the type system.

Happy to trim to fewer cases if you still think three is too many, but I'd keep at least the cause-chain one since it's the only coverage of the second guard.

PR #2201, opened 2026-09-19
Force LF line endings for Docker shell scripts

### LINKED ISSUE #198: [Bug] exec ./docker-entrypoint.sh: no such file or directory #85[Bug]
Still seeing [Bug] exec ./docker-entrypoint.sh: no such file or directory #85

using Windows 10; Docker Desktop 4.23.0.  Have looked at the ./docker-entrypoint.sh in the container and have changed line endings to LF (they were CRLF for that file).  Problem remains

danielvanza:
## Description

- Add `text eol=lf` rules for `docker-entrypoint.sh`, `process-environment.sh`, and `setup-ssl.sh` so Windows checkouts cannot produce CRLF and break the container entrypoint.
- All three files are already LF in the repository, so this prevents regressions without rewriting history.

## Validation

- `git check-attr eol text -- docker-entrypoint.sh process-environment.sh setup-ssl.sh` — each file reports `text: set`, `eol: lf`.
- `git -c core.autocrlf=true clone --branch main` (pre-fix) yields `with CRLF line terminators` (25 CRs in `docker-entrypoint.sh`); the same clone of `shell-scripts-lf` yields `ASCII text` (0 CRs).
- `docker build -t graph-explorer .` → success; `docker run -p 8080:80 ... graph-explorer` + `curl http://localhost:8080/status` → `OK` (HTTP 200); container log shows `SSL disabled. Skipping self-signed certificate generation.`
- `docker exec ge sh -c "tr -cd '\r' < /graph-explorer/docker-entrypoint.sh | wc -c"` — `0` carriage returns.
- CRLF-before reproduction: `docker build` of the pre-fix clone → success, but `docker run` exits 1 with `exec ./docker-entrypoint.sh: no such file or directory` and `/status` is unreachable.
- `pnpm checks` — passed.
- `pnpm test` — passed (222 files, 2705 tests).

## Related Issues

- Fixes #198

### Check List


danielvanza:
Quick heads up that the workflows on this one are pending approval, so nothing has run yet. Once they're approved I'll keep an eye on the results. This just pins the three Docker shell scripts to LF to close out #198. Thanks!

